PR #47, opened 2023-07-31
Occupancy state graph

richard-oven-vanti:
Screenshot 2023-07-31 at 12 52 10

> Added background to bar chart interval segments which going to give visual feedback about the occupancy state


mattnathan:
I don't believe anybody will pick this up. We can do it from scratch when it gets to the top of the list

PR #70, opened 2023-09-27
Add explicit child devices in proxy driver

EllisBarnwell:
If the list of children is not present of empty in config then we announce all children of the parent node.

We currently list the traits explicitly too. This could be avoided by calling the meta data trait on the child device.


deanredfern26:
I have resurrected this PR but have removed the metadata lookup, the trait must be explicitly defined for each child. If no children are defined then the driver operates as before

mattnathan (approved):

PR #113, opened 2023-12-13
traits: add Point trait

mattnathan:


mattnathan:
Closing in favour of a future Points layer underlying most drivers/devices. Will revisit as part of that.

PR #183, opened 2024-02-13
driver/xovis: Add udmi trait to xovis driver

deanredfern26:
Adds the UDMI trait to the xovis driver. 

Similair pattern to the steinel hpd udmi implementation

### DIFF
--- a/pkg/driver/xovis/udmi.go
+++ b/pkg/driver/xovis/udmi.go
@@ -1,0 +1,150 @@
+package xovis
+
+import (
+	"context"
+	"encoding/json"
+	"path"
+	"time"
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/smart-core-os/sc-api/go/traits"
+	"github.com/smart-core-os/sc-golang/pkg/resource"
+	"github.com/vanti-dev/sc-bos/pkg/gen"
+)
+
+const PointsetVersion = "1.0.0"
+
+type PointsetEventMessage struct {
+	Version       string      `json:"version"`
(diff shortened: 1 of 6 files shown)


pkg/driver/xovis/udmi.go
```diff
@@ -0,0 +1,150 @@
...
+				airQuality := change.Value.(*traits.EnterLeaveEvent)
+				appendEnterLeaveEventPoints(airQuality, &eventPoints)
+			}
+
+		case change, _ := <-occupancyChanges:
+			if change != nil {
+				temperature := change.Value.(*traits.Occupancy)
+				appendOccupancyEventPoints(temperature, &eventPoints)
+			}
+		}
```
hexaglow, line 113: When a channel closes, selects on it will complete immediately, which can cause a busy loop. It's not a big problem here because these channels only close when the context is cancelled, which would be caught by the top case shortly afterwards. But in general, when you have channels that might close, you should do something like this to avoid the busy-loop problem:
```go
for {
    select {
    case x, ok := <-ch1:
        if !ok {
            ch1 = nil
            break
        }
        process(x)
    case x, ok := <-ch2:
        if !ok {
            ch2 = nil
            break
        }
        process(x)
}
```

If you decide it's not a problem, then prefer `change := <-occupancyChanges` to `change, _ := <-occupancyChanges` - they have the same meaning.

hexaglow (changes requested):
I don't believe this actually works as-is. The UDMI implementation is never advertised on the registration. An integration test targeting the driver itself could have revealed this problem.
I believe we have a Xovis sensor in stock you can use to do end-to-end testing. Alternatively you could poke the webhook interface manually with Postman etc. to confirm it's working.

hexaglow (approved):

PR #192, opened 2024-02-15
opsui/overview: fix `energy graph` update on navigation

richard-oven-vanti:
We've noticed, that the energy graph doesn't trigger a fetch action when the active overview-child path changes. This has been fixed.

Jira: PRJ-117


deanredfern26 (approved):
Tested on site, working

PR #200, opened 2024-02-20
opsui/notifications: allow showing only unacknowledged alerts

richard-oven-vanti:
Refactored the _Filter.vue_ component to use a v-for loop when creating dropdown selectors and added a new selector option for listing _**All/Acknowledged/Unacknowledged**_ notifications.

allow showing only unacknowledged alerts

Jira: SC-4

### DIFF
--- a/ui/conductor/src/routes/ops/notifications/Filters.vue
+++ b/ui/conductor/src/routes/ops/notifications/Filters.vue
@@ -2,29 +2,14 @@
   <v-toolbar color="transparent" elevation="0">
     <v-spacer/>
     <v-select
-        hide-details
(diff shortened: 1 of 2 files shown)


mattnathan (approved):
Good work reducing the code bloat here and unifying all the dropdown filters.

I'm not 100% convinced that Acknowledged deserves to be a dropdown though, the items in it are fixed [any, yes, no] and the filter bar is starting to get cluttered now there are 4 independent filters showing there.

I'll make a ticket to revisit the filter in the future to try and rework the design for it.
https://vanti.atlassian.net/browse/SC-461

PR #202, opened 2024-02-20
driver/bacnet/marge: Add artus unit support

deanredfern26:
Adds in support for setting high/low deadband config regisiters required to control the Artus unit at Arup. Appreciate this might  not be the best way of doing this but it pretty clean and it works on site. Happy to create another ticket to refactor this to a more generalised solution after Arup's done. Closes [SC-447]

### DIFF
--- a/pkg/driver/bacnet/merge/air_temperature.go
+++ b/pkg/driver/bacnet/merge/air_temperature.go
@@ -152,7 +152,6 @@ func (t *airTemperature) pollPeer(ctx context.Context) (*traits.AirTemperature, 
 			if err != nil {
 				return comm.ErrReadProperty{Prop: "setPoint", Cause: err}
 			}
-			t.logger.Debug("DEAN: got set point: ", zap.Float64("setPoint", setPoint))
 			data.TemperatureGoal = &traits.AirTemperature_TemperatureSetPoint{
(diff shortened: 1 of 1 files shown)


pkg/driver/bacnet/merge/air_temperature.go
```diff
@@ -88,11 +90,37 @@ func (t *airTemperature) GetAirTemperature(ctx context.Context, request *traits.
...
+	return false
+}
+
 func (t *airTemperature) UpdateAirTemperature(ctx context.Context, request *traits.UpdateAirTemperatureRequest) (*traits.AirTemperature, error) {
 	if request.GetState().GetTemperatureSetPoint() == nil {
 		return t.GetAirTemperature(ctx, &traits.GetAirTemperatureRequest{Name: request.Name})
 	}
 	newSetPoint := float32(request.GetState().GetTemperatureSetPoint().GetValueCelsius())
+
+	if t.isArtusUnit() {
```
mattnathan, line 108: I'd be happy if this condition were more like `if t.config.SetPointLow != nil && t.config.SetPointHigh != nil` instead of using the metadata to change logic. Can't think of any reason why not to do it this way

mattnathan (changes requested):

mattnathan (approved):

PR #211, opened 2024-02-21
opsui/lightingTrait: refactor Lighting trait

richard-oven-vanti:
This refactor eliminates the _WithLighting_ renderless component, improves and extends the _useLightingTrait_ composable logic by centralising all values associated with the lighting trait used across the app, so they are consistently accessible from this composable.

Jira: SC-450


ui/conductor/src/traits/lighting/useLightingTrait.js
```diff
@@ -1,89 +1,133 @@
...
+      pullBrightness(req, light);
+      return light;
+    }] : [])
+  ]);
 
+  // Utility function to call the API with the query and the resource
+  watchResource(
+      () => toValue(queryObject),
+      () => toValue(paused),
+      ...apiCalls.value
```
mattnathan, line 56: There's no point in `apiCalls` being computed. Using it here in the function body of this composable negates the reactivity.

As `apiCalls` doesn't use any refs anyway you should just be able to either inline the array or convert `apiCalls` into a normal array.

ui/conductor/src/traits/lighting/useLightingTrait.js
```diff
@@ -1,89 +1,133 @@
...
+    return light.loading || lightUpdate.loading;
   });
 
+  // ---------------- Return ---------------- //
   return {
-    lightValue,
-    updateValue,
-    doUpdateBrightness
+    light,
+    lightUpdate,
```
mattnathan, line 122: Can these be named after the resource not the trait? With traits that have multiple resources this will get confusing.

mattnathan (changes requested):

mattnathan (changes requested):

mattnathan (approved):

PR #213, opened 2024-02-22
driver/mock: add fixed presets to the mock lighting driver

mattnathan:
Jira: SC-463


deanredfern26 (commented):
In mock/driver there are a bunch of todos in that switch statement, this could be a good time to clean those up? Looks like they are simple enough to resolve

mattnathan:
> In mock/driver there are a bunch of todos in that switch statement, this could be a good time to clean those up? Looks like they are simple enough to resolve

You're right, there are a bunch. Unfortunately none of the code they mention actually exists yet in sc-golang which is why they are listed as todo instead of just code. There's a bunch of tickets in Asana (yet to migrate over to Jira) that say things like "Add Model and ModelServer for input select trait" which need to be implemented before these todos can be removed.

deanredfern26 (approved):

PR #214, opened 2024-02-22
zone/lighting: add preset support

mattnathan:
Have zones implement the LightInfo api backed by the members of the zone - if they support it.

Jira: SC-463


deanredfern26 (approved):
Tested on site at OCW & working

PR #215, opened 2024-02-22
opsui/lightingTrait: add Lighting Presets support

richard-oven-vanti:
Lighting Preset support in action

Jira: SC-434 sc-435


deanredfern26 (approved):
Tested on site at OCW, working

mattnathan (approved):

PR #219, opened 2024-02-24
opsui/traits: combine trait values into its own composable

richard-oven-vanti:
This refactoring centralises all values associated with each trait used across the app, ensuring they are consistently accessible from a single composable.

Jira: SC-451 SC-452 SC-453 SC-454 SC-455 SC-456 SC-457 SC-458 SC-459 SC-460


mattnathan (changes requested):
I've got through 3/10 of the files here but I assume as all three have similar issues they'll apply to all the other files too, please fix them all.

More generally, you appear to be relying on Copilot a lot without reading/reviewing the output it generated - I assume this is what's happening. I need to you to review your own code please _before_ you submit it to other to review.

Also please read https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/ for general code comment advice, most of my comment comments can be avoided if you follow these guidelines.

ui/conductor/src/traits/status/useStatusTrait.js
```diff
@@ -7,8 +7,7 @@ import {StatusLog} from '@sc-bos/ui-gen/proto/status_pb';
 import {ref} from 'vue/src/v3/index.js';
```
rvnlive, line 7: This won't work. Need to move the import 2 lines above! @mattnathan

PR #220, opened 2024-02-24
opsui/airQuality: relocate useAirQuality composable from traits to routes

richard-oven-vanti:
This composable contains the main logic for the airQuality line chart, which should not be part of the traits folder.

Jira: N/A


mattnathan (commented):
This doesn't look like a relocation, it looks like a copy-paste? Is there a "remove from /traits" change that's missing?

mattnathan:
> @mattnathan it is a relocation and renaming. There is a new useAirQualityTrait composable in the traits folder.

I don't understand what you mean?

The PR says 'relocate and rename' but the body of the PR doesn't include any moving or renaming. I'm trying to understand how the description relates to the changed code/files.

mattnathan (approved):

PR #221, opened 2024-02-24
opsui/lightCard: truncate overflowing preset titles

richard-oven-vanti:
This PR contains a fix for an issue, where long preset button text was overflowing to each side of the button.
This has been fixed with truncating the labels.

Truncate long example preset title

Jira: SC-464

### DIFF
--- a/ui/conductor/src/traits/lighting/LightCard.vue
+++ b/ui/conductor/src/traits/lighting/LightCard.vue
@@ -39,25 +39,27 @@
     </v-card-actions>
 
     <!-- Presets -->
-    <v-container v-if="brightnessPresets.length > 0" class="pa-0">
+    <v-container v-if="brightnessPresets.length > 0" class="pa-0 px-4">
       <v-list tile class="ma-0 mt-2 pa-0">
-        <v-list-item class="py-1">
+        <v-list-item class="py-1 pl-0">
           <v-list-item-title class="text-body-small text-capitalize">Presets</v-list-item-title>
         </v-list-item>
       </v-list>
-      <v-card-actions class="flex-column justify-center px-4">
-        <v-btn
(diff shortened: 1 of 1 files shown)


ui/conductor/src/traits/lighting/LightCard.vue
```diff
@@ -115,12 +120,37 @@ const brightnessControl = computed(() => {
...
+    width: 100%;
+    max-width: 575px;
+
+    &--content {
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+    }
+  }
+}
```
mattnathan, line 155: This is _waaayyy_ more complicated than it needs to be

```html
<v-btn block ...>
  <span class="text-truncate">{{ preset.title ? preset.title : preset.name }}</span>
</v-btn>
```

Should be enough right?

mattnathan (changes requested):

ui/conductor/src/traits/lighting/LightCard.vue
mattnathan, line 1: This works for me:
```
<v-btn class="... preset" ...>
  <span class="text-truncate">{{ ... }}</span>
</v-btn>

<style scoped>
.preset ::v-deep(.v-btn__content) {
  max-width: 100%;
}
</style>
```

If you're going to change it to a list then you shouldn't need the enclosing div. But that's a bigger change than is needed to fix this

mattnathan (changes requested):

ui/conductor/src/traits/lighting/LightCard.vue
richard-oven-vanti: I've kept it as a list and removed the div.

mattnathan (approved):

PR #223, opened 2024-02-26
opsui/notifications: upgraded and expanded filtering of alerts on the notifications page

richard-oven-vanti:
Completely redesigned the filter and its functionality.
Improved flexibility, added extra logic and missing functionality.

> Note: at the moment it is only being used on the Notifications page.

Redesigned filter and its functionality

Jira: SC-3, SC-461

### DIFF
--- a/ui/conductor/src/components/VSwitch2.vue
+++ b/ui/conductor/src/components/VSwitch2.vue
@@ -1,0 +1,45 @@
+<template>
+  <v-switch
+      v-bind="$attrs"
+      :class="{indeterminate: _indeterminate}"
+      :input-value="_value"
+      @change="emit('input', $event)"
+      v-on="$listeners">
+    <template #label v-if="$slots.label">
+      <slot name="label"/>
+    </template>
+  </v-switch>
+</template>
+<script setup>
+import {computed, ref, watch} from 'vue';
+
+const props = defineProps({
+  value: {
+    type: null,
+    default: undefined
+  },
+  indeterminate: {
+    type: Boolean,
+    default: false
+  }
+});
+const emit = defineEmits(['input', 'update:indeterminate']);
+
+const _indeterminate = ref(/** @type {boolean} */ props.indeterminate ?? false);
+watch(() => props.indeterminate, () => {
+  _indeterminate.value = props.indeterminate;
+});
+watch(_indeterminate, (newValue, oldValue) => {
+  if (newValue === oldValue) return;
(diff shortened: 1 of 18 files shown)


mattnathan (changes requested):
image

1. [x] Don't use `primary` colour for any buttons that aren't the primary action on the screen. Applying filters is not the primary action
2. [x] Remove the "no filters" label - if there are no filters just show nothing.
3. [x] The alignment of the button seems off to me, it's not aligned with the table or anything else
4. [x] Can you add a badge/dot to the filter button (or some indication) when (non-default) filters are active please. I know we have the chips but I still think this would be nice to have

image

5. [x] Please auto focus the search box, here and on the sub page
6. [ ] Keyboard interaction is a little off. You search, you then have to press down, then enter to navigate to the sub page. Is there a way to auto select (for interaction) the first item in the list when using the search box?
7. [x] Can you make sure there's a way to clear each filter from the filter menu please
8. [x] Can you add a "clear all" option to the filter menu

image

9. [x] I think a better UX for the severity menu would be a range slider
10. [ ] I think a better UX for the severity chips would be a single chip that describes the range: `Severity: INFO-WARN` or `Severity: INFO+` or something - you might even be able to remove the `Severity: ` part as `INFO` or `-WARN` is quite descriptive as it is.

image

11. [x] The chips don't look right, too small maybe I don't know. Maybe it's the comparison with the filter button 🤷 
12. [ ] I don't like the side scrolling. I get that we need to handle this case but if you fix number 6 then this can be replaced with a `...` or something indicating there are more filters and to look in the filter box for details.
13. [x] Can you use as much space as is available for showing the filter chips
14. [x] I want to be able to click on chips to adjust their filter settings: change the floor, adjust the severity, etc. If this is easy can you change it, if not we can add it as a later PR.

image

15. [x] The ack chip doesn't need the `yes` or `no` option, it can just say `acknowledged` or `unacknowledged`
16. [x] The ack filter is only ever true or false (or not set), a selection list isn't the right UX for this type of choice: a check box or a switch or something would be better. Not sure how that plays with the unset option but there's probably a way. 

Overall I really like the design you've come up with, it just needs a few tweaks and a bit of polish to improve the UX.

I haven't done a code review yet.

JamesHollidayVanti (approved):
All looks good, quite a few file changes. Improvement to the filter toggling is appreciated.

mattnathan (approved):

PR #224, opened 2024-02-26
opsui/OpsNav: Eliminate pullAlertMetadata error snackbar

richard-oven-vanti:
New - Error badge on Notifications navigation icon instead of snack bar

Jira: SC-1

### DIFF
--- a/ui/conductor/src/routes/ops/OpsNav.vue
+++ b/ui/conductor/src/routes/ops/OpsNav.vue
@@ -48,10 +48,10 @@
         <v-list-item-icon>
           <v-badge
               class="font-weight-bold"
-              :color="item.countType && counts[item.countType] ? 'primary' : 'transparent'"
-              :content="counts[item.countType]"
+              :color="item.badgeType ? badges[item.badgeType].color : 'transparent'"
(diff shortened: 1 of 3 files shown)


ui/conductor/src/routes/ops/OpsNav.vue
```diff
@@ -160,10 +161,47 @@ const enabledMenuItems = computed(() => menuItems.value.filter((item) => uiConfi
...
 const notificationEnabled = computed(() => uiConfig.pathEnabled('/ops/notifications'));
 
+/**
+ * Set the badge for the notification icon depending on the count or error
+ *
+ * @param {string} navItem
+ * @param {string} type
+ * @return {{color: string, value: number|string|null}}
+ */
+const setIconBadge = (navItem, type) => {
```
mattnathan, line 171: Can you rename this function. `set` is used for setting a variable/property/data somewhere, this function doesn't set anything

mattnathan (changes requested):
I think this change could be simpler and more maintainable if you changed `counts` to `badges`, got rid of the setIconBadge function, and added colour and the other logic in setIconBadge into the new `badges` reactive object.

Other things that could change: `menuItems.countType` => `menuItems.badgeType`

PR #230, opened 2024-03-01
auth: add missing Content-Type header to client credential token responses

mattnathan:
Without this the clientcredentials package fails to parse the response, assuming it's a form body and not json.


hexaglow (commented):
Does this need doing in `passwordFlow` too, as that also returns a JSON payload?

PR #233, opened 2024-03-21
auto/meteremail: add configurable retries & backoff timing 

deanredfern26:
Removes hardcoded backoffStart, backoffMax and noRetries when sending meteremails and allows these to be configured from config. If not present in config then default values are set.

Fixes test-meteremail program & adds new timing config to test program

### DIFF
--- a/pkg/auto/meteremail/config/root.go
+++ b/pkg/auto/meteremail/config/root.go
@@ -52,10 +52,10 @@ type Source struct {
 }
 
 type Timing struct {
-	Timeout      time.Duration `json:"timeout,omitempty"`
-	BackoffStart time.Duration `json:"backoffStart,omitempty"`
-	BackoffMax   time.Duration `json:"backoffMax,omitempty"`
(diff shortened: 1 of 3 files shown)


pkg/auto/meteremail/config/root.go
```diff
@@ -51,20 +51,22 @@ type Source struct {
 	Subsystem string `json:"subsystem,omitempty"`
 }
 
+type Timing struct {
+	Timeout      jsontypes.Duration `json:"timeout,omitempty"`
+	BackoffStart jsontypes.Duration `json:"backoffStart,omitempty"`
+	BackoffMax   jsontypes.Duration `json:"backoffMax,omitempty"`
+	NoRetries    int                `json:"noRetries,omitempty"`
```
mattnathan, line 58: Can you rename this to `NumRetries` or `MaxRetries` - _NoRetries_ reads like "there will be no retries" to me

mattnathan (changes requested):

mattnathan (approved):

PR #235, opened 2024-04-29
Fix es lint problems

deanredfern26:


ui/conductor/src/components/filter/BooleanChooserChip.vue
```diff
@@ -25,7 +27,13 @@ const props = defineProps({
...
 });
 
-const {title, choose, clear, value, text, defaultChoice} = props.ctx;
+const {title} = toRefs(props.ctx);
+const {choose} = toRefs(props.ctx);
+const {clear} = toRefs(props.ctx);
+const {value} = toRefs(props.ctx);
+const {text} = toRefs(props.ctx);
+const {defaultChoice} = toRefs(props.ctx);
+
```
mattnathan, line 36: These should all be like this:
```js
const {title, choose, clear, value, text, defaultChoice} = toRefs(props.ctx);
```

mattnathan (changes requested):

mattnathan:
I don't see those `toRef` issues when I run lint against the source project. Wonder what's going on.

mattnathan (approved):

PR #236, opened 2024-05-08
util/fetch: allow multi-level sub domains when using wildcard in cert

deanredfern26:
Allows validation of certifcates that use single wildcards for hosts with multiple sub-domain prefixes. More detail in the link below:

https://vanti.youtrack.cloud/issue/SC-528/Add-ability-for-http-client-to-validate-certificate-that-uses-wildcard


mattnathan:
We should not be doing this.

PR #237, opened 2024-05-22
auth: allow operators to start/stop lighting tests

mattnathan:
Adjust the OPA policies to grant access to these methods if you are an operator.

Issue: SC-540


hexaglow (approved):
No objections here, nice and simple

PR #241, opened 2024-06-05
ui: hide power values on electric demand card if not implemented

deanredfern26:
Hides power values on Electric card instead of showing zeros if they are not present

### DIFF
--- a/ui/conductor/src/traits/electricDemand/electric.js
+++ b/ui/conductor/src/traits/electricDemand/electric.js
@@ -61,7 +61,7 @@ export function usePullElectricDemand(query, paused = false) {
 export function useElectricDemand(value) {
   const _v = computed(() => toValue(value));
 
-  // return (v / d) or 0 if v is undefined
+  // return (v / d) or null if v is undefined or 0
   const div = (v, d) => v ? v / d : null;
 
   const realPower = computed(() => div(_v.value?.realPower, 1000));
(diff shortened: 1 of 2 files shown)


ui/conductor/src/traits/electricDemand/electric.js
```diff
@@ -62,15 +62,15 @@ export function useElectricDemand(value) {
...
+  const div = (v, d) => v ? v / d : null;
 
   const realPower = computed(() => div(_v.value?.realPower, 1000));
   const realPowerUnit = computed(() => 'kW');
   const apparentPower = computed(() => div(_v.value?.apparentPower, 1000));
   const apparentPowerUnit = computed(() => 'kVA');
   const reactivePower = computed(() => div(_v.value?.reactivePower, 1000));
   const reactivePowerUnit = computed(() => 'kVAr');
-  const powerFactor = computed(() => _v.value?.powerFactor ?? 0);
+  const powerFactor = computed(() => _v.value?.powerFactor ? _v.value?.powerFactor : null);
```
mattnathan, line 73: This isn't correct. If `powerFactor` is 0 then this returns null, whereas I expect you want to return null if powerFactor is null or undefined?

ui/conductor/src/traits/electricDemand/electric.js
```diff
@@ -62,15 +62,15 @@ export function useElectricDemand(value) {
   const _v = computed(() => toValue(value));
 
   // return (v / d) or 0 if v is undefined
-  const div = (v, d) => v ? v / d : 0;
+  const div = (v, d) => v ? v / d : null;
```
mattnathan, line 65: Not quite correct. 
1. The comment no longer matches the code
2. Returns null if v is 0, not just if v is null or undefined

mattnathan (changes requested):

PR #245, opened 2024-06-10
Dockerfile to build sc-bos entirely

hexaglow:
This isn't part of a ticket, but I was doing some related work on another repo and though sc-bos could benefit from the same thing.

Add a new Dockerfile which performs the build process internally, instead of copying in prebuilt artifacts. Move the old Dockerfile, and use it in the Github Action as before, so there's no behaviour change to the CI here (we could modify this later).

The new Dockerfile allows the user to simply run `docker/podman build -t sc-bos --secret=id=npmrc,src=$HOME/.npmrc .` and they will have a working sc-bos docker container. This is a smoother developer experience.

Multi-stage builds are used to ensure the final image is minimal, and doesn't contain the build tools or secrets. Cross-compilation is supported (using the `--platform` argument to `docker build`).

### DIFF
--- a/.github/docker/prebuilt.Dockerfile
+++ b/.github/docker/prebuilt.Dockerfile
@@ -1,0 +1,22 @@
+# This dockerfile is for copying in pre-built binaries and assets.
+# It is used by the GitHub Action docker-package.yml build process to create the release docker containers.
+
+FROM alpine:3.18
+LABEL vendor="Vanti Ltd"
+
+# automatically populated by the builder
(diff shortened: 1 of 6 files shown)


hexaglow:
I had to remove that git-revision-vite-plugin, as there's no way to stop the build crashing when the Git directory is unavailable (which it is when building inside Docker). I recreated its functionality in `vite.config.js`, which turned out to be pretty easy. 
Probably a good idea to remove this dependency anyway - it has very few downloads and is documented entirely in Chinese.

PR #246, opened 2024-06-10
SC-566: Remove private dependencies

hexaglow:
It should be possible to build sc-bos without credentials for our Nexus instance.

- [x] Redirect the npm-public mirror references directly to registry.npmjs.org
- [x] Integrate eslint config directly in repo, instead of including a package
- [x]  Publish `@smart-core-os/sc-api-grpc-web` ~~to NPM~~ to Github Packages (this is not quite as good but still solves the problem)


hexaglow:
This should be merged after https://github.com/smart-core-os/sc-api/pull/14

mattnathan (approved):

PR #247, opened 2024-06-13
auto/meteremail: Meteremail multiple recipients

deanredfern26:
Add support to send to multiple recipients, just wrap a loop around what we already have. 
Google no longers support multiple recipients in the to header https://westoahu.hawaii.edu/it/news/googleuh-gmail-will-reject-messages-with-non-compliant-headers-beginning-april-24-2023/


mattnathan (approved):
Looks good, I assume it was tested ;)

PR #250, opened 2024-07-25
proxy: update/add go docs

mattnathan:
They were lacking which made understanding the system and package hard.

I thought I'd do this before tackling upgrading the package to support dynamic API proxying.


hexaglow (approved):
This helps my understanding, thanks

PR #252, opened 2024-07-30
SC-582: use 0-100 scale for relative humidity

hexaglow:
Relates to https://github.com/smart-core-os/sc-api/pull/15

The hpd driver implemented it incorrectly.

The only other driver in sc-bos using this field is airthings, but I have not been able to find documentation on what scale the airthings structures use, so I'll leave that one.


hexaglow:
@mattnathan the new proxy test is failing, might want to take a look at that

PR #255, opened 2024-08-13
Add dynamic proxying support

mattnathan:
This PR adds the infrastructure and code to allow dynamic proxying to occur. Dynamic proxying is the ability to proxy requests for RPCs that the proxy doesn't know about in advance.

Each commit in the PR represents a step towards the dynamic proxy support:
1. The `unknown` package allows us to serve RPCs using reflection based proto descriptors, backed by a remote grpc server.
2. The `reflectionapi` package allows us to change the advertised types the grpc server hosting the proxy returns from its reflection api
3. The controller has new hooks to allow systems to adjust the above at runtime
4. The proxy system has been rewritten to allow the use of reflection to discover and proxy RPCs

## Implementation Notes

The new proxy system code is organised into two stages: scan and announce.

The scan stage performs a number of RPCs against the hub and enrolled nodes to discover the names and types associated with each node. Everything it discovers is placed into the structs defined in model.go.

The announce stage watches the model and updates the node with names and metadata, and configures the unknown and reflection objects associated with the controller.


hexaglow (commented):
Looks good in approach, apart from the limitations we have already discussed. I look forward to seeing the revised approach for dynamic registration.

PR #257, opened 2024-08-21
SC-606: Fix case of calling stop in task.PollErr and more tasks running

hexaglow:
Because select is nondeterministic when multiple cases are ready, Poll/PollErr might run one time extra after the context is cancelled. This is because the ticker channel may already contain an element sent during the previous poll operation. Add an explicit check for a completed context before checking for a tick.

### DIFF
--- a/pkg/task/poll.go
+++ b/pkg/task/poll.go
@@ -147,6 +147,11 @@ func (p poller) poll(ctx context.Context, run func(context.Context) error) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
+			if ctx.Err() != nil {
+				// If both <-ctx.Done() and <-ticker.C are ready, <-ticker.C could be selected, which we don't want.
+				// If the context is complete, we want to stop the poll immediately.
+				return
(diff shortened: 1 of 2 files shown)


hexaglow:
Hang on, the test still fails somehow - didn't see that locally. Looking deeper

hexaglow:
@mattnathan thinking about this some more, I don't believe we can totally eliminate this without changing the test. The change I've added makes it much less likely, but it can still happen occasionally. In the `PollErr/stops` test, there is always a (small) chance the context is cancelled in between the loop `select` and calling the action. In this case the action will be called with a cancelled context. 

I have loosened the test to permit a delta of 1 to account for this. In practice I don't think this will cause a problem - the extra run will receive a cancelled context, so will fail immediately assuming the action uses the context.

hexaglow:
The test fail is because of `TestPirsTurnLightsOn` which we have a separate ticket for

mattnathan (approved):
Good change. Also good catch with atomic.Int32

PR #259, opened 2024-08-21
Add package block for content-aware JSON config diffing and patching

hexaglow:
New package `block` can be used for diffing JSON-like config structures. Function `Diff` produces a slice of `Patch`es, which can be applied to the original A structure to transform it into the B structure. `Diff` takes a list of `Block`s, which define the boundaries withing the structure between changes. 

By designing the `Block`s to match the logical structure of the config, the user can apply patches to different config structures than were originally used to produce the `Diff`, with limited chance of conflict.


hexaglow:
@mattnathan I've separated this out from the original PR (#256) as discussed

pkg/block/block.go
```diff
@@ -0,0 +1,708 @@
...
+	} else if seg.IsArrayElem() {
+		return []any{}
+	} else {
+		panic("invalid path segment")
+	}
+}
+
+type Patch struct {
+	// How to reach the section to be patched from the root of the data structure
+	Path []PathSegment `json:"path"`
```
mattnathan: Can this be changes to `Path` please.

pkg/block/block.go
```diff
@@ -0,0 +1,708 @@
...
+	//      if the section is an array element, the element will be removed
+	Deleted bool `json:"deleted,omitempty"`
+}
+
+func (p *Patch) UnmarshalJSON(data []byte) error {
+	type patchAlias Patch
+	err := json.Unmarshal(data, (*patchAlias)(p))
+	if err != nil {
+		return err
+	}
```
mattnathan, line 654: Nice trick!

pkg/block/block.go
mattnathan: In fact I think most uses of `[]PathSegment` can be replaced with `Path`

pkg/block/block.go
```diff
@@ -0,0 +1,708 @@
...
+
+		subpatches := diff(c.A.Value, c.B.Value, c.B.Block)
+		prefixPatches(subpatches, fieldPathSegments(c.Path))
+		patches = append(patches, subpatches...)
+	}
+
+	return patches
+}
+
+func prefixPatches(patches []Patch, prefix []PathSegment) {
```
mattnathan, line 197: I think it'd make sense for all `[]PathSegment` to be changed to `Path`

pkg/block/path.go
```diff
@@ -0,0 +1,340 @@
...
+	var arr []PathSegment
+	err := json.Unmarshal(data, &arr)
+	if err != nil {
+		return err
+	}
+	*p = arr
+	return nil
+}
+
+func (p *Path) MarshalJSON() ([]byte, error) {
```
mattnathan: Please define MarshalJSON on `Path` not `*Path` - same with `String()`

mattnathan (changes requested):
I'm about half way through a review - most of path is reviewed and some of Diff. As there are a few bugs it'd be good to get them fixed and I'll finish the review later today or early next week.

pkg/block/block.go
```diff
@@ -0,0 +1,708 @@
...
+			return compareAny(a.ArrayElem, b.ArrayElem)
+		}
+	} else if a.IsField() {
+		return -1
+	} else if b.IsField() {
+		return 1
+	} else {
+		return 0
+	}
+}
```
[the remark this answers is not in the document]
hexaglow: It's passed directly to `slices.CompareFunc` which requires that signature. Or do you mean you think it should be public?

pkg/block/path.go
```diff
@@ -0,0 +1,340 @@
...
+func ParsePath(p string) (Path, error) {
+	if strings.TrimSpace(p) == "/" {
+		// single slash is an empty path, this would be rejected by main parsing logic
+		return nil, nil
+	}
+
+	r := strings.NewReader(p)
+	var parsed Path
+	for {
+		err := readSeparator(r)
```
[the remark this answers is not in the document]
hexaglow: I haven't actually implemented support for top-level arrays in `Diff`, so a path like that can't be produced currently. I'll document this current limitation.

pkg/block/block.go
```diff
@@ -0,0 +1,708 @@
...
+		return []Patch{{Value: b}}
+	}
+}
+
+func diffMap(a, b map[string]any, blocks []Block) []Patch {
+	tree := buildFieldTree(blocks)
+	a, aPages := splitMap(a, tree)
+	b, bPages := splitMap(b, tree)
+
+	patches := diffPages(aPages, bPages)
```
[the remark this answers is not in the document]
hexaglow: I've expanded on the docs on splitMap and fieldTree (now blockTree) - its purpose should be clearer now

pkg/block/block.go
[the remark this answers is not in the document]
hexaglow: I've added a `ComparePaths` function

pkg/block/path.go
```diff
@@ -0,0 +1,340 @@
...
+		b.WriteByte('/')
+		writeIdentifier(b, ps.Field)
+	} else if ps.IsArrayElem() {
+		b.WriteByte('[')
+		writeIdentifier(b, ps.ArrayKey)
+		b.WriteByte('=')
+		value, err := json.Marshal(ps.ArrayElem)
+		if err != nil {
+			panic(err)
+		}
```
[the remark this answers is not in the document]
hexaglow: I think the behaviour is correct - I've added some more tests to cover quotes and backslashes in array elements.

pkg/block/path.go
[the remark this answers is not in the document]
hexaglow: I've added support for parsing a Path string with a top-level array subscript like you mentioned.

pkg/block/block.go
```diff
@@ -0,0 +1,715 @@
...
+		}
+		for i := range a {
+			if !equal(a[i], b[i]) {
+				return false
+			}
+		}
+		return true
+
+	default:
+		return a == b
```
mattnathan, line 449: I'm not sure if this can actually happen, but just to be aware that comparing two `any` types that actually aren't comparable will result in a runtime panic

pkg/block/block.go
```diff
@@ -0,0 +1,715 @@
...
+	for k := range a {
+		if _, ok := b[k]; !ok {
+			return false
+		}
+	}
+	for k := range b {
+		if _, ok := a[k]; !ok {
+			return false
+		}
+	}
```
mattnathan, line 467: This second loop isn't needed, the length comparison and previous loop are enough as map keys are unique.

mattnathan (changes requested):

pkg/block/block.go
hexaglow: This can't happen because of the round-trip through json.Marshal/Unmarshal.
That will only output `map[string]any`, `[]any` (which are special-cased), `string`, `bool` and `float64` which are all comparable.

The comment indicates the permitted types. I'm happy to leave as-is because:

1. The logic of the program should prevent an invalid value appearing here
2. If there is somehow a bug which allows this, it'll panic and we'll know.
3. I can't think of a sensible behaviour for incomparable types (return `false`?)

mattnathan: I'm happy to leave as is.

PR #261, opened 2024-08-23
ui/ops: upgrade to Vue3, Vuetify3, vue-router4 and other required updates

mattnathan:
This PR includes all the changes needed to convert the SC BOS Ops UI from vue2 to vue3. Each commit in this PR takes one small step towards that goal - there are a lot of steps.

My development process followed these phases:
1. Update versions and bootstrap code
2. Use automated tooling (eslint-plugin-vuetify mostly) to fix specific migration problems one at a time, test it using eslint directly or using eslint-nibble to target specific checks.
3. Keep making changes until vite would run the app and it would load in the browser - this was a good day
4. Run the vanti-ugs example project with as many features turned on as possible and run the ui clicking on pages and checking the console - fix any warnings/errors you see in the console (that weren't there already)
5. Do a side-by-side comparison of every page, popup, interaction, component, etc of the main branch and this branch to check for differences
  1. For each difference, check (via text search) how many other places that pattern/property/etc was used and fix it
  2. Fix any related style relating to the component/page
6. Do a side-by-side comparison of this branch vs deployed ui at EW for all pages/popups/features available

The new styles do not match 100% with the old ones, they are not pixel perfect. Where decisions around layout appeared to be arbitrary I've erred on the side of fewer changes. Some examples: icon sizes are slightly smaller in some places. Font size too. Some icons have changed where the default vuetify ones were used. Table pagination has some more buttons. Select boxes and inputs have a different background colour. And so on.

I'm not expecting anyone to actually do a code review of the components/js, however the setup code should be looked at:
- vite.conf.js
- eslint.cjs
- src/main.js
- src/plugins/*
- src/routes/**/route.js
- there may be others

Instead I'd like reviewers to check out this branch and run the app against sites and local setups they know how to test with. This is a huge change that has likely broken something that I haven't spotted yet. I haven't been able to test the new version with all of the environments and sites out there, this is where I need help.


deanredfern26 (changes requested):
Couldn't see any issues at OCW, at EW I noticed the headings in the modes component are cut off. Can't see any other issues

image

mattnathan:
> Couldn't see any issues at OCW, at EW I noticed the headings in the modes component are cut off. Can't see any other issues
> 
> image

OK, that should be fixed now. I also noticed some issues with the UDMI card which I've fixed.

image

deanredfern26 (approved):

hexaglow (changes requested):
Visually looks good, I wouldn't notice any differences if you didn't point them out.

One issue - when running `yarn install` with Node 18 I encountered the following:

```
yarn install v1.22.22
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
warning chart.js@4.4.2: The engine "pnpm" appears to be invalid.
error glob@11.0.0: The engine "node" is incompatible with this module. Expected version "20 || >=22". Got "18.20.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

But the build-sc-bos action still specifies Node 18. If Node 20 is required then this needs to be updated.

mattnathan:
> Visually looks good, I wouldn't notice any differences if you didn't point them out.
> 
> One issue - when running `yarn install` with Node 18 I encountered the following:
> 
> ```
> yarn install v1.22.22
> [1/4] 🔍  Resolving packages...
> [2/4] 🚚  Fetching packages...
> warning chart.js@4.4.2: The engine "pnpm" appears to be invalid.
> error glob@11.0.0: The engine "node" is incompatible with this module. Expected version "20 || >=22". Got "18.20.0"
> error Found incompatible module.
> info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
> ```
> 
> But the build-sc-bos action still specifies Node 18. If Node 20 is required then this needs to be updated.

I'll update to the current release (`22.x`)

hexaglow (approved):

PR #263, opened 2024-08-30
traits: Add timestamp & zone to access attempt

deanredfern26:
Adds a timestamp and the name of the zone being accessed to the access trait


proto/access.proto
```diff
@@ -79,6 +79,12 @@ message AccessAttempt {
   // Actor describes the entity attempting to gain access.
   // Optional.
   Actor actor = 3;
+
+  // The time the access attempt was made.
+  google.protobuf.Timestamp time = 4;
+
+  // The name of the zone that was being accessed.
+  string zone = 5;
```
mattnathan, line 87: Maybe use a different name to `zone` (as we use that elsewhere in sc-bos) or better describe what zone means here. Is it a sc-bos zone name, is it a human readable name representing the room, does it come from the device? etc.

proto/access.proto
```diff
@@ -79,6 +79,12 @@ message AccessAttempt {
   // Actor describes the entity attempting to gain access.
   // Optional.
   Actor actor = 3;
+
+  // The time the access attempt was made.
+  google.protobuf.Timestamp time = 4;
```
mattnathan, line 84: Times should describe what they are, not just that they are a time: https://google.aip.dev/142#timestamps
For example `attempt_time` or `record_time`

mattnathan (changes requested):

proto/access.proto
```diff
@@ -79,6 +79,12 @@ message AccessAttempt {
   // Actor describes the entity attempting to gain access.
   // Optional.
   Actor actor = 3;
+
+  // The time the last successful access attempt was made.
+  google.protobuf.Timestamp last_successful_access_time = 4;
```
mattnathan, line 84: Better, I had no idea this was about successful times only. `Grant` uses the term `GRANTED`, it would be good use the same terminology here.

proto/access.proto
```diff
@@ -79,6 +79,12 @@ message AccessAttempt {
...
+
+  // The name of the zone that was last successfully accessed.
+  string last_successful_access_zone = 5;
```
mattnathan, line 87: I'm still not sure what this field is for, when I should use it (as a driver or as a client), or why it's different from looking up the zone in the metadata of the access device. What is a `zone` in the context of an access device?

mattnathan (changes requested):
Sorry to be picky, but these things tend to stick around for a long time so best to get them right

proto/access.proto
deanredfern26: It is something that is exposed through Gallagher's API, the zones are something defined on their system and is the last zone a cardholder successfully accessed

https://gallaghersecurity.github.io/cc-rest-docs/ref/cardholders.html

To be honest I was just trying to get as much useful stuff integrated as possible so we can show that we are surfacing data from their system.

mattnathan: Yeah, that's fine, if it's a concept access devices have then we should support it. I'm just trying to make sure someone looking at this trait in 6 months doesn't get confused because SC also has the concept of zones.

Reading the Gallagher docs it looks like these properties belong on the Actor type rather than the AccessAttempt type. They seem to describe what the cardholder last did, which maps to Actor in the Access trait.

I'd be happy to add this to that type, could shorten the names slightly too:

```protobuf
message Actor {
 ...
  // The time this actor was last granted access.
  // The grant may have been performed by a different device to this one.
  google.protobuf.Timestamp last_grant_time = ?;

  // The access zone the actor was last granted access to.
  // The zone will be provided by the device and may not map to any Smart Core zone or device.
  string last_grant_zone = ?;
}
```

Something like that, does it make sense?

mattnathan: You could see in the future also adding `last_seen_time` and other things like that to the actor message.

deanredfern26: Ah I missed the Actor field, thats where it belongs. I have moved the fields there and stuck them as 2 & 3 after the name as the stuff further down looked more like metadata

mattnathan (approved):

PR #264, opened 2024-09-03
SC-610: rename proxy system to gateway

hexaglow:
We call a node that uses the proxy system a gateway.
Rename the system to 'gateway' to make this clearer, and to differentiate it from the proxy driver.

The default system set still recognises 'proxy' as a valid name for the gateway system, to preserve compatibility with existing system config files.


pkg/system/gateway/factory.go
```diff
@@ -298,23 +301,23 @@ func (s *System) announceControllerNode(ctx context.Context, hubNode *gen.HubNod
...
 
 	// proxy any non-trait apis that also use routing
 	go s.retry(ctx, "proxyNodeApis", func(ctx context.Context) (task.Next, error) {
 		return s.announceNodeApis(ctx, hubNode, nodeConn)
 	})
 }
 
-// isProxy discovers if nodeConn is a proxy node, a node that has an enabled proxy system.
-func (s *System) isProxy(ctx context.Context, nodeConn *grpc.ClientConn) (bool, error) {
+// isGateway discovers if nodeConn is a proxy node, a node that has an enabled proxy system.
```
mattnathan, line 319: couple of missed renames here

hexaglow:
This implementation is the most straightforward way, but there's a downside: currently there is nothing to stop you from instantiating both a `gateway` and a `proxy` system, even though they are aliases.
This breaks the assumption that only one of each system can exist.

Fixing this would require adding information somewhere to say which service types are aliases, so we can resolve the true service type and prevent duplicates. This hard but doing it without breaking backwards compatibility for external users of `app.Controller` might be.

@mattnathan do you think this is a problem?

pkg/system/gateway/factory.go
mattnathan: Also I think you need to change the logic of this function to also check for the legacy name too.

mattnathan (changes requested):
Nice and simple.

How would you feel about also making the ops ui support both `"gateway"` and `"proxy"` when doing its checks too. I believe there's code in there that is similar to isGateway/isProxy that will need updating.

hexaglow:
@mattnathan I'll have a go at making the equivalent changes in the ops UI.
Any thoughts on my question above?

mattnathan:
> This implementation is the most straightforward way, but there's a downside: currently there is nothing to stop you from instantiating both a `gateway` and a `proxy` system, even though they are aliases. This breaks the assumption that only one of each system can exist.
> 
> Fixing this would require adding information somewhere to say which service types are aliases, so we can resolve the true service type and prevent duplicates. This hard but doing it without breaking backwards compatibility for external users of `app.Controller` might be.
> 
> @mattnathan do you think this is a problem?

Sorry, didn't see this.

While I'd like to spend some effort to prevent this kind of behaviour, in practice I think we'll be ok. I guess the easiest way would be to create a `legacy` system package that wraps other factories and prints a sufficiently scary log message about its use.

For now lets leave it and communicate between the 4 SC devs about it. In the future it might even be worth adding a general purpose JSON alias package so we can finally rename `"automation"` to `"automations"` 😉

hexaglow:
@mattnathan I've attempted the UI work, but it could do with a thorough review & test because of my lack of familiarity with the UI codebase

mattnathan (approved):
LGTM

PR #265, opened 2024-09-05
fix intermittent test failures

dloveen-v:
fix: order emitter bus elements to prevent rare test failure condition
  `pkg/auto/lights/auto_test.go` -> the lack of ordering on the bus means some states were being played later than usual causing the `waitForState()` call on line 158 to race ahead to the next assertion before the **_**actual**_** call to updateBrightness had been issued.

fix: move config changes up to mitigate risk of config not yet being applied before states are processed
  sometimes, in very rare cases, the config passed from the test isn't applied before calls to `setOccupancy` are issued, meaning the timestamps look at real time instead of unix zero time. There is still a very very low risk this can happen, but trying to sync the config with the first call to `setOccupancy` seemed like overkill for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ results_*.csv
 .secret/
 .secrets/
 .build/
+.vscode
--- a/pkg/auto/lights/auto_test.go
+++ b/pkg/auto/lights/auto_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/olebedev/emitter"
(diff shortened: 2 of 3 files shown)


dloveen-v:
I have also ran 
```bash
while true; do go test --count=200 -v ./pkg/auto/lights -run 'TestPirsTurnLightsOn'  || return; done
```

overnight to make sure the tests never fail as often as before.

mattnathan (approved):
Looks good, deceptively simple. Thanks for sticking with it

PR #266, opened 2024-09-05
SC-617: allow drivers, automations and zones to update their own config

hexaglow:
In preparation for user-editable config, drivers (etc.) need a way to trigger their own configuration to be modified.

Provide a new `service.ConfigUpdater` to each driver/auto/zone when constructing it. 
The service may use it to change its own configuration. The new config will be applied (using `service.Lifecycle.Configure(...)`), and if successful, persisted into the app's active config.

Also modify the implementation of `ServicesApi.ConfigureService` to store the updated configurations in the app's active config, instead of the old overlay directories. 

This isn't quite complete yet: we should have an example and/or tests of using the new functionality.
Also, `ServicesApi.CreateService` doesn't result in a runnable service - the app needs to restart first.


pkg/app/appconf/store.go
```diff
@@ -0,0 +1,298 @@
...
+		if typ != "" && typ != existing.Type {
+			return errors.New("driver type cannot be changed")
+		}
+		driverCfg = existing
+		driverCfg.Raw = data
+		updated.Drivers[idx] = driverCfg
+	}
+
+	// test that the updated config marshalls successfully - we don't want to store a bad config
+	_, err := json.Marshal(updated)
```
mattnathan, line 121: Could we get away with just testing the new/updated driver config marshals? I'm afraid that if we add bulk update features later this json.Marshal will cause it to be O(n!) complexity

pkg/app/appconf/store.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	s.active = updated
+	return nil
+}
+
+type DriverStore struct {
+	store *Store
+}
+
+func (ds *DriverStore) SaveConfig(_ context.Context, name string, typ string, data []byte) error {
```
mattnathan, line 82: It is very annoying that these type stores can't be more generic

pkg/app/services.go
```diff
@@ -224,3 +228,22 @@ func announceSystemServices[M ~map[string]T, T any](c *Controller, services *ser
...
+
+func (s *serviceConfigStore) UpdateConfig(ctx context.Context, data []byte) error {
+	serv := s.services.Get(s.id)
+	if serv == nil {
+		return fmt.Errorf("service %s not found", s.id)
+	}
+	_, err := serv.Service.Configure(data)
+	if err != nil {
+		return err
+	}
```
mattnathan, line 246: I'm not 100% sold on this bit. 

I think as the driver/etc will be the one calling this method, it already knows its config is updating and should handle it internally. I also think we might get issues here with ctx being cancelled before Save is called, it entirely depends on whether a driver/etc links the ctx given during `applyConfig` to the one passed to UpdateConfig.

If I'm thinking about it I think there are two separate concerns going on here:
1. The system wants to tell the driver the config has updated
2. The driver wants to tell the system the config has updated

Number 1 is for the general case: boot, generic APIs, getting config from the cloud, etc.
Number 2 is for the runtime case: submitting forms, discovery (of devices), user editing, etc.

pkg/task/serviceapi/api.go
```diff
@@ -568,4 +538,12 @@ func (a *Api) storeConfig(ctx context.Context, id, kind string, data []byte) err
...
 
-var fileNameEncoder = base32.StdEncoding.WithPadding(base32.NoPadding)
+type Store interface {
+	// SaveConfig will persist the configuration for a service.
+	// The data must be an encoded JSON object.
+	// name identifies the service, and is required.
+	// typ is the type of the service. If the store already contains a service with the given name, typ may be empty,
+	// in which case the existing type is used. If the store does not contain a service with the given name, typ must
+	// be non-empty.
+	SaveConfig(ctx context.Context, name string, typ string, data []byte) error
```
mattnathan, line 548: `name string, typ string` -> `name, typ string`

mattnathan (changes requested):

pkg/app/appconf/store.go
hexaglow: I'll have a go at it, but my gut feeling is that it won't reduce the line count much

pkg/app/appconf/store.go
hexaglow: Yes, good idea. Done.

pkg/app/services.go
hexaglow: I see what you mean. It's more flexible if we let the service handle reconfiguring itself. 

I'll try to make this clear in the documentation.

hexaglow: I've change the semantics of `ConfigUpdater` to not call `Configure()` any more.

mattnathan (approved):

hexaglow:
> Also, `ServicesApi.CreateService` doesn't result in a runnable service - the app needs to restart first.

Going to leave this as a future ticket because it's a nice-to-have rather than strictly needed.

PR #267, opened 2024-09-09
Feature/sc 613 check yarn.lock ci

dloveen-v:
.github/workflows: Add CI workflow and step to verify `yarn.lock` file doesn't contain references to private `nexus.vanti.co.uk` repositories

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -8,10 +8,6 @@ jobs:
     name: UI integrity check
     steps:
       - uses: actions/checkout@v4
-      - name: Private repo auth
-        run: |
-          git config --global url."https://${{ secrets.GO_MOD_TOKEN }}:x-oauth-basic@github.com/smart-core-os".insteadOf "https://github.com/smart-core-os"
-          git config --global url."https://${{ secrets.GO_MOD_TOKEN }}:x-oauth-basic@github.com/vanti-dev".insteadOf "https://github.com/vanti-dev"
       - name: Verify yarn.lock dependencies are public
         run: |
           if grep -E "https://nexus\.vanti\.co\.uk/repository/npm-public/" ui/yarn.lock; then
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -28,5 +28,5 @@ jobs:
           git config user.name github-actions
           git config user.email github-actions@github.com
           git add ui/yarn.lock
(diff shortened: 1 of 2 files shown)


dloveen-v:
I realised the ticket requests that the CI pipeline should amend the commit with the references in the `yarn.lock` file to private repositories removed.

Rather than amending the commit, I have opted to push a new commit labelled `auto` so we can see this was automatically generated easily.

.github/workflows/ui.yml
```diff
@@ -0,0 +1,32 @@
...
+          else
+            echo "No private URLs found."
+          fi
+      - name: Commit & push auto-edited yarn.lock file
+        if: ${{ env.YARN_LOCK_CLEANED }}
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git add ui/yarn.lock
+          git commit -m "[auto] fix: remove private repository references from ui/yarn.lock"
```
mattnathan, line 31: I don't think we need both `[auto]` and `fix:` - I'd be happy with just `[auto]`

mattnathan (changes requested):

.github/workflows/ui.yml
```diff
@@ -0,0 +1,32 @@
...
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+      - name: Clean yarn.lock private dependencies
+        run: |
+          echo "Checking for private yarn.lock dependencies"
+          if grep -E "nexus\.vanti\.co\.uk/repository/npm-public/" ui/yarn.lock; then
+            echo "Private URLs found in yarn.lock"
+            echo "Auto cleaning yarn.lock dependencies"
+            echo "YARN_LOCK_CLEANED=true" >> $GITHUB_ENV
+            sed -i -e 's/nexus\.vanti\.co\.uk\/repository\/npm-public/registry\.npmjs\.org/g' ui/yarn.lock
```
mattnathan, line 21: It occurred to me that this change is being made without testing that it works before committing the change. It might be worth running a `yarn install --frozen-lockfile` after the replace to make sure that the changes haven't broken dependency resolution.

mattnathan (changes requested):

.github/workflows/ui.yml
dloveen-v: Ahh good shout. I wasn't familiar with yarn. It looks like --frozen-lockfile is getting deprecated so I've used the recommended flags. Thanks!

mattnathan (approved):

PR #268, opened 2024-09-12
[SC-527] feat: upgrade go-jose from v3 to v4

dloveen-v:
Upgrade from go-jose v3 to v4

I have left a `TODO:` that we need to lock down the permitted algorithms used for signature creation / verification.

We can specify the cipher suite in KeyCloak installation but unfortunately, KeyCloak documentation doesn't specify what it uses as a default. It just says "reasonable default". I think we should start supplying the `https-cipher-suites` CLI flag to improve this in future.


pkg/auth/jwks/keyset.go
```diff
@@ -14,5 +16,5 @@ var (
 )
 
 type KeySet interface {
-	VerifySignature(ctx context.Context, jws string) (payload []byte, err error)
+	VerifySignature(ctx context.Context, jws string, permittedSignatureAlgorithms []jose.SignatureAlgorithm) (payload []byte, err error)
```
mattnathan, line 19: I'm not sure the algorithms belong as part of this interface. I think it might be better if the KeySet impl was created with permitted algorithms instead. Also I don't think it's a good idea to bake jose types into our interfaces.

mattnathan: At the very least we should either specify this as a `[]string` or our own signature type. `jose.SignatureAlgorithm` is just a string underneath anyway.

mattnathan (changes requested):

pkg/auth/jwks/keyset.go
dloveen-v: > I don't think it's a good idea to bake jose types into our interfaces.

I agree. I was conscious of not introducing the jose package in any package that didn't previously use it and also, trying to mitigate circular dependencies.

I didn't think introducing it in the interface was a bad idea as the package already included go-jose. I didn't think about keeping the interface clean though 🤦 

Should be a lot better now. Thanks!

internal/auth/tenant/token_server.go
```diff
@@ -67,6 +72,16 @@ func WithPasswordFlow(v Verifier, validity time.Duration) TokenServerOption {
...
 
+func WithPermittedSignatureAlgorithms(applyTo tokenSinkOption) TokenServerOption {
+	return func(ts *TokenServer) {
+		applyTo(ts)
+	}
+}
+
+func (s *TokenServer) GetSignedToken() auth.SignedToken {
+	return s.tokens
+}
```
mattnathan, line 83: I think this can be greatly simplified
```go
func WithPermittedSignatureAlgorithms(algs []jose.SignatureAlgorithm) TokenServerOptions {
  return func(ts *TokenServer) {
    ts.TokenSource.PermittedSignatureAlgorithms = algs
  }
}
```

You can then delete `auth.SignedToken`, `auth.TokenSink`, `SetPermittedSignatureAlgorithms` and so on.

mattnathan: If we want to avoid the jose package leaking into our types we can use `string` instead and do the conversion somewhere lower down.

mattnathan (commented):
Code looks good. What kind of e2e testing have you done with this setup?

dloveen-v:
> Code looks good. What kind of e2e testing have you done with this setup?

I've ran the backend and frontend locally with docker-compose. I then tried clicking around to see if the values are changing. Is there any more reliable way to test the auth flow?

mattnathan:
> > Code looks good. What kind of e2e testing have you done with this setup?
> 
> I've ran the backend and frontend locally with docker-compose. I then tried clicking around to see if the values are changing. Is there any more reliable way to test the auth flow?

Clicking around is what I expect, I just want to make sure you've run this against a keycloak instance, tried logging in/out, checked that when you have certain roles you can/can't do things you'd expect.

dloveen-v:
I thought it'd be good to confirm. 

I deleted my .data directory. Then ran SC-BOS with auth run config: 
```
go run cmd/bos/main.go \
--data .data/auth --appconf example/config/app.conf.json --sysconf example/config/system.auth.json
```
I imported the .pem cert file in to my keychain and all the gRPC calls are now using TLS SSL. I presume this confirms the PR if the gRPC XHRs are returning http.OK?

mattnathan (approved):
I've verified this locally by running sc-bos with keycloakAccounts setup in system.json, having --policy-mode=on and configuring the ops ui with the equivalent auth/keycloak settings.

PR #270, opened 2024-09-17
[SC-484] feat: refactor steinel driver to use 1 polling loop

dloveen-v:
Use a single polling loop for all `GetUpdate()` REST API `GET` calls in steinel driver

### DIFF
--- a/pkg/driver/steinel/hpd/poller.go
+++ b/pkg/driver/steinel/hpd/poller.go
@@ -1,0 +1,61 @@
+package hpd
+
+import (
+	"context"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+type sensor interface {
+	GetUpdate(response *SensorResponse) error
+	GetName() string
+}
+
+type Poller struct {
+	client       *Client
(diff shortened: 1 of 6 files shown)


pkg/driver/steinel/hpd/poller.go
```diff
@@ -0,0 +1,49 @@
...
+	ticker := time.NewTicker(p.pollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			for _, sensor := range p.sensors {
+				if err := sensor.GetUpdate(); err != nil {
```
deanredfern26, line 44: In each sensor's implementation of GetUpdate, they are all performing a GET request to the same URL. The idea of this ticket was to refactor out these identical requests so just one GET request is performed by the driver and then the sensors trait properties are updated based on the response from that

deanredfern26 (changes requested):

deanredfern26 (approved):

PR #271, opened 2024-09-17
ui/ops: allow users to edit the json config for drivers and automations

mattnathan:
image

This PR adds an edit button to the driver and automation sidebars that redirect the user to a new service edit page where they will be greeted with a json text editor to change the config to their hearts content.

Concurrent editing of the config has been accounted for (though is not allowed).

If the current user lacks the relevant permissions, they can only view the json config, not edit or save it.

I've tested the json editing against large (5k line) json documents and it's all fine as far as I can see.

WARNING: https://vanti.youtrack.cloud/issue/SC-628/Saving-config-results-in-type-in-active.json must be fixed before merging as any saved config before that will result in corrupted config and the sc-bos instance will no longer load. You can patch your local sc-bos by adding the following to `pkg/app/appconf/store.go#106`

```
} else {
	_, existingType := ops.getMetadata(serviceCfg)
	typ = existingType
}
```


mattnathan:
I forgot to mention that this change adds some new routes to the UI. Because of the way routes are shown/hidden from view you'll need to add/update your ui-config.json to make sure that the following routes are available in the `features` config.
- `automation`
- `system/driver`

The sample I used to test with looked like this:
```json
  "features": {
    "auth": {
      "third-party": true
    },
    "device": true,
    "devices": true,
    "ops": true,
    "site": false,
    "automations": true,
    "automation": true,
    "system": true
  },
```

deanredfern26 (approved):
I have tested this at OCW, modifying a driver of an area controller on the OPS UI of one edge gateway, saving and can see the changes reflected when viewing the config on the OPS UI of the other edge gateway. Looks like it is working well, I used the bms driver as that has the largest config and it seemed fine.

PR #273, opened 2024-09-18
all: add auto/driver service blocks

mattnathan:
This PR adds a few config blocks to some autos and drivers as a sample for how this should be done.

The blocks aren't added to all built in services, as the block system can be updated over time when we need to. However, I feel like these are a good starting point and should make the user config editing a bit more useful.

If we find in the future that we're writing tools that update only a small part of this config we can update the blocks then.
Similarly if/when we write dedicated forms for each of the drivers to edit the config outside a json editor, we can update them then too.

Fixes SC-629


deanredfern26 (approved):
Thanks, those examples are helpful

PR #276, opened 2024-09-23
driver/steinel: allow older TLS ciphers

deanredfern26:
Fixes [SC-635](https://vanti.youtrack.cloud/issue/SC-635/Fix-breaking-change-when-communicating-with-Steinel-sensors)

Allow all the ciphers to maintain backwards and forwards compatability for our Steinel sensors

### DIFF
--- a/pkg/driver/steinel/hpd/client.go
+++ b/pkg/driver/steinel/hpd/client.go
@@ -17,6 +17,38 @@ type Client struct {
 	Password string `default:""`
 }
 
+func getAllowedCiphers() []uint16 {
+	return []uint16{
+		// TLS 1.0 - 1.2 cipher suites.
+		tls.TLS_RSA_WITH_RC4_128_SHA,
+		tls.TLS_RSA_WITH_3DES_EDE_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_128_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_256_CBC_SHA,
+		tls.TLS_RSA_WITH_AES_128_CBC_SHA256,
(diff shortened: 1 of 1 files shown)


dloveen-v (commented):
`Nitpicking`
I did some investigation on this as it was likely me who broke this driver.

The comment on the field states:
>      // CipherSuites is a list of enabled TLS 1.0–1.2 cipher suites. The order of
> 	// the list is ignored. Note that TLS 1.3 ciphersuites are not configurable.
> 	//
> 	// If CipherSuites is nil, a safe default list is used. The default cipher
> 	// suites might change over time. In Go 1.22 RSA key exchange based cipher
> 	// suites were removed from the default list, but can be re-added with the
> 	// GODEBUG setting tlsrsakex=1. In Go 1.23 3DES cipher suites were removed
> 	// from the default list, but can be re-added with the GODEBUG setting
> 	// tls3des=1.

It might be worth removing the TLS 1.3 suites.

Also, RSA suites were removed as they don't provide forward secrecy. Once compromised, previous communications could also be decrypted if they were stored somewhere. I'm not sure on the implications on this. As all comms to steinel could be via a secure vLAN anyways. But I thought it was worth highlighting for the future potentially.

deanredfern26:
The steinels use RSA out of the box so we have to support it, even after updating to the latest firmware. 
While we can generate & upload new certs to them, there are a lot of these things out in the field and not really worth our time.  

SteinelKey

dloveen-v:
Approve. (I don't have a button to click)

deanredfern26:
@dloveen-v If you click on the files changed tab I think you should see a button on the top right where you can add your review and approve

dloveen-v (approved):

PR #277, opened 2024-09-23
auto/wordpress: voodoo chili wp automation

dloveen-v:
This is a helpful command to echo out POST requests to a local HTTP server 
```bash
docker run -p 80:8080 --rm -t mendhak/http-https-echo:31
```

I have marked this as WIP but feel free to review in the mean time.

Impls:
 - support units in mock automation Meter traits
 - added some more example JSON config to show this working

 - scaffolding for running wordpress automations
 - sensor job definitions
 - meter job definitions
 - refactored to reduce duplicate code in meter job definitions
 - JSON type defs for wordpress POST requests


pkg/auto/wordpress/job/energy.go
```diff
@@ -0,0 +1,101 @@
...
+func (e *EnergyJob) GetClients() []any {
+	return []any{&e.client, &e.infoClient}
+}
+
+func (e *EnergyJob) GetInterval() time.Duration {
+	return e.Interval
+}
+
+func (e *EnergyJob) Do(ctx context.Context, sendFn sender) error {
+	if len(e.Meters) < 1 {
```
deanredfern26, line 34: Wouldn't it be better to do these checks for meters, sensors etc before you create the jobs and not start them if there are no devices?

pkg/auto/wordpress/job/temperature.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+// TemperatureJob gets average air temperature at current point in time
+type TemperatureJob struct {
+	BaseJob
+	client  traits.AirTemperatureApiClient
+	Sensors []string
+}
+
+func (b *TemperatureJob) GetName() string {
```
deanredfern26, line 18: Receiver name is different

deanredfern26 (dismissed):
Have you tested this auto against their actual endpoint? I tried but it fails with a 403 `Request forbidden by administrative rules`. The request works fine in postman, so not sure what is going on there, I've never seen that error message before. I have stored their API key in Passportal but you will unlikely be able to get access to that until your probation is up, I'll DM it to you

pkg/auto/wordpress/job/energy.go
dloveen-v: Ahh yes I did think about this. I thought the config could eventually be changed dynamically so didn't want this defined during the Smart-core bootup.

pkg/auto/wordpress/job/air_quality.go
```diff
@@ -0,0 +1,61 @@
...
+	if len(a.Sensors) < 1 {
+		return nil
+	}
+
+	sum := float32(0)
+
+	for _, sensor := range a.Sensors {
+		resp, err := a.client.GetAirQuality(ctx, &traits.GetAirQualityRequest{Name: sensor})
+
+		if err != nil {
```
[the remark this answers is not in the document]
dloveen-v: Good point. Showing some of the data is better than erroring out each time. Thanks! :)

pkg/auto/wordpress/job/energy.go
deanredfern26: That's a fair point but based on how the Job configs are set up, if you dynamically changed the url, ticker duration, endpoints etc. in the auto config you would have to restart the auto to capture those changes anyway, so I don't think it matters too much to restart to capture device changes either in this case

pkg/auto/wordpress/job/energy.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		records, err := getAllRecords(ctx, e.client.ListMeterReadingHistory, meter, now, e.GetInterval())
+
+		if err != nil {
+			e.Logger.Error("getting all records", zap.String("meter", meter), zap.Error(err))
+			continue
+		}
+
+		consumption += processMeterRecords(multiplier, records)
```
mattnathan, line 55: This _get all, then process_ is pretty inefficient wrt memory considering you only really need the first/last record within the range to compute the consumption.

pkg/auto/wordpress/auto.go
```diff
@@ -0,0 +1,95 @@
...
+
+	go func() {
+		mulpx := job.Multiplex(jobs...)
+
+		// tear down
+		defer func() {
+			for _, jb := range jobs {
+				jb.GetTicker().Stop()
+			}
+			close(mulpx.Done)
```
mattnathan, line 78: It's the job of the chan _owner_ to close it, in this case that might just look like adding a `Close` method to Muplx.

mattnathan: Thinking about it some more maybe even adding a Stop to Job would also work

pkg/auto/wordpress/job/meter.go
```diff
@@ -0,0 +1,74 @@
...
+		timeJ := j.GetRecordTime().AsTime()
+
+		if timeI.Before(timeJ) {
+			return 1
+		}
+		if timeI.After(timeJ) {
+			return -1
+		}
+		return 0 // If times are equal
+	})
```
mattnathan, line 67: I'm pretty sure that the history api already returns records in RecordTime order. Even if it doesn't, you should need to sort (`O(nlogn)`) as you're just looking for the min/max which only required a loop.

example/config/app.conf.json
```diff
@@ -98,11 +98,82 @@
           "membership": {"subsystem": "metering"},
           "appearance": {"title": "PV Array"},
           "location": {"title": "Roof", "floor": "Roof"}
+        },
+        {
+          "name": "smart-core/meters/03",
+          "traits": [{"name": "smartcore.bos.Meter", "more": {"unit":  "litres"}}],
```
mattnathan, line 104: Interesting approach to this. I expect we'll need something else for more complicated mock setup, but it works 👍

example/config/app.conf.json
```diff
@@ -98,11 +98,82 @@
...
+          "interval": "10s",
+          "sensors": ["smart-core/iaq/01"]
+        },
+        "water": {
+          "path": "water",
+          "interval": "10s",
+          "meters" : ["smart-core/meters/03"]
+        }
+      }
+    },
```
mattnathan, line 176: I don't think this test belongs in the core example config. Have a look at some of the email autos, they have custom testing tools written to do this kind of thing

mattnathan (changes requested):

pkg/auto/wordpress/config/root.go
```diff
@@ -0,0 +1,89 @@
...
+
+type Authentication struct {
+	Type       string `json:"type"`
+	SecretPath string `json:"secretFile"`
+	Token      string `json:"-"`
+}
+
+type Source struct {
+	Path     string             `json:"path"`
+	Interval jsontypes.Duration `json:"interval"`
```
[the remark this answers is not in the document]
dloveen-v: Good shout. I'll add protection in the `NewTicker` constructor to protect against panics

pkg/auto/wordpress/job/energy.go
dloveen-v: This is true. I wasn't familiar with the database API in terms of meter sampling frequency. Ideally, I just submit 2 ranges and take first and last for each range and subtract, or is there an easier way to do first/last in the listMeterReading call?

mattnathan: There isn't, unfortunately. However you can get away with only keeping one page in memory at a time instead of all the records in range.

example/config/app.conf.json
dloveen-v: I have updated with a tool package for the wordpress automation

pkg/auto/wordpress/job/meter.go
```diff
@@ -32,43 +32,32 @@ func getAllRecords(ctx context.Context, historyFn listMeterReadingFn, meter stri
...
-		}
-		if timeI.After(timeJ) {
-			return -1
-		}
-		return 0 // If times are equal
-	})
-
-	latest := records[0]
-	earliest := records[len(records)-1]
+func processMeterRecords(multiplier float32, records [2]*gen.MeterReadingRecord) float32 {
```
mattnathan, line 58: Similarly this should accept `earliest, latest *gen.MeterReadingRecord` instead of an array

mattnathan (changes requested):
Looking pretty good, nice tests too

pkg/auto/wordpress/job/meter.go
```diff
@@ -14,7 +14,7 @@ import (
 
 type listMeterReadingFn func(ctx context.Context, in *gen.ListMeterReadingHistoryRequest, opts ...grpc.CallOption) (*gen.ListMeterReadingHistoryResponse, error)
 
-func getRecordsByTime(ctx context.Context, historyFn listMeterReadingFn, meter string, now time.Time, filterTime time.Duration) ([2]*gen.MeterReadingRecord, error) {
+func getRecordsByTime(ctx context.Context, historyFn listMeterReadingFn, meter string, now time.Time, filterTime time.Duration) (*gen.MeterReadingRecord, *gen.MeterReadingRecord, error) {
```
mattnathan, line 17: Repeated types in return should be named, otherwise you have no idea what they are.

mattnathan (changes requested):

mattnathan (approved):

PR #278, opened 2024-09-24
app/server: fix intermittent test failures

dloveen-v:
Insert this in line 49:
```go
	serveErr := make(chan error, 1)
	go func() {
                 time.Sleep(time.Second)
		err := ServeHTTPS(ctx, server, time.Second, zap.NewNop())
		if !errors.Is(err, context.Canceled) {
			t.Error(err)
		}
		serveErr <- err
	}()
```

and the test hangs forever due to TLS not being configured with a timeout.

### DIFF
--- a/pkg/app/servers_test.go
+++ b/pkg/app/servers_test.go
@@ -53,6 +53,10 @@ func TestServeHTTPS(t *testing.T) {
 		serveErr <- err
 	}()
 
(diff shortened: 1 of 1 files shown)


hexaglow (approved):
I don't have a good way to verify this, but I trust that it prevents the condition you identified.

I think I will have another look at this test, to see if there are ways to make it more robust.

PR #281, opened 2024-09-25
auto/lights: add ability to retry if updating brightness fails

dloveen-v:
auto/lights: add tests for retrying when updateBrightness fails

I changed the constants to be part of the config.
I also added config options for retry-ability in the auto/lights package. Updated some tests to verify this.

### DIFF
--- a/pkg/auto/lights/auto.go
+++ b/pkg/auto/lights/auto.go
@@ -173,8 +173,6 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 		cancelTtlTimer()
 		cancelRetryTimer()
 
-		var after time.Duration
-
 		ttl, err := processState(ctx, readState, writeState, actions, b.logger.Named("Process State"))
 		if err != nil {
 			// if the context has been cancelled, stop
@@ -184,7 +182,7 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 
 			// if the context remains live, schedule another update soon
 			retryCounter++
-			after = backoffutils.JitterUp(time.Duration(retryCounter)*readState.Config.OnProcessError.BackOffMultiplier.Duration, 0.2)
+			after := backoffutils.JitterUp(time.Duration(retryCounter)*readState.Config.OnProcessError.BackOffMultiplier.Duration, 0.2)
 
 			b.logger.Error("processState failed; scheduling retry",
 				zap.Error(err),
@@ -193,6 +191,10 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 
 			if retryCounter > readState.Config.OnProcessError.MaxRetries {
 				retryCounter = 0
+				// reset retries as new valid state received
+				if !cancelTtlTimer() {
+					<-ttlExpired
+				}
 			} else {
 				ttl = after
 			}
@@ -201,11 +203,9 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 		// ensure it's not too long before we wake up, so the lights are refreshed regularly
 		// so external changes don't stick around forever
 		if ttl < 0 {
-			b.logger.Warn("ttl < 0; using refreshEvery instead")
 			ttl = readState.Config.RefreshEvery.Duration
 		}
 		if ttl > readState.Config.RefreshEvery.Duration {
-			b.logger.Warn("ttl > refreshEvery; using refreshEvery instead")
 			// b.logger.Debug("waking up sooner to ensure lights aren't stale",
 			// 	zap.Duration("after", refreshEvery))
 			ttl = readState.Config.RefreshEvery.Duration
--- a/pkg/auto/lights/auto.go
+++ b/pkg/auto/lights/auto.go
@@ -170,7 +170,6 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 	writeState := NewWriteState(time.Now())
 
 	processStateFn := func(readState *ReadState) error {
-		cancelTtlTimer()
 		cancelRetryTimer()
 
 		ttl, err := processState(ctx, readState, writeState, actions, b.logger.Named("Process State"))
@@ -190,8 +189,8 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 			)
 
 			if retryCounter > readState.Config.OnProcessError.MaxRetries {
-				retryCounter = 0
 				// reset retries to prevent too many repeated attempts
+				retryCounter = 0
 				if !cancelTtlTimer() {
 					<-ttlExpired
 				}
--- a/pkg/auto/lights/auto.go
+++ b/pkg/auto/lights/auto.go
@@ -191,7 +191,7 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 
 			if retryCounter > readState.Config.OnProcessError.MaxRetries {
 				retryCounter = 0
-				// reset retries as new valid state received
+				// reset retries to prevent too many repeated attempts
 				if !cancelTtlTimer() {
 					<-ttlExpired
 				}
--- a/pkg/auto/lights/auto.go
+++ b/pkg/auto/lights/auto.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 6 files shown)


deanredfern26 (approved):

pkg/auto/lights/auto.go
```diff
@@ -181,25 +185,32 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 			}
 
 			// if the context remains live, schedule another update soon
-			// TODO: make this delay configurable, or add backoff etc.
-			after := 5 * time.Second
+			b.retries++
+			after := time.Duration(b.retries) * b.backoffMultiplier
```
hexaglow, line 189: We already use a [backoff package](https://pkg.go.dev/github.com/cenkalti/backoff/v4) in sc-bos, which can help with things like randomised exponential backoff.
You could consider using it here, although a manual implementation is fine too.

Other things I can see:
 - If `b.retries == 0`, then `ttl` will be set to 0, and the retry will be immediate. I think there should be at least a short delay.

pkg/auto/lights/auto.go
```diff
@@ -181,25 +185,32 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
...
+			b.logger.Warn("ttl < 0; using refreshEvery instead")
+			ttl = b.refreshEvery.Load()
+		}
+		if ttl > b.refreshEvery.Load() {
+			b.logger.Warn("ttl > refreshEvery; using refreshEvery instead")
 			// b.logger.Debug("waking up sooner to ensure lights aren't stale",
 			// 	zap.Duration("after", refreshEvery))
-			ttl = refreshEvery
+			ttl = b.refreshEvery.Load()
 		}
```
hexaglow, line 212: Careful here - The first `b.refreshEvery.Load()` could return a different value to the second one, causing an unintended value to be stored in `ttl`. Load the value only once.

pkg/auto/lights/auto.go
```diff
@@ -181,25 +185,32 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
...
 				zap.Duration("retryAfter", after),
 			)
 			ttl = after
 		}
 
-		if ttl < 0 {
-			b.logger.Warn("ttl < 0; using refreshEvery instead")
+		if err == nil || b.retries > b.maxRetries {
+			b.retries = 0
 		}
```
hexaglow, line 199: By the name, I would expect that `maxRetries` is intended to limit the number of time a retry happens. 
But when `retries` exceeds `maxRetries`, all that happens is `retries` is set back to 0, which will reset the backoff duration calculated above but the retry will still happen.

pkg/auto/lights/patch.go
```diff
@@ -157,6 +157,11 @@ type source struct {
 
 func (b *BrightnessAutomation) processConfig(ctx context.Context, cfg config.Root, sources []*source, changes chan<- Patcher) (sourceCount int) {
 	logger := b.logger.With(zap.String("auto", cfg.Name))
+
+	b.refreshEvery.Store(cfg.RefreshEvery.Duration)
+	b.maxRetries = cfg.OnProcessError.MaxRetries
+	b.backoffMultiplier = cfg.OnProcessError.BackOffMultiplier.Duration
```
hexaglow, line 163: Consider what happens if any of these are 0 or negative.
There should be some bounds checking; there is already some validation in `config.Read` which could be extended for example.

hexaglow (changes requested):
The config values can be read more simply from `ReadState`, so they don't need to be manually stored inside `BrightnessAutomation`. That also means no synchronisation required.

There's a bit of concurrency code that's not quite right.

Values read from config need to be validated, because the algorithm won't handle zero/negative values well.

dloveen-v:
Hey @hexaglow thanks for the review.

I wanted to raise a few concerns I had:
> retries is only used inside processStateChanges, so it can be a local variable in there.
I placed it as a field as it seems the convention for this package is to permit tests intercepting the `BrightnessAutomation` states and fields. So if we wanted to test a state at a point in time, this is permitted as a field, but not as a local variable.

> We already use a [backoff package](https://pkg.go.dev/github.com/cenkalti/backoff/v4) in sc-bos, which can help with things like randomised exponential backoff.
My understanding of exponential backoff is it tends to be required in heavy or slow operations where receiving 2 requests simultaneously may be unnecessary or have side-effects. Otherwise, constant or linear backoff is fine. Here, the error is in setting the brightness level of a device, which I imagine is a small network request that doesn't necessarily take a while. It's just that the request _may_ fail sometimes. So linear backoff is sufficient? Again, my understanding of devices and how they work is fairly limited.

> But when retries exceeds maxRetries, all that happens is retries is set back to 0, which will reset the backoff duration calculated above but the retry will still happen.
Retries only leads to a linear back-off delay when there is a device error in setting brightness levels. So it sets to 0 and is unused until there is an error.
The intended flow is:
```
updateBrightness() -> succeed
...
updateBrightness() -> error
  retry = 1, backoff in linear
  retry = 2, backoff in linear
  ...
  retry = x, backoff in linear
updateBrightness() -> succeed
set retry = 0
```

Alternatively, a new state received resets retry = 0 too, to avoid replaying old states. I'll add a test for this too.

I agree with the other points around validating the config. I think there is a high likelihood of errors in the config we write or push via the conductor/. So I'll add the validation and reduce the number of duplicated struct members ( as opposed to copying the values into the `BrightnessAutomation` struct, like I did ).

hexaglow:
@dloveen-v I'm going to copy your replies into the individual comment threads. If you could reply to each individually, that would help me keep track of the progress on each.
Discussions about the PR in general can stay in the top-level thread.

pkg/auto/lights/auto.go
hexaglow: > My understanding of exponential backoff is it tends to be required in heavy or slow operations where receiving 2 requests simultaneously may be unnecessary or have side-effects. Otherwise, constant or linear backoff is fine. Here, the error is in setting the brightness level of a device, which I imagine is a small network request that doesn't necessarily take a while. It's just that the request may fail sometimes. So linear backoff is sufficient? Again, my understanding of devices and how they work is fairly limited.

Some of the uses of this automation are with lights controlled over the DALI protocol, which is really slow (~100ms / operation). With one automation, that's no problem. But with dozens of automations controlling lights on the same DALI bus, if there's no randomisation then the retries for all will tend to happen at the same time, which would cause a lot of bus congestion.

hexaglow:
@dloveen-v I see you marked this as a draft. Do you still want a review now?

pkg/auto/lights/patch.go
```diff
@@ -157,6 +157,11 @@ type source struct {
 
 func (b *BrightnessAutomation) processConfig(ctx context.Context, cfg config.Root, sources []*source, changes chan<- Patcher) (sourceCount int) {
 	logger := b.logger.With(zap.String("auto", cfg.Name))
+
+	if cfg.OnProcessError.BackOffMultiplier.Duration.Nanoseconds() <= 0 {
+		cfg.OnProcessError.BackOffMultiplier.Duration = config.DefaultBackOffMultiplier
+	}
```
hexaglow, line 163: Need the same kind of check for `cfg.RefreshEvery` - don't want this to be negative or 0 either.

pkg/auto/lights/config/root.go
```diff
@@ -32,16 +34,28 @@ type Root struct {
...
 	// Now returns the current time. It's configurable for testing purposes, typically for testing the logic.
 	Now func() time.Time `json:"-"`
 }
 
+// OnProcessError if error is encountered during processing of a state, we set ttl to these defaults with option for backOff
+type OnProcessError struct {
+	// BackOffMultiplier is the wait duration multiplier for each iteration
+	BackOffMultiplier jsontypes.Duration `json:"backOffMultiplier,omitempty"`
+	MaxRetries        int                `json:"maxRetries,omitempty"`
+}
```
hexaglow, line 51: The comments here are hard to understand for someone who is not familiar with the internals of the automation.
Someone who is trying to configure the automation might wonder "what is ttl?" and "for each iteration of what?"

Please describe what these config options do to the behaviour of the automation.

From what I can tell by reading the code, when the number of failed attempts reaches MaxRetries, the automation will stop retrying until the state changes again. That seems good to me; we need a comment saying what the behavior is supposed to be.

hexaglow (changes requested):
The behavior looks good now.

A couple of things that need tidying up.

PR #283, opened 2024-09-26
SC-592: Add dynamic proxying support

hexaglow:
Derived from the original PR #255 

This PR adds the infrastructure and code to allow dynamic proxying to occur. Dynamic proxying is the ability to proxy requests for RPCs that the proxy doesn't know about in advance.

- The router package allows us to serve RPCs using using a dynamic routing table, based on service name and device name.
- The reflectionapi package allows us to change the advertised types the grpc server hosting the gateway returns from its reflection api
- The controller has new hooks to allow systems to adjust the above at runtime
- The node package has API changes to allow dynamically registering both routed and unrouted services
- The gateway system has been rewritten to allow the use of reflection to discover and proxy RPCs

Implementation Notes

The new proxy system code is organised into two stages: scan and announce.

The scan stage performs a number of RPCs against the hub and enrolled nodes to discover the names and types associated with each node. Everything it discovers is placed into the structs defined in model.go.

The announce stage watches the model and updates the node with names and metadata, and announces the discovered services and devices on the Node.


hexaglow:
Made a new PR because I couldn't request review from @mattnathan as he opened the original PR

hexaglow:
Current (major) problem is that because we deserialise request messages into an `emptypb.Empty`, all interceptors that rely on access to request fields are broken. This includes the security policies.

hexaglow:
Fixed the interceptor problem by using `dynamicpb` instead of `emptypb` so field access by name is available.

pkg/system/gateway/announce.go
```diff
@@ -0,0 +1,398 @@
...
+	defer r.mu.Unlock()
+	r.m[k]--
+	if r.m[k] <= 0 {
+		delete(r.m, k)
+	}
+	return r.m[k]
+}
+
+// fullNameToRpcPath maps proto full names (package.Service.Method) to rpc paths (/package.Service/Method).
+func fullNameToRpcPath(fullName protoreflect.FullName) string {
```
mattnathan, line 393: This looks to be unused

proto/devices.proto
```diff
@@ -21,6 +21,8 @@ service DevicesApi {
 message Device {
   string name = 1;
   smartcore.traits.Metadata metadata = 2;
+  // The name of the node that this device is announced from.
+  string node = 3;
```
mattnathan, line 25: I don't think this change is related and deserves some more thought to implement properly as a "where did it come from" concept. I can't see anything in the rest of the PR that requires this, can you remove it please?

mattnathan (changes requested):

proto/devices.proto
hexaglow: Yep, added this by mistake

internal/router/stream.go
```diff
@@ -0,0 +1,216 @@
...
+			msgCap := &captureMsgRecver{MsgRecver: src}
+			client, err = resolver(msgCap)
+			if err != nil {
+				return err
+			}
+			if msg, ok := msgCap.msg.(proto.Message); ok {
+				// this case means the resolver read a message to do the resolve
+				m = msg
+			} else if msgCap.msg == nil {
+				// this case means no message was read, so we need to do it ourselves
```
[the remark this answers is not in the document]
hexaglow: Not sure I understand, the comment is already in the negative?

mattnathan (approved):

PR #285, opened 2024-10-02
Feature/seeder tool

dloveen-v:
Add a tool for seeding database given time parameters.

Add to app.conf.json more history automations so UI graphs are populated.


cmd/tools/db-seeder/main.go
```diff
@@ -0,0 +1,205 @@
...
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"path"
+	"sync"
+	"time"
+
+	"github.com/jackc/pgx/v5/pgxpool"
```
hexaglow, line 12: The rest of sc-bos uses pgx v4, is there a reason this can't also use the same version?

cmd/tools/db-seeder/main.go
```diff
@@ -0,0 +1,205 @@
...
+	"github.com/smart-core-os/sc-golang/pkg/trait"
+	"github.com/vanti-dev/sc-bos/pkg/app/appconf"
+	"github.com/vanti-dev/sc-bos/pkg/app/sysconf"
+	"github.com/vanti-dev/sc-bos/pkg/driver/alldrivers"
+	mockConfig "github.com/vanti-dev/sc-bos/pkg/driver/mock/config"
+	airqualityConfig "github.com/vanti-dev/sc-bos/pkg/zone/feature/airquality/config"
+	occupancyConfig "github.com/vanti-dev/sc-bos/pkg/zone/feature/occupancy/config"
+
+	"github.com/vanti-dev/sc-bos/pkg/zone/allzones"
+	metersConfig "github.com/vanti-dev/sc-bos/pkg/zone/feature/meter/config"
```
hexaglow, line 26: Convention is that package (or import) names should be lowercase, ideally single words (but two is OK) without underscores. https://go.dev/doc/effective_go#package-names

cmd/tools/db-seeder/pkg/seeder/air_quality/air_quality.go
```diff
@@ -0,0 +1,65 @@
...
+			ParticulateMatter_1:      &particulate1,
+			ParticulateMatter_10:     &particulate10,
+			ParticulateMatter_25:     &particulate25,
+			AirChangePerHour:         &airChange,
+		})
+
+		if err != nil {
+			return err
+		}
+		cmd, err := db.Exec(ctx, `INSERT INTO history (source, create_time, payload) VALUES ($1, $2, $3)`, source, current.Format(time.RFC3339Nano), payload)
```
hexaglow, line 52: I'm not a big fan of embedding these queries directly into the tool - it couple the tool to the database schema.
Have you considered extending the `history/pgxstore` package to add what you need?

(same applies to the `air_quality.Seed` and `meter.Seed` too)

cmd/tools/db-seeder/go.mod
```diff
@@ -0,0 +1,71 @@
+module sc-bos-db-seeder
```
hexaglow, line 1: This is not a valid module name; it would have to be `github.com/vanti-dev/sc-bos/cmd/tools/db-seeder` (but it shouldn't be a module anyway)

hexaglow (changes requested):
The functionality looks good. I can see it works with the `vanti-ugs` example configuration.

Please add documentation about what the tool is for and how to use it, in a README.

The directory structure does not match the existing tools in the repo.

- The tool should be part of the same module as everything else, not in its own module.
- Don't add nested packages under a main package. The code here is small enough that it can all live in the main package. The `cmd/` tree should only contain `main` packages. Other packages live under `internal/` or `pkg/`.

For the tools, we assume the user has a working Go compiler and knows 
how to compile/run the program. So I don't think the build scripts and 
Dockerfiles are very useful. Is there a special reason why they are here
for this tool, considering we don't have them for the other tools?

cmd/tools/db-seeder/go.mod
dloveen-v: Hi this was script to run locally until @mattnathan wanted it as part of the tools library. Between setting up and seeding the laptop for the smartshow and getting that to work, this PR was created to review any issues with seeding but wasn't intended to be merged into the main branch, originally. This is why there are issues with the module naming.

As discussed this morning, this functionality will be part of a separate repository for running a demo. 

I plan to have a CI that will run when merging to main in sc-bos, to push to a hosted backend/frontend, but also build MacOS and Windows executables and publish them on slack so anyone is able to download from anywhere to any laptop.

cmd/tools/db-seeder/pkg/seeder/air_quality/air_quality.go
dloveen-v: I tried but history/pgxstore does it's own `RecordTime` values for create time and I didn't have time to go into that functionality and edit it. Writing my own query was MVP.

hexaglow: I think the way to put this into `pgxstore` would be:
- Change `pgxstore.New`, `pgxstore.SetupStoreFromPool` and `pgxstore.NewStoreFromPool` to return a `*pgxstore.Store` instead of a `history.Store` (I think this is a good thing to do anyway - it's idiomatic in Go to [accept interfaces as arguments, and return structs](https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8))
- Add a method `Insert(ctx context.Context, at time.Time, payload []byte) (history.Record, error)` to `pgxstore.Store` that allows specifying the time for the new record.
  - `Append` could then be trivially reimplemented in terms of `Insert`.

That way, you can specify the time and the SQL is still encapsulated in that package.

PR #286, opened 2024-10-15
Remove references to private nexus from docs & packages

deanredfern26:
We shouldn't be referencing nexus in sc-bos now except for the check in the workflow that we aren't referencing it


dloveen-v (approved):
I have verified we don't need the commit hashes for dependencies in the `yarn.lock` file. The tilde or caret should suffice. 

LGTM. Sorry I missed those workflows in the first instance.

PR #288, opened 2024-11-01
zone: add openclose trait support to zones

mattnathan:
Fixes: SC-905


mattnathan:
The question I have is whether a zone should add semantic meaning to the devices in it... should this be `shades` rather than `openClose` - especially as doors and windows also implement OpenClose traits.

hexaglow (approved):

hexaglow:
@mattnathan I think this is fine because the other features are also trait-based. You can use convention to name the openclose groups I suppose.

It would be nice to encode the semantic information directly in the config. But if you had a `blinds` field then that wouldn't help us when we have two kinds of blinds, so you'd still need the groups concept.

PR #289, opened 2024-11-01
driver/bacnet: add configurable emergency based on point value

deanredfern26:
Add the ability to set an emergency when a given bacnet point reads the configured value, intended for binary alarm states, fire alarm, water level low etc

### DIFF
--- a/pkg/driver/bacnet/merge/emergency.go
+++ b/pkg/driver/bacnet/merge/emergency.go
@@ -141,12 +141,12 @@ func (t *emergencyImpl) pollPeer(ctx context.Context) (*traits.Emergency, error)
 		requestNames = append(requestNames, "alarmConfig")
 		readValues = append(readValues, t.config.AlarmConfig.ValueSource)
 		resProcessors = append(resProcessors, func(response any) error {
-			enum, err := comm.EnumValue(response)
+			value, err := comm.IntValue(response)
 			if err != nil {
 				return comm.ErrReadProperty{Prop: "alarmConfig", Cause: err}
 			}
 
-			if t.config.AlarmConfig.OkValue != int(enum) {
+			if int64(t.config.AlarmConfig.OkValue) != value {
 				data.Reason = t.config.AlarmConfig.AlarmReason
 				data.Level = traits.Emergency_EMERGENCY
 			} else {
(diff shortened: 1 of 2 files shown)


pkg/driver/bacnet/merge/emergency.go
```diff
@@ -21,9 +21,18 @@ import (
 	"github.com/vanti-dev/sc-bos/pkg/task"
 )
 
+// AlarmConfig allows configuring a specific bacnet point to become an Emergency if the
+// value read from that point matches the ActivePolarity.
+type AlarmConfig struct {
+	config.ValueSource
+	ActivePolarity int    `json:"activePolarity"` // e.g. 1 if the alarm is active high, 0 if the alarm is active low
+	AlarmReason    string `json:"alarmReason"`    // the reason of the alarm
+}
```
mattnathan, line 30: I like this, however may I suggest a few tweaks:
1. I think it would be better to model the OK case rather than the Emergency/Active case - especially with BACnet there's usually a single OK state and anything else is a bad state.
2. With the above in mind, I think the name of the type (and config field) should be something like `Value` or `Match`. It'd be nice to be able to support matching against Binary and Analog values: "read the val, if it's anything other than 1 then EMERGENCY" kind of thing.

mattnathan (changes requested):

mattnathan:
As an aside, the Emergency trait is supposed to model critical life safety (like there's a fire, evacuate) rather than system alerts (like water level low, go check it out).

pkg/driver/bacnet/merge/emergency.go
```diff
@@ -128,6 +137,26 @@ func (t *emergencyImpl) pollPeer(ctx context.Context) (*traits.Emergency, error)
 		})
 	}
 
+	if t.config.AlarmConfig != nil {
+		requestNames = append(requestNames, "alarmConfig")
+		readValues = append(readValues, t.config.AlarmConfig.ValueSource)
+		resProcessors = append(resProcessors, func(response any) error {
+			enum, err := comm.EnumValue(response)
```
mattnathan, line 144: I can't help thinking this is going to surprise us later. There's no indication from the config names (or doc) that we expect the value of AlarmConfig to be an enum type. What's the reason for this? Can we rename AlarmConfig to `Enum` or something? Can we loosen this line so it can read analog or binary values as well as enums? etc.

mattnathan (changes requested):

mattnathan (approved):

PR #293, opened 2024-11-06
node,proxy: fix incorrect 'owner node' when using proxy.skipChild

mattnathan:
Fix an issue that had symptoms like 'name not found' when it should exist.

The root cause was the proxy drivers `skipChild` feature not working since the reworking of the node package to support ClientConnInterface (instead of generated service wrappers/routers). The change was that metadata was always added for all names, which resulted in those named being queryable via the DevicesApi. The DevicesApi is used by both gateways and proxies, and having authoritative devices in that list (when we didn't want them) could cause unexpected issues.

In summary, when a device name was discovered in more than one place, the routing could break causing not-found issues for some apis but still have discovery work.

This change makes it possible again to announce a name without it appearing in the DevicesApi. The proxy driver is updated to make use of this feature.

Fixes SC-908
Fixed SC-909


deanredfern26 (approved):
I have tested this at OCW and it has fixed the issue

PR #294, opened 2024-11-11
driver/mock,zone/openclose: add support for open close presets

mattnathan:
We now support configuring presets for the mock driver and as part of a zone.

Fixes SC-920

### DIFF
--- a/pkg/zone/feature/openclose/group_test.go
+++ b/pkg/zone/feature/openclose/group_test.go
@@ -57,8 +57,8 @@ func Test_mergeOpenClosePositions(t *testing.T) {
 		// preset tests
 		{"[{open}]", []value{valPres("open")}, ocpPres("open"), false},
 		{"[{nil },{open},{nil }]", []value{valPres(""), valPres("open"), valPres("")}, ocpPres("open"), false},
-		{"[{open},{open},{open}]", []value{valPres(""), valPres("open"), valPres("")}, ocpPres("open"), false},
-		{"[{shut},{open},{nil }]", []value{valPres("closed"), valPres("open"), valPres("")}, ocpPres(""), false},
+		{"[{open},{open},{open}]", []value{valPres("open"), valPres("open"), valPres("open")}, ocpPres("open"), false},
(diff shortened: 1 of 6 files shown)


pkg/zone/feature/openclose/group_test.go
```diff
@@ -54,6 +54,11 @@ func Test_mergeOpenClosePositions(t *testing.T) {
 		{"[{10% held},{10% held}]", []value{val(pos(10).Held()), val(pos(10).Held())}, ocp(pos(10).Held()), false},
 		// combine it all together
 		{"[{10% up slow,90% down},{30% up,70% down held}]", []value{val(up(10).Slow(), down(90)), val(up(30), down(70).Held())}, ocp(up(20).Slow(), down(80).Held()), false},
+		// preset tests
+		{"[{open}]", []value{valPres("open")}, ocpPres("open"), false},
+		{"[{nil },{open},{nil }]", []value{valPres(""), valPres("open"), valPres("")}, ocpPres("open"), false},
+		{"[{open},{open},{open}]", []value{valPres(""), valPres("open"), valPres("")}, ocpPres("open"), false},
```
hexaglow, line 60: This test has the same input as the one above. Looks like it should be `[]value{valPres("open"), valPres("open"), valPres("open")}` to match the name

pkg/zone/feature/openclose/group_test.go
```diff
@@ -54,6 +54,11 @@ func Test_mergeOpenClosePositions(t *testing.T) {
...
+		{"[{shut},{open},{nil }]", []value{valPres("closed"), valPres("open"), valPres("")}, ocpPres(""), false},
```
hexaglow, line 61: Tiny nitpick - use `closed` in the test name instead of `shut` to be consistent with the tested value.

hexaglow (changes requested):
Implementation looks fine.

A couple of minor issues in the zone tests - look like they could be copy/paste mistakes.

pkg/zone/feature/openclose/group_test.go
mattnathan: Changed the test to `shut` - I used shut because it's the same number of characters as `open` and lined up nicely

hexaglow (approved):

PR #296, opened 2024-11-14
Update vue, pinia & sc-api-grpc-web

deanredfern26:
Upgrade a few dependencies for the OPS UI. I had issues when building the latest main Ops UI from arg-ocw, updating a few dependencies fixed the issue.


mattnathan (approved):
I updated the other ui package.json and the yarn.lock files to match.

PR #300, opened 2024-11-22
auto/azureiot: support loading connection string from file

hexaglow:
To avoid having the connection string (which is a secret) mixed in with other configuration data, support loading the CS from a file instead.

Works much like the existing `GroupKeyFile` support.


deanredfern26 (approved):

hexaglow:
I pushed my branches in the wrong order so Github didn't detect the merge, but it is merged.

PR #302, opened 2024-11-29
app: allow adding to the self-signed SAN list from config

mattnathan:
This allows more flexible hosting infrastructure, including containers accessing host systems and inter-container comms via host names

### DIFF
--- a/internal/util/pki/ca.go
+++ b/internal/util/pki/ca.go
@@ -10,6 +10,10 @@ import (
 	"io"
 	"net"
 	"time"
+
+	"golang.org/x/exp/slices"
+
+	"github.com/vanti-dev/sc-bos/pkg/util/netutil"
 )
 
 // CreateCertificateChain creates a new x509.Certificate using id and pub signed by the leaf and private key of authority.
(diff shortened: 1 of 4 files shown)


internal/util/pki/ca.go
```diff
@@ -82,6 +82,14 @@ func WithAuthority(authority string) CSROption {
 	})
 }
 
+// WithSAN adds the given host or IP to the certificates Subject Alternative Names list.
+// As a convenience, if the hostOrIP has a port it will be stripped before adding.
```
hexaglow, line 86: I would prefer to see this separated out.
Have the caller do `WithSAN(StripPort(foo))` instead of adding this extra logic to `hydrateTemplate`

hexaglow (changes requested):
Suggestion: deduplicate `cert.DNSNames` and `cert.IPAddresses` before generating the cert.
Can just do `slices.Sort` followed by `slices.Compact`

internal/util/pki/ca.go
mattnathan: That would be a breaking change as the existing authority logic strips the ports. Well maybe not breaking, but introducing inconsistencies.

hexaglow: Fair enough, keep it as-is

mattnathan:
> Suggestion: deduplicate `cert.DNSNames` and `cert.IPAddresses` before generating the cert. Can just do `slices.Sort` followed by `slices.Compact`

I considered this, it's not as simple though as net.IP isn't comparable

hexaglow:
Try `slices.SortFunc` with `bytes.Compare` and `slices.CompactFunc` with `bytes.Equal`

hexaglow (approved):

PR #304, opened 2024-12-05
hub/pgxhub: return the error in TestHubNode

dloveen-v:
Screenshot 2024-12-05 at 15 04 54

### DIFF
--- a/pkg/system/hub/pgxhub/server.go
+++ b/pkg/system/hub/pgxhub/server.go
@@ -294,7 +294,7 @@ func (n *Server) TestHubNode(ctx context.Context, request *gen.TestHubNodeReques
 	_, err = client.GetMetadata(ctx, &traits.GetMetadataRequest{})
 	if err != nil {
 		logger.Debug("failed api request", zap.Error(err))
-		return nil, status.Error(codes.Unavailable, fmt.Errorf("failed api request: %w", err).Error())
+		return nil, status.Errorf(codes.Unavailable, "failed api request: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/system/hub/pgxhub/server.go
```diff
@@ -294,7 +294,7 @@ func (n *Server) TestHubNode(ctx context.Context, request *gen.TestHubNodeReques
 	_, err = client.GetMetadata(ctx, &traits.GetMetadataRequest{})
 	if err != nil {
 		logger.Debug("failed api request", zap.Error(err))
-		return nil, status.Error(codes.Unavailable, "failed api request")
+		return nil, status.Error(codes.Unavailable, fmt.Errorf("failed api request: %w", err).Error())
```
mattnathan, line 297: I think you want `status.Errorf(codes.Unavailable, "failed api request: %v", err)`

mattnathan (changes requested):

pkg/system/hub/pgxhub/server.go
dloveen-v: `%v` seemed to generic initially. The proper way to wrap errors is with `%w` but it only works in `fmt.Errorf` calls. I've amended.

mattnathan: That's correct, but you aren't wrapping errors (as status.Error doesn't support that), you're formatting strings.

mattnathan (approved):
Can you make sure the changes are squashed before merging please

PR #307, opened 2024-12-13
devices: allow downloading a CSV of devices and their trait data

mattnathan:
This change introduces a new DevicesApi download function that can be used to retrieve a CSV file of a filtered set of devices (like you get on the devices table) along with selected information gathered from their traits.

The devices table now has a new Download CSV button that will use the current filters (floor, subsystem, etc) as the filter for the device data to download.

The implementation uses two steps:
1. Use gRPC to retrieve a URL where the file can be downloaded from. The gRPC request accepts the various options for what data should be downloaded, like the device query.
2. GET that URL to download the CSV file.

The reason for the two-step process is to allow the browser to stream the CSV download, avoiding the need to collect all the information into memory before constructing the CSV.

While the gRPC step is authenticated, the GET URL step isn't. We need to think of a way to reduce access to this link, either via time or securing the access directly. I suspect adding an expiry and signature to the download url will solve this problem reasonably well.

I think future changes could improve things like:

1. Better filenames
2. More customisation of the output CSV
3. More output formats

Though I think what we have here counts as a decent MVP.

Here's some example output:

image
[devices (8).csv](https://github.com/user-attachments/files/18127153/devices.8.csv)


internal/manage/devices/download.go
```diff
@@ -0,0 +1,343 @@
...
+		case protopath.RootStep:
+			// skip writing the root
+		case protopath.UnknownAccessStep:
+		case protopath.AnyExpandStep:
+		}
+	}
+	return strings.Join(parts, ".")
+}
+
+func encodeCsvField(s string) string {
```
dloveen-v, line 273: Why not make use of `encoding/csv` package here? 

```go
        ..
	w.Header().Set("Content-Type", "text/csv")
	w.Header().Set("Content-Disposition", "attachment;filename=data.csv")
        ...
        writer := csv.NewWriter(w)
        
        for d := range usefulDevices() {
            ...
            err := writer.Write(row)
            // handle non-nil err
            writer.Flush()
            ...
        }
```

mattnathan: Good point, I will change it

internal/manage/devices/download.go
```diff
@@ -0,0 +1,337 @@
...
+					continue
+				}
+				row[index] = v
+			}
+		}
+		if err := csvOut.Write(row); err != nil {
+			return
+		}
+	}
+	csvOut.Flush()
```
dloveen-v, line 204: This isn't a show-stopper, but I'd just point out that `csvOut.Write()` won't write to the underlying http.ResponseWriter until `csvOut.Flush()` is called. For a big list of `usefulDevices()`, this may all reside in memory and cause OOM errors, instead of frequently flushing to the network.

Disclaimer, I'm not familiar with podman memory defaults nor do I know how much memory the Beckoff controllers have and if they handle things like paging to mitigate these failures.

dloveen-v: Oh yeah I'm not sure I was thinking of very extreme cases. Also, I didn't realise that when the buffer limit is reached, it auto flushes. The http ResponseWriter must write headers and status codes too which after looking up, looks to be 132 bytes long. So for every 4096 bytes from the CSV writer to the ResponseWriter, there are 132 bytes in the CSV Writer waiting to be buffered out the network. If each row is around or over 3964 bytes long, then there will be a build up in memory, depending on the loop iteration speed vs the speed of the http ResponseWriter's IO flush. But again, it's relatively insignificant depending on the number of users clicking download simultaneously. A good thought experiment though.

deanredfern26 (commented):
Testing this at OCW, when I click on the download link the GET request becomes `https://10.89.1.112/dl/devices?ddt=ClsKTwopCh1tZXRhZGF0YS5tZW1iZXJzaGlwLnN1YnN5c3RlbRoIbWV0ZXJpbmcKIgoXbWV0YWRhdGEubG9jYXRpb24uZmxvb3IaB0Zsb29yIDESCHRleHQvY3N2`
Which does not work.

`10.89.1.112` is the incorrect address, it looks like the podman interface, the edge gateway is on '172.16.100.11', is there something else I need to change?

internal/manage/devices/download.go
dloveen-v: Actually, maybe headers and status codes are only sent once, and the CSV Writer only writes to the response body. 🤔

mattnathan:
> Testing this at OCW, when I click on the download link the GET request becomes `https://10.89.1.112/dl/devices?ddt=ClsKTwopCh1tZXRhZGF0YS5tZW1iZXJzaGlwLnN1YnN5c3RlbRoIbWV0ZXJpbmcKIgoXbWV0YWRhdGEubG9jYXRpb24uZmxvb3IaB0Zsb29yIDESCHRleHQvY3N2` Which does not work.
> 
> `10.89.1.112` is the incorrect address, it looks like the podman interface, the edge gateway is on `172.16.100.11`, is there something else I need to change?

The host comes from one of these places (first wins):
1. sysconf / listenHttps if that property has a binding interface - i.e. `"listenHttps": "1.2.3.4:8443"` would cause the url to start with `https://1.2.3.4:8443/`
2. sysconf / httpAddr + the port from `listenHttps` if configured
3. The IP of the network interface that advertises as non-loopback (see netutil.OutboundAddr()) + the port from `listenHttps`

Looking at this now I can see that OCW has the `grpcAddr` sysconf property configured but not the `httpAddr` property. I think this is actually the norm so I'll update the config parsing to `httpAddr = grpcAddr` when grpcAddr is explicitly configured and httpAddr is not.

mattnathan:
@deanredfern26 I'd also say that httpAddr should really be the hostname people access the gateway via. Check BSP for an example of this.

internal/manage/devices/download.go
```diff
@@ -0,0 +1,337 @@
...
+				return false
+			}
+			device := &gen.Device{
+				Name:     id,
+				Metadata: item.(*traits.Metadata),
+			}
+			return deviceMatchesQuery(token.Request.Query, device)
+		}),
+	)
+	usefulDevices := func() iter.Seq[*traits.Metadata] {
```
hexaglow, line 83: Why not move the filtering inside the `resource.WithInclude` above? 
We iterate `usefulDevices()` 3 times and perform the same filtering each time.

internal/manage/devices/download.go
```diff
@@ -0,0 +1,337 @@
...
+		return func(yield func(*traits.Metadata) bool) {
+			for _, d := range deviceList {
+				if len(d.Traits) == 0 {
+					continue
+				}
+				if proto.Equal(d, &traits.Metadata{Name: d.Name, Traits: []*traits.TraitMetadata{{Name: string(trait.Metadata)}}}) {
+					continue // skip boring devices
+				}
```
hexaglow, line 91: What's the rationale for hiding these devices?

internal/manage/devices/download.go
```diff
@@ -0,0 +1,337 @@
...
+	if err := s.encodeDownloadUrlToken(&u, tokenStr); err != nil {
+		return nil, err
+	}
+	return &gen.DownloadDevicesUrl{
+		Url: u.String(),
+	}, nil
+}
+
+// DownloadDevicesHTTPHandler responds to HTTP request urls returned by GetDownloadDevicesUrl.
+func (s *Server) DownloadDevicesHTTPHandler(w http.ResponseWriter, r *http.Request) {
```
hexaglow, line 57: This is a big method that combines business logic with HTTP handling.
I would prefer to see a separate `func GenerateDevicesCSV(w io.Writer, devices []*traits.Metadata) error` (or something like that) with unit tests.

internal/manage/devices/download.go
```diff
@@ -0,0 +1,337 @@
...
+
+	tokenData := &DownloadToken{
+		Request: request,
+	}
+	tokenStr, err := encodeDownloadToken(tokenData)
+	if err != nil {
+		return nil, err
+	}
+	u := s.downloadUrlBase
+	if err := s.encodeDownloadUrlToken(&u, tokenStr); err != nil {
```
hexaglow, line 48: These encode/decode functions have very similar names and it's quite confusing. In particular `encodeDownloadUrlToken` which doesn't encode a token like I first thought based on the name, it encodes a URL based on a token which is already encoded. 

Could we merge these two steps? After all, all we really care about is being able to turn a `DownloadToken` into a URL and back again.

hexaglow (changes requested):
I have attached some suggestions for refactoring. 

Before merging, cryptographic authentication and expiry needs implementing for the tokens (which you have identified already)

internal/manage/devices/download.go
mattnathan: Good point, I'll make the change

internal/manage/devices/download.go
mattnathan: Mostly it's to remove rows from the CSV that have no data other than the name of the device - typically these are `automations` and things like that. It might be that future updates add options to include them, but for the cases here they just filled the CSV with (almost) empty rows

hexaglow:
@mattnathan just taken a look at your example CSV, the order of the columns isn't ideal.
I would expect `name` to be column A, but in the example it's in the middle

internal/manage/devices/download.go
mattnathan: I've changed the names to `read/writeDownloadToken` and `downloadTokenFrom/ToString` which I think helps. It makes sense to me to keep them separate to allow easier flow in the error cases.

mattnathan:
> @mattnathan just taken a look at your example CSV, the order of the columns isn't ideal. I would expect `name` to be column A, but in the example it's in the middle

Yeah, I noticed that too. I'm considering prefixing all the metadata bits with `metadata.` (or `md.`) to at least put them all together, then I might implement a custom sort order, but haven't decided if that's worth it or not. Thoughts?

internal/manage/devices/download.go
mattnathan: Have a look at that, I think it's better

mattnathan:
> > @mattnathan just taken a look at your example CSV, the order of the columns isn't ideal. I would expect `name` to be column A, but in the example it's in the middle
> 
> Yeah, I noticed that too. I'm considering prefixing all the metadata bits with `metadata.` (or `md.`) to at least put them all together, then I might implement a custom sort order, but haven't decided if that's worth it or not. Thoughts?

@hexaglow I have pushed the column ordering change now

mattnathan:
For visibility, @hexaglow and I realised that the sign/verify logic we have here only works on a single gateway - signing a dl url on one gateway will fail validation on another. We are going to accept this for now as our credential tokens (for tenant API) also suffer from this problem so a more general fix will be needed to support it, which we'll do later. See SC-960

proto/devices.proto
```diff
@@ -159,3 +161,21 @@ message PullDevicesMetadataResponse {
     google.protobuf.Timestamp change_time = 3;
   }
 }
+
+message GetDownloadDevicesUrlRequest {
+  // A query for filtering the devices that will be returned
+  Device.Query query = 1;
+  // Specifies the media type (or MIME type) of the file to be downloaded.
+  // Defaults to text/csv.
+  // Supported values are text/csv (currently).
```
hexaglow, line 170: Redundant "(currently)", we don't need to state that input parameters may support extra values in the future.

hexaglow (changes requested):
I've tested this manually and it seems to work correctly. This is a nontrivial function and I think it needs some tests. I would like to see the tests check that

- The right devices are included.
- The special column sorting rules work correctly.
- Trait values are correctly converted to columns.
- The token validation is correct:
   - Malformed tokens are rejected
   - Tokens signed with the wrong key are rejected
   - Expired tokens are rejected

The CSV download format is not documented. I would expect to have documentation for this feature that includes:

- What devices will be included in the resulting CSV file (the exclusion of "boring" devices)
- Which traits are supported
- The naming scheme for the columns

mattnathan:
> I've tested this manually and it seems to work correctly. This is a nontrivial function and I think it needs some tests. I would like to see the tests check that
> 
> * The right devices are included.
> * The special column sorting rules work correctly.
> * Trait values are correctly converted to columns.
> * The token validation is correct:
>   
>   * Malformed tokens are rejected
>   * Tokens signed with the wrong key are rejected
>   * Expired tokens are rejected
> 
> The CSV download format is not documented. I would expect to have documentation for this feature that includes:
> 
> * What devices will be included in the resulting CSV file (the exclusion of "boring" devices)
> * Which traits are supported
> * The naming scheme for the columns

As always keeping me to my own standards. I'll work on some unit tests for this

hexaglow (approved):

PR #309, opened 2024-12-19
trait/waste: migrate waste trait to sc-api

deanredfern26:
remove the waste proto and generated files now we have migrated


dloveen-v:
Do you need to change the import paths?

dloveen-v (approved):

PR #312, opened 2025-01-07
node: add AnnounceScope and ReplaceAnnouncer, deprecate AnnounceContext

hexaglow:
The existing AnnounceContext is insufficient for services, because there's no way to wait for previous announcements to be undone before announcing the devices again.
Announcements are then interleaved incorrectly with undos which produces missing metadata etc.

These new types provide a way for a driver to wait for its previous announcements to be undone before it makes new ones.

### DIFF
--- a/pkg/node/announce.go
+++ b/pkg/node/announce.go
@@ -50,43 +50,96 @@ func AnnounceWithNamePrefix(prefix string, a Announcer) Announcer {
 
 // AnnounceContext returns a new Announcer that undoes any announcements when ctx is Done.
 // This leaks a go routine if ctx is never done.
+//
+// Deprecated: AnnounceContext does not allow synchronising with the undos, which is usually necessary.
+// Use AnnounceScope instead, or a ReplaceAnnouncer if multiple generations are required.
 func AnnounceContext(ctx context.Context, a Announcer) Announcer {
-	mu := sync.Mutex{}
-	var undos []Undo
-	undone := make(chan struct{})
-	go func() {
-		defer close(undone)
-		<-ctx.Done()
-		mu.Lock()
-		defer mu.Unlock()
-		for _, undo := range undos {
-			undo()
-		}
-		undos = nil
-	}()
-	return AnnouncerFunc(func(name string, features ...Feature) Undo {
-		select {
-		case <-ctx.Done():
-			return NilUndo
-		default:
-		}
+	scoped, undo := AnnounceScope(a)
+	_ = context.AfterFunc(ctx, undo)
+	return scoped
+}
 
-		undo := a.Announce(name, features...)
-		mu.Lock()
-		defer mu.Unlock()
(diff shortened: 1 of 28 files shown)


pkg/node/announce.go
```diff
@@ -50,43 +50,94 @@ func AnnounceWithNamePrefix(prefix string, a Announcer) Announcer {
...
+	scoped, undo := AnnounceScope(a)
+	_ = context.AfterFunc(ctx, undo)
+	return scoped
+}
+
+// AnnounceScope returns a new scoped Announcer that will undo all announcements made to it when the returned Undo is called.
+// The Undo will block until all undos are complete.
+// Once the Undo is called, the Announcer is no longer valid and any further calls to Announce will do nothing.
+//
+// It is safe to call the Undo more than once - calls after the first will do nothing.
```
mattnathan, line 66: Worth mentioning that any Undos returned by Announce calls can be Undone early and safely if desired?

pkg/node/announce_test.go
mattnathan, line 1: Could do with adding a test for ReplaceAnnouncer. The code will become part of many critical paths.

pkg/system/publications/factory.go
```diff
@@ -29,7 +29,7 @@ func NewSystem(services system.Services) *System {
 	s := &System{
 		logger:    services.Logger.Named("publications"),
 		name:      services.Node.Name(),
-		announcer: services.Node,
+		announcer: node.NewReplaceAnnouncer(services.Node),
```
mattnathan, line 32: Worth standardising on `announcer` vs `announcements` - I see both spellings used by different services. My vote goes for `announcer`

mattnathan (changes requested):

mattnathan:
Also, the PR title no longer reflects the change being made.

pkg/system/publications/factory.go
mattnathan: Saw some `announce` fields too

pkg/node/announce_test.go
```diff
@@ -52,24 +49,104 @@ func TestAnnounceContext(t *testing.T) {
...
+	a1 := ra.Replace(ctx1)
+	a1.Announce("a")
+	ub := a1.Announce("b")
+	a1.Announce("c")
+
+	m.Lock()
+	if diff := cmp.Diff(names, []string{"a", "b", "c"}); diff != "" {
+		t.Errorf("unexpected names (-want +got):\n%s", diff)
+	}
+	m.Unlock()
```
mattnathan, line 95: Small nit pic, I think these tests would read better with the introduction of

```go
checkNames := func(want ...string) {
	t.Helper()
	m.Lock()
	if diff := cmp.Diff(want, names); diff != "" {
		t.Errorf("unexpected names (-want +got):\n%s", diff)
	}
	m.Unlock()
}
```

mattnathan (changes requested):

PR #314, opened 2025-01-09
util/pull: retry pull even when Unimplemented

mattnathan:
pull.OrPoll (and pull.Changes) used to have an assumption that if pull and get both returned errors with codes.Unimplemented then there was no point continuing to retry the requests as this couldn't change. This assumption was violated when we implemented dynamic gRPC apis as part of the dynamic gateway work.

This change adjusts the pull functions to retry even when Unimplemented is returned.

### DIFF
--- a/pkg/util/pull/pull.go
+++ b/pkg/util/pull/pull.go
@@ -15,14 +15,113 @@ import (
 
 type Getter = func(context.Context) error
 
-// OrPoll will attempt to call pull, falling back to poll if it's not supported.
-// pull subscribe to changes, blocking until the subscription fails, if possible failures will be retried.
-// poll should query for the current state, this function manages delays between calls.
-func OrPoll(ctx context.Context, pull, poll Getter, opts ...Option) error {
+// OrPoll calls pull, falling back to repeated calls to get if pull returns a codes.Unimplemented error.
+// Argument pull should subscribe to changes, blocking until the subscription fails.
+// Argument get should get the current state, blocking until the state is got or an error occurs.
+// OrPoll blocks until the context is cancelled, retrying either pull or get on error.
+func OrPoll(ctx context.Context, pull, get Getter, opts ...Option) error {
 	conf := calcOpts(opts...)
 
-	shouldPoll := false
+	return runBlocking(ctx, "blocking tasks", func(ctx context.Context) error {
+		err := runPull(ctx, pull, conf)
+		if shouldReturn(err) {
+			return err
+		}
+		// the runPull err is never nil, so if we're here then we need to try polling
+		err = get(ctx) // get one to see if this is unimplemented too
+		if shouldReturn(err) {
+			return err
+		}
+		if !isUnimplemented(err) {
+			// pull is unimplemented but get is not, there's a chance things have changed (i.e. during boot)
+			// so try pull again just to be sure.
+			pullErr := runPull(ctx, pull, conf)
+			if shouldReturn(pullErr) {
+				return pullErr
+			}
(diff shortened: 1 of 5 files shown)


pkg/util/pull/pull.go
```diff
@@ -15,14 +15,113 @@ import (
...
+	incErr := func() {
+		errCount++
+		pollDelay = time.Duration(float64(pollDelay) * 1.2)
+		if pollDelay > conf.fallbackMaxDelay {
+			pollDelay = conf.fallbackMaxDelay
+		}
+		ticker.Reset(pollDelay)
+		if errCount == 5 {
+			conf.logger.Warn("poll is failing, will try keep retrying", zap.Error(err))
+		}
```
hexaglow, line 92: Maybe use `github.com/cenkalti/backoff/v4` (which we already depend on) instead of manual backoff algorithm.

pkg/util/pull/pull.go
```diff
@@ -15,14 +15,113 @@ import (
...
+		opts...)
+}
+
+// runPull retries calling pull returning on context cancellation or when pull returns an error that has codes.Unimplemented.
+// runPull will never return a nil error.
+func runPull(ctx context.Context, pull Getter, conf changeOpts) error {
+	return runBlocking(ctx, "pulls", pull, conf, isUnimplemented)
+}
+
+// runPoll blocks repeatedly calling get after some delay until context is cancelled or get returns an error that is not codes.Unimplemented.
```
hexaglow, line 75: Comment says that runPoll returns when context is cancelled or a non-Unimplemented error is returned by get. However, the function will return when an Unimplemented error happens (line 108). It returns when the error is a context error or codes.Unimplemented, other errors will retry.

hexaglow (changes requested):
Algorithm looks good, I have some questions about `runPoll`

dloveen-v:
Just an opinion, but a lot of this feels like re-inventing the wheel. gRPC already supports retries depending on the status code errors returned. e.g. https://github.com/grpc/grpc-go/blob/master/examples/features/retry/client/main.go @mattnathan I believe we discussed this briefly when I first joined. I've decided to ask the question why it is beneficial to re-implement retries on specific status codes rather than refactoring to attempt a gRPC pull request with client-side retries configured for `UNIMPLEMENTED` codes?

pkg/util/pull/pull.go
mattnathan: Ooops, that'd be a typo then

pkg/util/pull/pull.go
mattnathan: I'll have a look, I was hoping to keep as much of the code the same as I could. Not sure what backoff's api looks like (or how it'll affect the tests)

mattnathan:
> Just an opinion, but a lot of this feels like re-inventing the wheel. gRPC already supports retries depending on the status code errors returned. e.g. https://github.com/grpc/grpc-go/blob/master/examples/features/retry/client/main.go @mattnathan I believe we discussed this briefly when I first joined. I've decided to ask the question why it is beneficial to re-implement retries on specific status codes rather than refactoring to attempt a gRPC pull request with client-side retries configured for `UNIMPLEMENTED` codes?

This would need some experimentation, but my guess is that our wrappers (and node) don't support the retry options you'd use for this. That's not to say we couldn't add that, but it's not part of this PR

pkg/util/pull/pull.go
mattnathan: This has been done now.

hexaglow (approved):

PR #316, opened 2025-01-10
traits: add security event trait

deanredfern26:
adds more fields to better describe the event and also share the Actor entity with the Access trait


pkg/gen/access_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.3.0
-// - protoc             v5.29.1
+// - protoc-gen-go-grpc v1.5.1
+// - protoc             v5.28.3
```
dloveen-v, line 4: This doesn't look right (for `protoc v5.29.1`). It looks like an accidental downgrade.

dloveen-v (dismissed):
LGTM. I have just noticed the CI has failed in several places. I am trying to revoke my approval until these are fixed.

proto/security_event.proto
```diff
@@ -0,0 +1,168 @@
...
+  State state = 6;
+
+  // Optional. The priority of the security event
+  int32 priority = 7;
+
+  // Optional. The type of the security event
+  EventType event_type = 8;
+}
+
+message ListSecurityEventsRequest {
```
dloveen-v, line 127: Hi, I'm wondering if anything like a security events query field needs adding here. You can see in `devices.proto` it supports filtering via a request query https://github.com/vanti-dev/sc-bos/blob/add-security-event-trait/proto/devices.proto#L85C10-L85C15. 

Not essential, but it sounds like a lot of dependencies on filtering, sorting etc... need supporting for the front end ops UI.

deanredfern26: Yes I think something like this would be a good thing to add, shall we add it as a future PR once we know a little more of what we want the front end to filter by? In the first instance I am just planning on making the UI list all the security events as there will only be one driver implementing this trait

dloveen-v (approved):
LGTM

PR #319, opened 2025-01-14
Add security events page

deanredfern26:
Adds a security event page to display all the security events on the building controller.

Bare minimum functionality for now, just displays the events in decending time order. No filtering, custom sorting nor CSV download.


deanredfern26:
@mattnathan Not sure if this is still on your radar? The security events are live at OCW so if you point the .env to scos-ocw.paradisebuild.uk (172.16.100.11) you can test

mattnathan:
sorry, I hope to get to this tomorrow.

deanredfern26:
> I've been thinking about this and I'd like to know how urgent this feature is. There are plans to refresh the existing security page as part of the SPOG work, and I think adding the log to that page instead of it being its own page might work better. My thoughts aren't 100% clear on this yet, and if we have client requirements pressing for it I'm also happy landing this change then adjusting to a more unified security section later.

This feature will be required for witnessing the upcoming sites, 40LH and HBD Island. I would love to give you a firm date but that won't happen. I agree that a security page needs more TLC, but this mvp impl gives us the ability to proof that we are receiving messages from the security system. Also, this page has already been seen for witnessing so it would not look good if we present the ops UI without this or an alternative already in place

mattnathan:
> This feature will be required for witnessing the upcoming sites

Understood, I'll get a review sorted now we have the mock trait impl

mattnathan:
@deanredfern26 can you rebasethis onto main - there are conflicts and the branch is missing the mock driver changes

deanredfern26:
> @deanredfern26 can you rebasethis onto main - there are conflicts and the branch is missing the mock driver changes

Done

mattnathan:
There's no way to configure which device should be used as the source of security events. The ui is hard coded to the hub name or the connected node name if there is no hub. I'll manually override this in the code for now so I can progress with the review, but it'll need fixing somehow. We should be able to configure this via the ui-config or select it on the ui

ui/yarn.lock
mattnathan, line 1: A lot of changes here that aren't related to this pr - there's no package.json changes so there should be no yarn.lock changes

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+  overviewPage: {
+    type: Boolean,
+    default: false
+  }
+});
+
+const sidebar = useSidebarStore();
+
+const cohort = useCohortStore();
+const name = computed(() => cohort.hubNode?.name ?? '');
```
mattnathan, line 53: We need a way to configure the name that is displayed by the table

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+import {computed, onUnmounted, ref} from 'vue';
+
+const props = defineProps({
+  overviewPage: {
+    type: Boolean,
+    default: false
+  }
+});
+
+const sidebar = useSidebarStore();
```
mattnathan, line 50: I can't see that the sidebar is used for anything with this table, I can't find a way to open the sidebar.

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+import {timestampToDate} from '@/api/convpb';
+import ContentCard from '@/components/ContentCard.vue';
+import {useSecurityEventsCollection} from '@/composables/securityevents.js';
+import {useDataTableCollection} from '@/composables/table.js';
+import {useCohortStore} from '@/stores/cohort.js';
+import {useSidebarStore} from '@/stores/sidebar';
+import {computed, onUnmounted, ref} from 'vue';
+
+const props = defineProps({
+  overviewPage: {
```
mattnathan, line 44: I think we can get rid of this property as it's never used and will simplify the logic in the template.

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+// Calculate the total number of items in the query
+const queryTotalCount = computed(() => {
+  return securityEventsCollection.totalItems.value;
+});
+
+const allHeaders = [
+  {title: 'Timestamp', value: 'createTime', width: '15em'},
+  {title: 'Description', value: 'description', width: '60%'},
+  {title: 'Source', value: 'source.name', width: '30%'},
+  {title: 'Priority', value: 'priority', width: '10em'}
```
mattnathan, line 74: Numerical fields should be right aligned.

mattnathan, line 74: Are there no other headers from the SecurityEvent type that would be useful to see here?

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+];
+
+const rowProps = ({item}) => {
+  if (item.resolveTime) return {class: 'resolved'};
```
mattnathan, line 78: `resolveTime` is not a property on SecurityEvent, not sure what this is supposed to be doing.

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+</script>
+
+<style lang="scss" scoped>
+:deep(table) {
+  table-layout: fixed;
+}
+
+:deep(.resolved) {
+  color: #fff5 !important;
+}
```
mattnathan, line 93: With the current code, this class will never be active as resolved isn't ever set on the row

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+
+.v-data-table :deep(tr:hover) {
+  cursor: pointer;
```
mattnathan, line 96: You can't click on rows, so changing the cursor to a point is misleading

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -0,0 +1,114 @@
...
+
+  :deep(.v-pagination__first) {
+    margin-left: 16px;
+  }
+}
+
+.character-counter {
+  font-size: 75%;
+  color: rgb(var(--v-theme-primary)); /* Adjust color as needed */
+}
```
mattnathan, line 113: I can't see any use for this class.

ui/ops/src/routes/ops/security-events/route.js
```diff
@@ -0,0 +1,18 @@
+export default [
+  {
+    path: 'security-events',
+    components: {
+      default: () => import('./SecurityEventsTable.vue'),
+      sidebar: () => import('./SecurityEventsSideBar.vue')
```
mattnathan, line 6: I don't think this route has a sidebar, I couldn't find a way to open it at least.

ui/ops/src/routes/ops/security-events/route.js
```diff
@@ -0,0 +1,18 @@
...
+    },
+    props: {
+      default: true,
+      sidebar: false
+    },
+    meta: {
+      authentication: {
+        rolesRequired: ['superAdmin', 'admin', 'commissioner', 'operator', 'viewer']
```
mattnathan, line 14: I don't think viewers should be allowed to see security events, though we're going to be changing all this after MR's work anyway so wouldn't worry too much about it.

mattnathan (changes requested):

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
deanredfern26: These are the ones the Gallagher driver implements, which is the only driver which needs the page. I was thinking to add functionality when we integrate with more systems?

mattnathan: Makes sense

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -1,24 +1,23 @@
 <template>
   <div class="ml-2">
-    <v-row v-if="!props.overviewPage" class="mt-0 ml-0 pl-0">
+    <v-row class="mt-0 ml-0 pl-0">
       <h3 class="text-h3 pt-2 pb-6">Security Events</h3>
       <v-spacer/>
     </v-row>
 
-    <content-card :class="['px-8', {'mt-8 px-4': !props.overviewPage}]">
+    <content-card :class="['px-8', 'mt-8 px-4']">
```
mattnathan, line 8: If `class` isn't dynamic, probably best to just write it `class="px-8 mt-8 px-4"` which immediately shows that `px-8` and `px-4` are conflicting styles and should be fixed.

ui/ops/src/routes/ops/security-events/SecurityEventsTable.vue
```diff
@@ -1,24 +1,23 @@
 <template>
   <div class="ml-2">
-    <v-row v-if="!props.overviewPage" class="mt-0 ml-0 pl-0">
+    <v-row class="mt-0 ml-0 pl-0">
       <h3 class="text-h3 pt-2 pb-6">Security Events</h3>
       <v-spacer/>
     </v-row>
```
mattnathan, line 6: This change duplicates the Security Events title when combined with line 20

mattnathan (changes requested):

mattnathan (approved):

PR #323, opened 2025-01-16
driver/bacnet: add tries as a configurable field for update poll tests

dloveen-v:
> It seems updating the set point leads to errors being thrown in the frontend.. even when the temperature does actually get updated... eventually. I am looking at the BacNet driver, it seems the number of tries in the poll + test functionality is set to 5. Does this need changing to pull the tries from the config? Or was it hardcoded for a reason and 5 tries should be enough pointing a deeper issue with BacNet?

### DIFF
--- a/pkg/driver/bacnet/config/traits.go
+++ b/pkg/driver/bacnet/config/traits.go
@@ -30,6 +30,9 @@ type Trait struct {
 	PollPeriod *Duration `json:"pollPeriod,omitempty"`
 	// how long to wait when running a poll for it to respond. Defaults to PollPeriod.
 	PollTimeout *Duration `json:"pollTimeout,omitempty"`
+	// after a write update, how long to wait by default before giving up on reads becoming consistent with the written value.
+	// defaults to 1m
+	DefaultRWConsistencyTimeout *Duration `json:"defaultRWConsistencyTimeout,omitempty"`
 	// When reading multiple properties, split the properties into chunks of this size and execute in parallel.
 	// 0 means do not chunk.
 	ChunkSize int `json:"chunkSize,omitempty"`
@@ -47,6 +50,13 @@ func (t *Trait) PollTimeoutDuration() time.Duration {
 		return t.PollTimeout.Duration
 	}
 	return t.PollPeriodDuration()
+}
+
+func (t *Trait) DefaultRWConsistencyTimeoutDuration() time.Duration {
+	if t.DefaultRWConsistencyTimeout != nil && t.DefaultRWConsistencyTimeout.Duration != 0 {
+		return t.DefaultRWConsistencyTimeout.Duration
+	}
+	return time.Minute
 }
 
 type RawTrait struct {
--- a/pkg/driver/bacnet/merge/poll.go
+++ b/pkg/driver/bacnet/merge/poll.go
@@ -38,17 +38,17 @@ func startPoll(init context.Context, name string, pollDelay, pollTimeout time.Du
 //
 //  1. ctx is done
 //  2. pollPeer returns an error
-//  3. defaultDelay has passed if the ctx doesn't have a non-zero ctx.Deadline assigned
-//     or time.Now() + defaultDelay is earlier than the deadline assigned to ctx
+//  3. timeout has passed if the ctx doesn't have a non-zero ctx.Deadline assigned
+//     or time.Now() + timeout is earlier than the deadline assigned to ctx
 //
 // An backoff delay will be added between each call to pollPeer
-func pollUntil[T any](ctx context.Context, defaultDelay time.Duration, poll func(context.Context) (T, error), test func(T) bool) (T, error) {
+func pollUntil[T any](ctx context.Context, timeout time.Duration, poll func(context.Context) (T, error), test func(T) bool) (T, error) {
 	fail := func(err error) (T, error) {
 		var zero T
 		return zero, err
 	}
 
-	ctx, cancel := context.WithTimeout(ctx, defaultDelay)
+	ctx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
 	var delay time.Duration
--- a/pkg/driver/bacnet/merge/poll.go
+++ b/pkg/driver/bacnet/merge/poll.go
@@ -3,7 +3,6 @@ package merge
 import (
 	"context"
 	"fmt"
-	"math"
 	"time"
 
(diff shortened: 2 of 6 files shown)


hexaglow (dismissed):
As discussed, please limit retries based on total duration, rather than count.

pkg/driver/bacnet/config/traits.go
```diff
@@ -30,6 +30,9 @@ type Trait struct {
 	PollPeriod *Duration `json:"pollPeriod,omitempty"`
 	// how long to wait when running a poll for it to respond. Defaults to PollPeriod.
 	PollTimeout *Duration `json:"pollTimeout,omitempty"`
+	// maximum period for tries for the test during a poll event
+	// defaults to 1m to not waste server resources
+	MaxTriesDelay *Duration `json:"maxTriesDelay,omitempty"`
```
mattnathan, line 35: I don't think the name or comment clearly describe what this property does.

Maybe this is clearer?
```go
// after a write, how long to wait before giving up on reads becoming consistent with the written value.
RWConsistencyTimeout *Duration `json:"rwConsistencyTimeout,omitempty"`
```

pkg/driver/bacnet/merge/poll.go
```diff
@@ -72,15 +70,26 @@ func pollUntil[T any](ctx context.Context, tries int, poll func(context.Context)
...
+	deadline, ok := ctx.Deadline()
+
+	if ok {
+		if deadline.Before(time.Now().Add(maxDelay)) {
+			// the parent context has a deadline earlier than maxDelay would allow
+			return context.WithCancel(ctx)
+		}
+	}
+
+	return context.WithTimeout(ctx, maxDelay)
```
mattnathan, line 94: This code isn't doing anything that `context.WithTimeout(ctx, maxDelay)` isn't already doing, see https://go.dev/play/p/Jf_ew904Uq0

I think what you meant to do was use `deadline.After(...)` as a way to say _if the caller is ok waiting an hour for this instead of our default of a minute, let them_.

pkg/driver/bacnet/merge/poll.go
```diff
@@ -38,18 +37,17 @@ func startPoll(init context.Context, name string, pollDelay, pollTimeout time.Du
 // Returns early with error if
 //
 //  1. ctx is done
-//  2. the number of polls is tries
-//  3. pollPeer returns an error
+//  2. pollPeer returns an error
```
mattnathan, line 40: You're missing `maxDelay time has elapsed`

mattnathan (changes requested):

pkg/driver/bacnet/merge/poll.go
dloveen-v: Oh so we want to allow applying a deadline that is potentially longer than the default, so long as there is a deadline? I must have misunderstood.

dloveen-v: You're right though, `context.WithDeadline()` already looks at which deadline is longer in the documentation. I'll change it

pkg/driver/bacnet/merge/poll.go
dloveen-v: It's given as an argument, so this is a different execution scenario. Thanks

hexaglow:
I don't think it's a good idea to extend the timeout if the RPC supplies a timeout. I find that very surprising and contrary to usual timeout semantics. In my mind, request timeouts are about restricting how long we wait, not extending it. It also doesn't work equivalently to request cancellation signals. I think these two scenarios should be semantically the same:
- Sending a request with a timeout of 5m
- Sending a request with no timeout, waiting 5m, then sending a cancellation signal

dloveen-v:
>I think these two scenarios should be semantically the same:

>Sending a request with a timeout of 5m
>Sending a request with no timeout, waiting 5m, then sending a cancellation signal

This does make sense. I guess the difficulty for me is figuring out what the upper timeout limit for BacNet should be as a default. 1m? 10m?

If I know how long before the update to BacNet is considered a failed request in usual cases, I can change the config to specify this upper limit as a default. Allowing for overrides in the config if necessary.

pkg/driver/bacnet/merge/poll.go
```diff
@@ -38,18 +37,19 @@ func startPoll(init context.Context, name string, pollDelay, pollTimeout time.Du
...
+//  3. defaultDelay has passed if the ctx doesn't have a non-zero ctx.Deadline assigned
+//     or time.Now() + defaultDelay is earlier than the deadline assigned to ctx
 //
 // An backoff delay will be added between each call to pollPeer
-func pollUntil[T any](ctx context.Context, tries int, poll func(context.Context) (T, error), test func(T) bool) (T, error) {
+func pollUntil[T any](ctx context.Context, defaultDelay time.Duration, poll func(context.Context) (T, error), test func(T) bool) (T, error) {
```
mattnathan, line 45: Sorry to be picky, but `delay` reads like things are going to pause for a bit. Can't this just be called `timeout` or something like that?

mattnathan (changes requested):

pkg/driver/bacnet/merge/poll_test.go
```diff
@@ -0,0 +1,163 @@
...
+						// succeed straight away
+						return true
+					}
+				},
+			},
+			wantErr: nil,
+			want:    true,
+		},
+		{
+			name: "deadline before defaultDelay - test succeeds after deadline",
```
mattnathan, line 71: your tests still mention the old name of the argument.

mattnathan (changes requested):

pkg/driver/bacnet/merge/poll_test.go
```diff
@@ -0,0 +1,163 @@
...
+			want:    true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			res, err := pollUntil[B](tt.args.ctx(), tt.args.timeout, poll, tt.args.test())
+
+			if diff := cmp.Diff(res, tt.want); diff != "" {
+				t.Errorf("pollUntil(): (-got +want)\n%s", diff)
```
mattnathan, line 155: Again, sorry I didn't spot this earlier. Standard usage is `cmp.Diff(want, got)` with the message saying `(-want,+got)`

mattnathan (changes requested):

mattnathan (approved):
LGTM

PR #324, opened 2025-01-17
auto/lights: write cache refactoring

mattnathan:
Three changes split across two commits:
1. Refactor the write cache logic into an actions middleware
2. Allow the write cache expiry it to be configured via config
3. Adjust the default values for the write cache

See the individual commit messages for details and reasoning.

As mentioned in [#320](https://github.com/vanti-dev/sc-bos/pull/320#discussion_r1918409806) this work is also one step towards potentially removing state based information in the write state (the Brightness map) as with this change we now have another place write time can be recorded, as part of the middleware instead of in WriteState. This change doesn't adjust this as there are other places that use this data for logic purposes, something they probably shouldn't do, but that can come later.

### DIFF
--- a/pkg/auto/lights/config/root.go
+++ b/pkg/auto/lights/config/root.go
@@ -12,10 +12,10 @@ import (
 )
 
 const (
-	DefaultRefreshEvery      = time.Minute
+	DefaultRefreshEvery      = 15 * time.Minute
 	DefaultMaxRetries        = 2
 	DefaultBackOffMultiplier = time.Millisecond * 500
-	DefaultWriteCacheExpiry  = 45 * time.Second
+	DefaultWriteCacheExpiry  = 10 * time.Minute
 )
 
 // Root represent the configuration parameters available for the lighting automation.


pkg/auto/lights/action.go
```diff
@@ -44,6 +42,43 @@ func (a *clientActions) UpdateBrightness(ctx context.Context, now time.Time, req
...
+// cacheValid returns true when a cache value exists, is in date, and hasn't changed according to eq.
+func cacheValid[T any](oldWrite Value[T], now time.Time, cacheExpiry time.Duration, eq func(v T) bool) bool {
+	if !eq(oldWrite.V) {
+		return false
+	}
+	if cacheExpiry > 0 && now.Sub(oldWrite.At) > cacheExpiry {
+		return false
+	}
+	// here is where logic that would have different cache expiry for error writes
+	return true
```
dloveen-v, line 79: Are you missing a `oldWrite.hit()` here?

mattnathan: Yes, good spot. Though I'll add it inside the if - don't want this helper to result in side effects.

mattnathan:
All changes were committed (though one was squashed) - GH just didn't pick it up

PR #325, opened 2025-01-17
auto/lights: write to all lights on error

mattnathan:
Before we used to return early when writing to the lights when the first error was encountered. Now we keep going, collecting errors as we go.


dloveen-v (approved):
LGTM

PR #326, opened 2025-01-17
Add bacnet emergency lowerthan

deanredfern26:
Add an option on the bacnet alarm config to trigger an emergency if a value drops below a good ok value.
We already supported an equality comparison, this adds support for an equality comparison.
Also adds support for float values

### DIFF
--- a/pkg/driver/bacnet/merge/emergency.go
+++ b/pkg/driver/bacnet/merge/emergency.go
@@ -26,9 +26,10 @@ import (
 // - value read from that point is less than GreaterThanOkValue
 type AlarmConfig struct {
 	config.ValueSource
-	OkValue            int    `json:"okValue"`            // what we expect to read from the point when it is ok, any other value is an emergency
+	OkValue            int    `json:"okValue"`            // what we expect to read from the point when it is ok, any other value is an emergency. Not supported for float values.
 	GreaterThanOkValue int    `json:"greaterThanOkValue"` // we expect the point to be greater than this value, if it isn't we have an emergency
 	AlarmReason        string `json:"alarmReason"`        // the reason of the alarm
+	ValueType          string `json:"valueType"`          // the type of the value we are expecting to read, int, float etc.
 }
 
 type emergencyConfig struct {
@@ -102,6 +103,41 @@ func (t *emergencyImpl) PullEmergency(request *traits.PullEmergencyRequest, serv
 	return t.EmergencyApiServer.PullEmergency(request, server)
 }
 
+func (t *emergencyImpl) checkIntValueForEmergency(response any) (*traits.Emergency, error) {
+	data := &traits.Emergency{}
+
+	value, err := comm.IntValue(response)
+	if err != nil {
+		return nil, comm.ErrReadProperty{Prop: "alarmConfig", Cause: err}
+	}
+
+	if value != int64(t.config.AlarmConfig.OkValue) ||
+		value < int64(t.config.AlarmConfig.GreaterThanOkValue) {
+		data.Reason = t.config.AlarmConfig.AlarmReason
+		data.Level = traits.Emergency_EMERGENCY
+	} else {
+		data.Level = traits.Emergency_OK
+	}
+	return data, nil
+}
+
+func (t *emergencyImpl) checkFloatValueForEmergency(response any) (*traits.Emergency, error) {
+	data := &traits.Emergency{}
+
+	value, err := comm.Float64Value(response)
+	if err != nil {
+		return nil, comm.ErrReadProperty{Prop: "alarmConfig", Cause: err}
+	}
+
+	if value < float64(t.config.AlarmConfig.GreaterThanOkValue) {
+		data.Reason = t.config.AlarmConfig.AlarmReason
+		data.Level = traits.Emergency_EMERGENCY
+	} else {
+		data.Level = traits.Emergency_OK
+	}
+	return data, nil
+}
+
 // pollPeer fetches data from the peer device and saves the data locally.
 func (t *emergencyImpl) pollPeer(ctx context.Context) (*traits.Emergency, error) {
 	data := &traits.Emergency{}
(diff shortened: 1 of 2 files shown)


dloveen-v (approved):
Logic looks good to me.

Just a note that the types on `OkValue` and `GreaterThanOkValue` in the config can be `int64` to avoid the need to convert later on in one place.

pkg/driver/bacnet/merge/emergency.go
```diff
@@ -21,12 +21,15 @@ import (
 	"github.com/vanti-dev/sc-bos/pkg/task"
 )
 
-// AlarmConfig allows configuring a specific bacnet point to raise an Emergency if the
-// value read from that point is anything other than OkValue
+// AlarmConfig allows configuring a specific bacnet point to raise an Emergency if either:
+// - value read from that point is anything other than OkValue
+// - value read from that point is less than GreaterThanOkValue
```
mattnathan, line 26: Fyi lists in go docs should [have some blank space at the start](https://tip.golang.org/doc/comment#lists)

pkg/driver/bacnet/merge/emergency.go
```diff
@@ -21,12 +21,15 @@ import (
...
 type AlarmConfig struct {
 	config.ValueSource
-	OkValue     int    `json:"okValue"`     // what we expect to read from the point when it is ok, any other value is an emergency
-	AlarmReason string `json:"alarmReason"` // the reason of the alarm
+	OkValue            int    `json:"okValue"`            // what we expect to read from the point when it is ok, any other value is an emergency. Not supported for float values.
+	GreaterThanOkValue int    `json:"greaterThanOkValue"` // we expect the point to be greater than this value, if it isn't we have an emergency
+	AlarmReason        string `json:"alarmReason"`        // the reason of the alarm
+	ValueType          string `json:"valueType"`          // the type of the value we are expecting to read, int, float etc.
```
mattnathan, line 32: We should be able to do some type checking on the response without the user needing to configure this. BACnet is already typesafe in its transport and that filters through the gobacnet library. What I'm saying is that the response will be a `float32` or `int32` by the time we see it so you can check for this instead of configuring it.

pkg/driver/bacnet/merge/emergency.go
```diff
@@ -21,12 +21,15 @@ import (
...
-// value read from that point is anything other than OkValue
+// AlarmConfig allows configuring a specific bacnet point to raise an Emergency if either:
+// - value read from that point is anything other than OkValue
+// - value read from that point is less than GreaterThanOkValue
 type AlarmConfig struct {
 	config.ValueSource
-	OkValue     int    `json:"okValue"`     // what we expect to read from the point when it is ok, any other value is an emergency
-	AlarmReason string `json:"alarmReason"` // the reason of the alarm
+	OkValue            int    `json:"okValue"`            // what we expect to read from the point when it is ok, any other value is an emergency. Not supported for float values.
+	GreaterThanOkValue int    `json:"greaterThanOkValue"` // we expect the point to be greater than this value, if it isn't we have an emergency
```
mattnathan, line 30: I think we can shorten this name. Ok was used for the equality prop because `Value` wasn't descriptive enough. Maybe `OkAbove`, though checking your code it's really a `>=` condition, so `BadBelow` might be better.

Also can you add a line above OkValue to indicate that only one of these props should be specified. Come to think of it there's no way to know given this structure which condition should be checked for. For example:
`{"greaterThanOkValue": 0}` and a response value of `1` would result in an incorrect emergency value because `1 != OkValue`, which is unset and `0`

pkg/driver/bacnet/merge/emergency.go
```diff
@@ -100,6 +103,41 @@ func (t *emergencyImpl) PullEmergency(request *traits.PullEmergencyRequest, serv
...
+	value, err := comm.IntValue(response)
+	if err != nil {
+		return nil, comm.ErrReadProperty{Prop: "alarmConfig", Cause: err}
+	}
+
+	if value != int64(t.config.AlarmConfig.OkValue) ||
+		value < int64(t.config.AlarmConfig.GreaterThanOkValue) {
+		data.Reason = t.config.AlarmConfig.AlarmReason
+		data.Level = traits.Emergency_EMERGENCY
+	} else {
```
mattnathan, line 118: See comment on the config structure, this condition incorrect results in error due to default/zero values of either configured condition.

pkg/driver/bacnet/merge/emergency.go
mattnathan: I expect a test or two will go a long way to finding this kind of thing.

mattnathan (changes requested):

pkg/driver/bacnet/merge/emergency.go
deanredfern26: Oh dear oh dear *facepalm*. Yes this really should have some tests I will add them

pkg/driver/bacnet/merge/emergency.go
```diff
@@ -141,18 +204,18 @@ func (t *emergencyImpl) pollPeer(ctx context.Context) (*traits.Emergency, error)
...
+			case int32, int16, int8, uint32, uint16, uint8:
+				if e, err := t.checkIntValueForEmergency(response); err == nil {
+					data = e
+				}
+			case float32, float64:
+				if e, err := t.checkFloatValueForEmergency(response); err == nil {
+					data = e
+				}
+			default:
+				t.logger.Warn("alarmConfig unknown value type", zap.String("type", reflect.TypeOf(response).String()))
```
mattnathan, line 217: Looking at this, I'm wondering if there's a way to just do all the checking on a single type, say float64. Can you think of a reason we couldn't just do that? `comm.Float64Value()` already converts int types, we could change the config to be `*float64` too. I guess the only thing to think about is floats aren't great for `==` so some kind of _within range_ would need to be used.

If we did that we'd get better error handling, because `response` can have type `error` too and comm.Float64Value handles that correctly.

Thoughts?

mattnathan (changes requested):

pkg/driver/bacnet/merge/emergency.go
deanredfern26: The `==` issue is the reason I chose not to implement the equality check for floats. Having a within range would solve that as long as its sensible. I suppose implementing it like that would also mean we would only have to implement one check for all 3 cases; lower than, ~equal to and greater than. 

The person creating the config would just have to be sensible in setting the bounds for the ~equal case but thats the only issue

mattnathan (approved):
The change looks good, though it is a breaking change and will need some coordination to roll out to site.

Consider keeping the OkValue config field and converting it to bounds during config read, maybe deprecating the field too. Otherwise LGTM.

PR #329, opened 2025-01-21
ui/ops: replace a.hasOwnProperty(b) with Object.hasOwn(a, b)

mattnathan:
Calling hasOwnProperty on an instance was already discouraged in favour of Object.prototype.hasOwnProperty. As that's a mouth full modern browsers also have Object.hasOwn to replace this.


dloveen-v (approved):
Code LGTM. I am unsure of the compatibility chain and what browsers sc-bos is needed to support. I was expecting some changes to the transpiler or polyfill config to be paired with these changes. But I am guessing we don't need to support browsers not updated since 2021.

deanredfern26 (approved):

PR #331, opened 2025-01-22
ui/ops: update eslint from 8.57.0 to 9.18.0 - flat config update

mattnathan:
eslint 9 introduced a reimplemented config system that changed how config is loaded in a breaking (but mostly compatible) way. This change updates to v9 and makes the necessary changes which were fairly straight forward when following their guide.

One change that I made that might seem surprising is the removal of eslint-plugin-vuetify, this plugin was only intended for use during the v2->v3 migration and was causing compatibility issued with eslint 9.

eslint related plugins have also been updated and any adjustments to the code to fix warnings have been made.


deanredfern26 (approved):
ops ui runs fine and no lint errors after `yarn run lint:nofix`, looks good

PR #332, opened 2025-01-28
fix: add SecurityEvent support to allnodes serviceRegistry

deanredfern26:
@dloveen-v while reviewing the hbd Island PR I noticed the warning `cannot determine services to support for trait`, this PR fixes that as security event support was not added in allnodes


dloveen-v (approved):
Code LGTM. I don't know what the process is for resolving the CI failure. There is this commit which should resolve the CI failures with govulncheck: https://github.com/vanti-dev/sc-bos/commit/7f8985cde0f9edc9c2f264a1cefaf697ca1c9348

PR #334, opened 2025-01-28
router: fix panic when downloading devices from the gateway

mattnathan:
The cause of the panic was proto.Merge not behaving as advertised when merging proto.Messages that have equivalent but != protoreflect.Descriptor instances, like we have on the gateway as we use protodesc and dynamicpb packages to create dynamic descriptors and messages.

The fix is to fall back to the slow path in these cases and round trip the merge via proto.Mershal/Unmarshal. proto.Merge is advertised to be equivalent to this anyway.

Fixes SC-1002

### DIFF
--- a/internal/router/loopback.go
+++ b/internal/router/loopback.go
@@ -2,12 +2,14 @@ package router
 
 import (
 	"context"
+	"fmt"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/dynamicpb"
 )
 
 // Loopback implements grpc.ClientConnInterface using a MethodResolver to find the real connection for each request.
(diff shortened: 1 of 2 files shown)


internal/router/loopback.go
```diff
@@ -196,6 +198,33 @@ func (c copyRecver) RecvMsg(m any) error {
...
+	// fall back to marshalling and unmarshalling,
+	// we assume that the descriptors are the same if their names are the same
+	if want, got := dst.ProtoReflect().Descriptor().FullName(), src.ProtoReflect().Descriptor().FullName(); want != got {
+		panic(fmt.Sprintf("descriptor mismatch: %v != %v", got, want))
+	}
+	data, err := proto.Marshal(src)
+	if err != nil {
+		panic(fmt.Sprintf("src marshal: %v", err))
+	}
+	if err := proto.Unmarshal(data, dst); err != nil {
```
dloveen-v, line 227: When the generated Golang proto wrappers [initialize the same proto definition](https://github.com/smart-core-os/sc-api/blob/main/go/traits/air_quality_sensor_grpc.pb.go#L141), is there a risk the response proto.Message has an unknown field (if something goes wrong and a server is updated before we manage to update the client with the same proto file version).

Do we need Unmarshal options such as https://github.com/protocolbuffers/protobuf-go/blob/v1.36.4/proto/decode.go#L38?

mattnathan: I'm not 100% sure what you mean about unknown fields. In the fixed case one of the proto.Message is a static type (like *traits.Metadata) and the other is a dynamic type (*dynamicpb.Message) - there's no server/client interaction here.

However I do think it's worth including the Merge and Partial options with both the marshal and unmarshal operations here.

deanredfern26 (approved):
Tested at OCW, this has fixed the panic caused by the download CSV button

internal/router/loopback.go
dloveen-v: Unknown fields are present in one message but not the other. They happen when the proto definitions are updated by adding a field in one place but not the other. `Unmarshal` will throw an error if this happens unless the right unmarshal options are given.

dloveen-v: As an aside, I believe `AllowPartial` is made redundant by the removal of `required / optional` tags in proto3.

mattnathan: I'm still not sure I understand what you're saying. I definitely don't want to ignore unknown fields, that would break the transparent nature of this package and force it to be the most up to date of all cohort nodes, something we can't guarantee.

Can you give a concrete example of what you mean? I can't think of a case where unknown fields would cause Unmarshal to error, it would just put those unknown fields into the unknown fields map in the message.

In my head it looks like this:

```
// foo.proto v1
message Foo {
  string a = 1;
}

// foo.proto v2
message Foo {
  string a = 1;
  string b = 2;
}

// v1 src, v2 dst
m1 := &v1.Foo{a:"a"}
m2 := &v2.Foo{}
data, _ := proto.Marshal(m1) // data looks like [1:"a"]
proto.Unmarshal(data, m2)    // m2 lookg like {a:"a"}

// v2 src, v1 dst
m1 := &v2.Foo{a:"a", b:"b"}
m2 := &v1.Foo{}
data, _ := proto.Marshal(m1) // data looks like [1:"a", 2:"b"]
proto.Unmarshal(data, m2)    // m2 looks like {a:"a", unknown:{2:"b"}}
```

dloveen-v: You're right. The `proto` package ignores unknown fields from bytes when unmarshalling into `m2`.

However, the `protojson` package does require being explicit for `DiscardUnknown` during unmarshal.

I have attached a small verifiable example of this.
[test-proto.zip](https://github.com/user-attachments/files/18586207/test-proto.zip)

dloveen-v (approved):
LGTM

PR #335, opened 2025-01-28
traits: add mock driver support for security event trait

deanredfern26:
add mock driver support for security event trait, add mock security driver to UGS config and add client test tool SC-999

### DIFF
--- a/pkg/driver/mock/auto/securityevent.go
+++ b/pkg/driver/mock/auto/securityevent.go
@@ -1,0 +1,33 @@
+package auto
+
+import (
+	"context"
+	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/vanti-dev/sc-bos/pkg/gentrait/securityevent"
+	"github.com/vanti-dev/sc-bos/pkg/task/service"
+)
+
+func SecurityEventAuto(model *securityevent.Model) *service.Service[string] {
+	slc := service.New(service.MonoApply(func(ctx context.Context, _ string) error {
+		go func() {
+			timer := time.NewTimer(30 * time.Second)
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-timer.C:
+					_, _ = model.GenerateSecurityEvent(timestamppb.Now())
+					timer.Reset(30 * time.Second)
+				}
+			}
+		}()
+		return nil
+	}), service.WithParser(func(data []byte) (string, error) {
+		return string(data), nil
+	}))
+	_, _ = slc.Configure([]byte{})
+	return slc
+}
--- a/pkg/gentrait/securityevent/model.go
+++ b/pkg/gentrait/securityevent/model.go
@@ -1,0 +1,83 @@
+package securityevent
+
+import (
+	"context"
+	"strconv"
+	"time"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/smart-core-os/sc-golang/pkg/resource"
+	"github.com/vanti-dev/sc-bos/pkg/gen"
+)
+
+var subSystems = [2]string{"access control", "cctv"}
+
+type Model struct {
+	lastSecurityEvent *resource.Value // of *gen.SecurityEvent
+	allSecurityEvents []*gen.SecurityEvent
+	ticker            *time.Ticker
+	ctx               context.Context
+	genId             int
+}
+
+func NewModel(opts ...resource.Option) *Model {
+	defaultOpts := []resource.Option{resource.WithInitialValue(&gen.SecurityEvent{})}
+	opts = append(defaultOpts, opts...)
+
+	m := &Model{
+		lastSecurityEvent: resource.NewValue(opts...),
+	}
+
+	// let's add some events to start with so we can test the list method without waiting
+	startTime := time.Now().Add(-100 * time.Minute)
+	for i := 0; i < 100; i++ {
+		_, _ = m.GenerateSecurityEvent(timestamppb.New(startTime))
+		startTime = startTime.Add(time.Minute)
+	}
+
+	return m
+}
+
+// AddSecurityEvent manually add a security event to the model
+func (m *Model) AddSecurityEvent(se *gen.SecurityEvent, opts ...resource.WriteOption) (*gen.SecurityEvent, error) {
+	v, err := m.lastSecurityEvent.Set(se, opts...)
+	if err != nil {
+		return nil, err
+	}
+	m.allSecurityEvents = append(m.allSecurityEvents, se)
+	return v.(*gen.SecurityEvent), nil
+}
+
+// GenerateSecurityEvent generates a security event and adds it to the model
+func (m *Model) GenerateSecurityEvent(time *timestamppb.Timestamp) (*gen.SecurityEvent, error) {
+	se := &gen.SecurityEvent{
+		SecurityEventTime: time,
+		Description:       "Generated security event",
+		Id:                strconv.Itoa(m.genId),
+		Source: &gen.SecurityEvent_Source{
+			Subsystem: subSystems[m.genId%2],
+		},
+		EventType: gen.SecurityEvent_EventType(m.genId % 22),
+	}
+	m.genId++
+	return m.AddSecurityEvent(se)
+}
+
+func (m *Model) PullSecurityEvents(ctx context.Context, opts ...resource.ReadOption) <-chan *gen.PullSecurityEventsResponse_Change {
+	send := make(chan *gen.PullSecurityEventsResponse_Change)
+
+	recv := m.lastSecurityEvent.Pull(ctx, opts...)
+	go func() {
+		defer close(send)
+		for change := range recv {
+			value := change.Value.(*gen.SecurityEvent)
+			send <- &gen.PullSecurityEventsResponse_Change{
+				OldValue: nil,
+				NewValue: value, // the mock driver only generates new security events and does not delete them
+			}
+		}
+	}()
+
+	return send
+}
--- a/pkg/gentrait/securityevent/model_server.go
+++ b/pkg/gentrait/securityevent/model_server.go
@@ -1,0 +1,73 @@
+package securityevent
+
+import (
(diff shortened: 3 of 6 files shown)


pkg/gentrait/securityevent/model_server.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	count := req.PageSize
+	if count == 0 {
+		count = 50
+	} else if count > 1000 {
+		count = 1000
+	}
+
+	var resp gen.ListSecurityEventsResponse
```
mattnathan, line 50: This is a strange way to do this, normally we'd write `resp := &gen.ListSecurityEventsResponse{}` as the non-pointer type isn't a `proto.Message`, only the pointer version is.

pkg/gentrait/securityevent/model_server.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// reverse to retrieve the latest events first
+	for i := startIndex - 1; i >= 0; i-- {
+		resp.SecurityEvents = append(resp.SecurityEvents, m.model.allSecurityEvents[i])
+		if len(resp.SecurityEvents) >= int(count) {
+			resp.NextPageToken = strconv.Itoa(i - 1)
```
mattnathan, line 56: Normally I'd say using an index as a page token isn't safe, indexes tend to change and you can get duplicated or missing items between pages if the list grows/shrinks between requests. I _think_ it's safe in this case though because the index starts at the oldest event so shouldn't change. Might be worth a comment line somewhere to let future us know why it's safe here.

pkg/gentrait/securityevent/model_server.go
```diff
@@ -0,0 +1,73 @@
...
+			break
+		}
+	}
+	resp.TotalSize = int32(len(m.model.allSecurityEvents))
+	return &resp, nil
+}
+
+func (m *ModelServer) PullSecurityEvents(request *gen.PullSecurityEventsRequest, server gen.SecurityEventApi_PullSecurityEventsServer) error {
+	for change := range m.model.PullSecurityEvents(server.Context(), resource.WithReadMask(request.ReadMask), resource.WithUpdatesOnly(request.UpdatesOnly)) {
```
mattnathan, line 65: It looks like this implementation only supports updates only, or at least when UpdatesOnly is false they only get 1 event returned. I'm not sure what the correct thing to do here is as the event list can be large. Maybe when UpdatesOnly is false we could sent a default page size of records back down the stream first?

pkg/gentrait/securityevent/model.go
```diff
@@ -0,0 +1,83 @@
...
+
+	recv := m.lastSecurityEvent.Pull(ctx, opts...)
+	go func() {
+		defer close(send)
+		for change := range recv {
+			value := change.Value.(*gen.SecurityEvent)
+			send <- &gen.PullSecurityEventsResponse_Change{
+				OldValue: nil,
+				NewValue: value, // the mock driver only generates new security events and does not delete them
+			}
```
mattnathan, line 78: You're missing some fields for this change: ChangeTime and ChangeType

mattnathan (changes requested):

pkg/driver/mock/auto/securityevent.go
```diff
@@ -0,0 +1,33 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/vanti-dev/sc-bos/pkg/gentrait/securityevent"
+	"github.com/vanti-dev/sc-bos/pkg/task/service"
+)
+
+func SecurityEventAuto(model *securityevent.Model) *service.Service[string] {
+	slc := service.New(service.MonoApply(func(ctx context.Context, _ string) error {
+		go func() {
+			timer := time.NewTimer(30 * time.Second)
```
mattnathan, line 16: You could just use a ticker here, can't think it's doing anything taxing that'd need the 30s to be strictly between tasks instead of periodic.

pkg/gentrait/securityevent/model.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/vanti-dev/sc-bos/pkg/gen"
+)
+
+var subSystems = [2]string{"access control", "cctv"}
+
+type Model struct {
+	lastSecurityEvent *resource.Value // of *gen.SecurityEvent
+	allSecurityEvents []*gen.SecurityEvent
+	genId             int
+	mu                sync.Mutex
```
mattnathan, line 22: Generally with Mutexes (and other sync types) we put it before the things that it guards and comment what it guards:
```go
type Foo struct {
  other Prop

  mu sync.Mutex // guards the props below, or guards a and b, or guards non-exported fields, or something like that
  a string
  b int

  other Prop
}
```

pkg/gentrait/securityevent/model.go
```diff
@@ -0,0 +1,87 @@
...
+	se := &gen.SecurityEvent{
+		SecurityEventTime: time,
+		Description:       "Generated security event",
+		Id:                strconv.Itoa(m.genId),
+		Source: &gen.SecurityEvent_Source{
+			Subsystem: subSystems[m.genId%2],
+		},
+		EventType: gen.SecurityEvent_EventType(m.genId % 22),
+	}
+	m.genId++
```
mattnathan, line 66: This is not safe for concurrent use. Either you need to document that only Add and Pull are safe or make this safe too.

pkg/gentrait/securityevent/model_server.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (m *ModelServer) ListSecurityEvents(_ context.Context, req *gen.ListSecurityEventsRequest) (*gen.ListSecurityEventsResponse, error) {
+
+	// page token is just the index of where we left off (if any)
+	// this works with the current basic implementation because we only support a list of all events without filtering/sorting
+	// and the events are stored in ascending chronological order. If this either of these things change, this will need to be rethought
+	pageToken := req.GetPageToken()
+	m.model.mu.Lock()
+	defer m.model.mu.Unlock()
+	startIndex := len(m.model.allSecurityEvents)
```
mattnathan, line 39: This is _really_ bad practice, mutexes have an owner and that owner is usually the type that contains them. Only the owner should directly use or reference that mutex. You need to move the functionality that needs the lock into the Model type, I guess on a new method.

pkg/gentrait/securityevent/model_server.go
```diff
@@ -0,0 +1,101 @@
...
+	resp.TotalSize = int32(len(m.model.allSecurityEvents))
+	return resp, nil
+}
+
+// PullSecurityEvents returns a channel of security events
+// If updatesOnly is false, only the previous 50 events will be sent before any new events
+// For historical events use ListSecurityEvents
+func (m *ModelServer) PullSecurityEvents(request *gen.PullSecurityEventsRequest, server gen.SecurityEventApi_PullSecurityEventsServer) error {
+	if !request.UpdatesOnly {
+		m.model.mu.Lock()
```
mattnathan, line 74: See other comment, nothing should reference the mutex than the Model type and its methods.

pkg/gentrait/securityevent/model_server.go
```diff
@@ -0,0 +1,101 @@
...
+				m.model.mu.Unlock()
+				return err
+			}
+		}
+		m.model.mu.Unlock()
+	}
+	for change := range m.model.PullSecurityEvents(server.Context(), resource.WithReadMask(request.ReadMask), resource.WithUpdatesOnly(request.UpdatesOnly)) {
+		var msg gen.PullSecurityEventsResponse
+		msg.Changes = append(msg.Changes, change)
+		if err := server.Send(&msg); err != nil {
```
mattnathan, line 96: When a type is defined in terms of its pointer type, try to define variables in that form. proto.Message implementations are all pointer types (*OnOff not OnOff). Instead of

```go
var msg OnOff
...
doSomething(&msg)
```

We should be true to type and write

```go
msg := &OnOff{} // or msg := new(OnOff)
...
doSomething(msg)
```

mattnathan (changes requested):

pkg/gentrait/securityevent/model.go
```diff
@@ -0,0 +1,166 @@
...
+	return send
+}
+
+func (m *Model) unlock() {
+	m.mu.Unlock()
+}
+
+func (m *Model) lock() {
+	m.mu.Lock()
+}
```
mattnathan, line 166: I don't think these are required

pkg/gentrait/securityevent/model.go
```diff
@@ -0,0 +1,166 @@
...
+			Subsystem: subSystems[m.genId%2],
+		},
+		EventType: gen.SecurityEvent_EventType(m.genId % 22),
+	}
+	m.genId++
+	m.mu.Unlock()
+	return m.AddSecurityEvent(se)
+}
+
+func (m *Model) ListSecurityEvents(req *gen.ListSecurityEventsRequest) (*gen.ListSecurityEventsResponse, error) {
```
mattnathan, line 73: All other models we have try not to require the rpc request/response objects to interact with the data. Could you add a `ListSecurityEvents(start, end int) []*gen.SecurityEvent` function? Would that work?

pkg/gentrait/securityevent/model.go
```diff
@@ -0,0 +1,166 @@
...
+		if len(resp.SecurityEvents) >= int(count) {
+			resp.NextPageToken = strconv.Itoa(i - 1)
+			break
+		}
+	}
+	resp.TotalSize = int32(len(m.allSecurityEvents))
+	return resp, nil
+}
+
+func (m *Model) PullSecurityEventsWrapper(request *gen.PullSecurityEventsRequest, server gen.SecurityEventApi_PullSecurityEventsServer) error {
```
mattnathan, line 110: Other models that need to do this have function that return both current data and a stream like resource.Collection does.

If you want to keep this func here then it shouldn't be exported as its not really something people who just have a model can use.

mattnathan (changes requested):

mattnathan (approved):

PR #337, opened 2025-01-30
history: don't call Slice.Len when it's unlikely to have changed

mattnathan:
Calling history.Slice.Len for each page has large performance implications on the application, especially in cases where many pages are being read. In situations where Len is backed by a database call it's generally implemented via a `select count(*) where d>%T AND d<=%T` which can be very slow.

The problem is easy to avoid as the total size for a pagination operation is unlikely to change between page calls, so we solve this by calling Len for the first page and keeping that value in the page token for reuse by the next page calls.

During testing this changed history download from ~60KB/s to ~2MB/s, or to put it another way downloading ~1m records now takes seconds instead of hours.


pkg/system/history/history_system_page.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
```
dloveen-v, line 1: Should this file be called `system_history_page.proto`?

dloveen-v (approved):
The code LGTM. Although, I am conscious this risks moving the database optimisation step into the application logic. Which works for now and granted, we need to support SQL and non-SQL databases too which is even more difficult to manage. But longer term, I suggest looking at building indexes, frequent table clustering or frequently generating materialised views for (SQL databases) in the history table itself to optimise performance there.

From an API user perspective, it would be quite strange if I issue a `ListHistoryRecords()` call, and wait a few days before clicking the next page to find the `totalCount` field is unchanged in the new response, when the expectation is many new rows would've been created in that time.

I have also tested by changing the time period for a device, but I forgot to remove the stale `nextPageToken` in the request and found it returned the `totalSize` from my previous request, rather than computing one for the current time period. FYI for when this is integrated to the Ops UI.

mattnathan:
@dloveen-v The intended behaviour of page tokens is that they should page over a consistent view of the data. New items added to a list after you've already got a nextPageToken technically shouldn't be shown on subsequent pages. This is actually quite difficult to accomplish in general, but we do our best. Similarly adjusting the query but keeping the page token is also something that should be invalid but are inconsistent in checking for it.

With that in mind, the total count for a query should remain constant throughout the paging by design.

Regarding the db leakage, not calling Len is an optimisation for _any_ Store implementation, it just happens to be a significant optimisation for DB backed stores. We do have indexes on these tables, though I expect they can be improved. The issue here is that to get the totalCount you need to count the unlimited result set, which is way more expensive than the limited rows we paginate through. This will always be the case without considerable DB restructuring.

PR #338, opened 2025-01-31
devices: allow bulk download of historical trait data

mattnathan:
This PR extends the existing DevicesApi.GetDownloadURL to support historical ranges of records to download.

The PR consists of a few bits:
1. Changes to the proto to support the additional fields
2. Changes to the DevicesApi server and HTTP handler to implement these methods
3. Updates to the Ops ui to allow user interaction with this feature

**Server Changes**

The server now knows how to collect historical information from traits via their History aspect. I've implemented a number of grouping/sorting options for the download as we don't know which the client will want. Instead of allowing the choice on the client end I've statically liked the different options in download.go so they are easy to change and rebuild.

I expect future changes will include the ability for people to choose their grouping and more control over the columns the CSV returns. I believe there is space in the api and impl to accomplish this in other PRs.

**Ops UI Changes**

The Download button on the devices page now has an option menu you can click to configure the period of historical data you wish to download.

download-history-btn

I hope future PRs will extend this button to allow it to replace most if not all of our download buttons around the app, and add download buttons to more places where we need them. I've designed the UX for the download button to allow space for column selection and other options when they are implemented.


internal/manage/devices/download.go
```diff
@@ -119,33 +120,189 @@ func (s *Server) DownloadDevicesHTTPHandler(w http.ResponseWriter, r *http.Reque
 	if err := csvOut.Write(headers); err != nil {
 		return
 	}
+	out := newCSVWriter(csvOut, headerIndex)
```
dloveen-v, line 123: I noticed in the retro one of the files you downloaded was over 100MB big.

What's the downside of wrapping a gzip.Writer? It uses more compute but can reduce the amount of data sent over the wire.


```
	w.Header().Set("Content-Disposition", "attachment; filename=devices.csv.gz")
	w.Header().Set("Content-Type", "application/gzip")

	gz := gzip.NewWriter(w)
	csvOut := csv.NewWriter(gz)

        defer func() {
		csvOut.Flush()
		err = gz.Flush()
		if err != nil {
			writeErr(err)
		}
		err = gz.Close()
		if err != nil {
			writeErr(err)
		}
	}()
```

Note: using `Content-Encoding` header is not to be used alongside `Content-Type`. `Content-Encoding` is used to tell the browser to auto decompress the file. In this case, the `content-type` would be `text/csv`.

I suggest experimenting with the 3 variants with your setup as you have more data in the history. But checking the timing tab in Inspector tools for me reveals marginally lower download times when gzipping the file at the server.
Screenshot 2025-02-03 at 12 29 23
Screenshot 2025-02-03 at 12 29 30

deanredfern26:
I don't know if I am doing something wrong here but when I hit download for 24h data for just one device, nothing happends for over 30 seconds then I get a 600kb file with ~4000 records. This is using OCW as a test, I will try tomorrow to set if it is temporary, but I think having something let the user know the button click has actually been received might be worth it 'your download should begin shortly' or somthing?

mattnathan:
> I don't know if I am doing something wrong here but when I hit download for 24h data for just one device, nothing happends for over 30 seconds then I get a 600kb file with ~4000 records. This is using OCW as a test, I will try tomorrow to set if it is temporary, but I think having something let the user know the button click has actually been received might be worth it 'your download should begin shortly' or something?

Hmm, I will investigate and either fix the issue or add something to show the user that the download will begin soon. I believe the download doesn't begin until the first chunk of data is ready, which depends on the history api for the devices you've selected.

Is the update deployed anywhere specific on site? Particular floor or gateway?

deanredfern26:
> > I don't know if I am doing something wrong here but when I hit download for 24h data for just one device, nothing happends for over 30 seconds then I get a 600kb file with ~4000 records. This is using OCW as a test, I will try tomorrow to set if it is temporary, but I think having something let the user know the button click has actually been received might be worth it 'your download should begin shortly' or something?
> 
> Hmm, I will investigate and either fix the issue or add something to show the user that the download will begin soon. I believe the download doesn't begin until the first chunk of data is ready, which depends on the history api for the devices you've selected.
> 
> Is the update deployed anywhere specific on site? Particular floor or gateway?

It is deployed to eg-02 (172.16.100.11) https://scos-ocw.paradisebuild.uk. Just tested again using 1 meter & 24h worth of data. Is very slow and the resulting file is only 530kB.

I can't see anything when following the backend logs that suggests an issue

mattnathan:
FYI, Dean's issue is related to postgres query performance. I am investigating alternate indexing/query approaches for the history queries.

mattnathan:
@deanredfern26 I've pushed a fix for the performance issues you were seeing, let me know if its any better. On my local machine with the production dataset (~16GB, 150m rows) it performs at about 500kB/s, which is surprisingly slow but much better than the old 1min / 100 rows behaviour.

Testing against the UGS example, the performance has increased from ~5mb/s to over 10mb/s for the download.

Running the underlying queries used by the history paging logic, I saw performance characteristics like: old query=1m 15s, new query=8ms. If things still aren't working on site then I expect it'll be a code problem.

Things to note:
1. The performance fix applies to the hub (BC) only as this is the node that connects to the db
2. The indexes for the history table have now changed which will take ~10m to create the first time the BC is booted up.

If desired I can pull these performance changes out into their own PR so they can be tested and deployed independently.

internal/manage/devices/download.go
mattnathan: I was going by the imported by stats on the go package site.

mattnathan:
I decided to pull out the performance change into another PR #340 - while this PR is technically independent, it makes sense to land the performance fix first so we can actually deploy this to sites with large historical data sets.

ui/ops/src/components/download/DeviceDownloadBtn.vue
```diff
@@ -0,0 +1,203 @@
+<template>
+  <v-btn-group class="menu-btn">
+    <v-btn v-bind="{...dlBtnProps, ...dlBtnLinkProps}" :disabled="!!dlBtnDisabled" v-tooltip:bottom="dlBtnTooltip">
+      <template v-if="dlBtnHasPreIcon">
+        <v-icon size="24" :start="dlBtnHasText">mdi-file-download</v-icon>
+      </template>
+      <template v-if="dlBtnHasText">
```
deanredfern26, line 7: When a user clicks this button, I think the text should change to 'your download should begin shortly, please wait' or something to that effect (I know the button is small). Also, I think the button should become disabled for a short time after clicking, so the user does'nt hammer the button if the download doesn't instantly begin.

deanredfern26 (changes requested):
This is still very slow at OCW, the I can run a select between .. query on the building controller for 24 and that runs in milliseconds, the download from the button is still taking a while. Not sure if this branch needs to be rebased or is waiting for other PRs to be merged in too?

ui/ops/src/components/download/DeviceDownloadBtn.vue
mattnathan: That's a good idea, I'll sort it out.

hexaglow:
Out of interest, what were the [query execution plans](https://www.postgresql.org/docs/8.3/sql-explain.html) for the old vs new schema/queries?

mattnathan:
> Out of interest, what were the [query execution plans](https://www.postgresql.org/docs/8.3/sql-explain.html) for the old vs new schema/queries?

@hexaglow here you go - against a prod dataset of >100m rows. The plan is different when the row size is smaller, ~2m rows (I don't have the plan).

```sql
-- before
explain analyze
select id, create_time, payload
from history
where source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'
  and create_time >= '2025-01-01T00:00:00+00'
order by id asc
limit 51;

-- local: 1m 15s
-- Limit  (cost=0.57..4691.17 rows=51 width=22) (actual time=45268.172..45277.116 rows=51 loops=1)
--   ->  Index Scan using history_id_source_idx on history  (cost=0.57..9181530.93 rows=99829 width=22) (actual time=45268.165..45277.092 rows=51 loops=1)
--         Index Cond: (source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'::text)
--         Filter: (create_time >= '2025-01-01 00:00:00+00'::timestamp with time zone)
--         Rows Removed by Filter: 317914


explain analyze
select count(*)
from history
where source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'
  and create_time >= '2024-06-01T00:00:00+00';
-- local: 17s
-- Finalize Aggregate  (cost=1218558.73..1218558.74 rows=1 width=8)
--   ->  Gather  (cost=1218558.51..1218558.72 rows=2 width=8)
--         Workers Planned: 2
--         ->  Partial Aggregate  (cost=1217558.51..1217558.52 rows=1 width=8)
--               ->  Parallel Bitmap Heap Scan on history  (cost=498714.41..1217474.53 rows=33592 width=0)
--                     Recheck Cond: ((source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'::text) AND (create_time >= '2025-01-01 00:00:00+00'::timestamp with time zone))
--                     ->  BitmapAnd  (cost=498714.41..498714.41 rows=80620 width=0)
--                           ->  Bitmap Index Scan on history_source_idx  (cost=0.00..4164.05 rows=368464 width=0)
--                                 Index Cond: (source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'::text)
--                           ->  Bitmap Index Scan on history_create_time_idx  (cost=0.00..494509.80 rows=26873231 width=0)
--                                 Index Cond: (create_time >= '2025-01-01 00:00:00+00'::timestamp with time zone)

-- after
explain analyze
select id, create_time, payload
from history
where source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'
  and id >=
      (select min(id)
       from history
       where source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'
         and create_time =
             (select min(create_time)
              from history
              where
                  source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'
                and create_time >= '2024-06-01T00:00:00+00'))
order by id asc
limit 51;
-- Limit  (cost=10.07..1114.60 rows=51 width=22) (actual time=4.831..17.526 rows=51 loops=1)
--   InitPlan 3 (returns $3)
--     ->  Aggregate  (cost=9.50..9.51 rows=1 width=8) (actual time=3.104..3.106 rows=1 loops=1)
--           InitPlan 2 (returns $1)
--             ->  Result  (cost=0.77..0.78 rows=1 width=8) (actual time=2.630..2.631 rows=1 loops=1)
--                   InitPlan 1 (returns $0)
--                     ->  Limit  (cost=0.69..0.77 rows=1 width=8) (actual time=2.628..2.628 rows=1 loops=1)
--                           ->  Index Only Scan using history_source_create_time_idx on history history_1  (cost=0.69..34852.37 rows=442119 width=8) (actual time=2.627..2.627 rows=1 loops=1)
--                                 Index Cond: ((source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'::text) AND (create_time IS NOT NULL) AND (create_time >= '2024-06-01 00:00:00+00'::timestamp with time zone))
--                                 Heap Fetches: 0
--           ->  Index Scan using history_source_create_time_idx on history history_2  (cost=0.69..8.71 rows=1 width=8) (actual time=3.099..3.100 rows=1 loops=1)
--                 Index Cond: ((source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'::text) AND (create_time = $1))
--   ->  Index Scan using history_id_source_idx on history  (cost=0.57..3192035.36 rows=147388 width=22) (actual time=4.830..17.516 rows=51 loops=1)
--         Index Cond: ((id >= $3) AND (source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'::text))

explain analyze
select count(*)
from history
where source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'
  and create_time >= '2024-06-01T00:00:00+00';
-- Finalize Aggregate  (cost=32628.80..32628.81 rows=1 width=8) (actual time=290.841..291.391 rows=1 loops=1)
--   ->  Gather  (cost=32628.59..32628.80 rows=2 width=8) (actual time=288.748..291.385 rows=3 loops=1)
--         Workers Planned: 2
--         Workers Launched: 2
--         ->  Partial Aggregate  (cost=31628.59..31628.60 rows=1 width=8) (actual time=270.218..270.219 rows=1 loops=3)
--               ->  Parallel Index Only Scan using history_source_create_time_idx on history  (cost=0.69..31168.05 rows=184216 width=0) (actual time=1.759..263.014 rows=135085 loops=3)
--                     Index Cond: ((source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'::text) AND (create_time >= '2024-06-01 00:00:00+00'::timestamp with time zone))
--                     Heap Fetches: 20
```

mattnathan:
Oh ffs. @deanredfern26 I've found the performance issue. Sorry for the churn. Can you test again please?

deanredfern26:
> Oh ffs. @deanredfern26 I've found the performance issue. Sorry for the churn. Can you test again please?

Ah backend change, does this branch contain this commit `history: fix performance issues with reading history pages` if not can you rebase as I will need this included.

mattnathan:
Hmm, actually maybe not. Though that change is a problem, it doesn't look like it's causing performance issues on site.

I currently suspect that the performance problem is due to the 2.5 load average on the area controller. It's taking over 3s to get 100 records when asking the area controller, but milliseconds when asking the db or building controller directly.

I'm trying to find a way to validate this theory

ui/ops/src/components/download/DeviceDownloadBtn.vue
mattnathan: This has been done now

deanredfern26 (approved):
Looks good, at OCW the download starts just before the pop-up message dissappears, which is nice.
The UI looks good and is working fine on the floors that are not 1-3, which we know have a separate issue not related to this PR.

deanredfern26 (changes requested):
Just noticed 'You're download should begin shortly'... it should be 'your'

deanredfern26:
Few questions after testing at Arup:

- How does this behave when it encounters a device that it can't get trait data for (i.e. see pic)? Does it ever finish or does it just hang forever? 
- This does not fetch any historical data at all when deployed Arup, I'm not sure why. The building controller has the history automations set up for each zone, and I can fetch the data from pg admin, but the downloaded csv doesn't not have any records


image

deanredfern26 (approved):

mattnathan:
One day I'll get this right. The changes were merged. This time GH didn't pick it up because my local branch was auto rebased due to upstream changes in the main branch immediately before I pushed these changes to main.

PR #340, opened 2025-02-05
history: fix performance issues with reading history pages

mattnathan:
This change adjusts the queries and indexes used by the postgres history store to be ~10,000x more performant. In general the changes involved replacing AND clauses with inner selects in certain cases where indexes could not be used for either selection or sorting.

I assume the difference in performance is due to Postgres not being aware that the create_time col is also consistently ordered, just like the PK. The change to the query makes use of this fact to better optimise how the query is run using nested selects.

### DIFF
--- a/pkg/history/pgxstore/schema.sql
+++ b/pkg/history/pgxstore/schema.sql
@@ -6,5 +6,7 @@ CREATE TABLE IF NOT EXISTS history
     payload     BYTEA       NULL
 );
 
-CREATE INDEX IF NOT EXISTS history_source_idx ON history (source);
-CREATE INDEX IF NOT EXISTS history_create_time_idx ON history (create_time);
+DROP INDEX IF EXISTS history_create_time_idx; -- replaced by history_id_source_idx;
+DROP INDEX IF EXISTS history_source_idx; -- replaced by history_source_create_time_idx
+CREATE INDEX IF NOT EXISTS history_id_source_idx ON history (id, source);
+CREATE INDEX IF NOT EXISTS history_source_create_time_idx ON history (source, create_time);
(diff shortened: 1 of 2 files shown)


deanredfern26 (approved):
Tested with query at OCW:
```
SELECT id, source, create_time, payload
	FROM public.history
	WHERE source = 'uk-ocw/floors/01/devices/CE1-electric-meter/NorthDBA/T1HVACTotalEnergy[smartcore.bos.Meter]'
	AND create_time BETWEEN '2025-01-01 00:00' AND '2025-02-01 00:00';
```

Speed before change: 54s
Speed after change: 0.5ms 

Very nice!

dloveen-v:
I didn't get the chance to see this one in time.

I thought I'd mention that clustering achieves the same thing as the query (that is, automatically only process rows belonging to a certain column range). It is effort to cluster every table though and the cluster order has to be maintained actively (so it works best with analytical queries in database environments were reading outpaces writes).

I'm going to try to test clustering with the legacy query when I get a chance.

mattnathan:
@dloveen-v I assume you mean partitioning? Unless clustering means something different for postgres?

dloveen-v:
Clustering refers to the physical ordering of rows on disk. My idea is to cluster on source, and create_time. Then when filtering on source, it theoretically, wouldn't require a full table scan to find the set of rows belonging to that source. When the database finds a row belonging to that source, clustering can also reduce the IO time since rows belonging to that source are physically grouped together on disk.

Likewise, when filtering on create_time, a table clustered on create_time can skip the scan and find the window of rows belonging to the create_time range, using any one of the search algorithms which are optimised, only if elements reside in order. 

**This is all theory** of course. The idea is you always test reoccurring queries in different setups before committing to a particular set of indexes and clustering configuration. But the aim remains: reduce the amount of required database optimisations in the application logic.

PR #341, opened 2025-02-06
demo/vanti-ugs: script to trust crts generated by docker on macOS

dloveen-v:
This should help when using the UI through a browser. Currently throws a certificate invalid page which isn't exactly purposeful for demos either. And when using the sc-bos demo as a backend to test site UIs, request don't even get through. So I thought I'd add a script for tooling purposes to trust these generated certificates.

Disclaimer, I have only tested this on MacOS. Feel free to add one for Windows too.


deanredfern26 (changes requested):
This is fine but can you add a line or two in the demo README so future people know that this exists and what it does

deanredfern26 (approved):
Nice we need more docs!

PR #342, opened 2025-02-06
app: enable pprof profiling endpoints

mattnathan:
The endpoints can be disabled via sysconf.

Fixes SC-1018


pkg/app/controller.go
```diff
@@ -279,6 +280,16 @@ func Bootstrap(ctx context.Context, config sysconf.Config) (*Controller, error)
...
 	mux.Handle("/__/version", httpAuth(Version))
+	if !config.DisablePprof {
+		// pprof endpoints, see net/http/pprof init() for more details
+		pprofMux := http.NewServeMux()
+		pprofMux.HandleFunc("GET /debug/pprof/", pprof.Index)
+		pprofMux.HandleFunc("GET /debug/pprof/cmdline", pprof.Cmdline)
+		pprofMux.HandleFunc("GET /debug/pprof/profile", pprof.Profile)
+		pprofMux.HandleFunc("GET /debug/pprof/symbol", pprof.Symbol)
+		pprofMux.HandleFunc("GET /debug/pprof/trace", pprof.Trace)
+		mux.Handle("GET /__/debug/pprof/", httpAuth(http.StripPrefix("/__", pprofMux)))
```
hexaglow, line 291: Equivalently, could just register the routes at their real path, instead of using `StripPrefix`

hexaglow (approved):
LGTM

pkg/app/controller.go
mattnathan: I wanted to avoid adding paths to the root as there's always a chance they will conflict with the root static hosting. It's not inconceivable that the ops ui will add a /debug page at some point in the future.

PR #344, opened 2025-02-11
traits: add transport trait

deanredfern26:
Adds a trait which represents a device that transports a load from one location to another. Common implementors within buildings will be lifts and escalators but could just as easily represent conveyors, dumb waiters etc. Also, the idea is this should be generic & expressive enough to extend to any mobility device which the building might offer, such as smart mobility scooters


proto/transport.proto
```diff
@@ -0,0 +1,178 @@
...
+    bool passenger_alarm = 8;
+    // Optional. The speed of the transport.
+    float speed = 9;
+    // Optional. The destinations that are supported by the transport.
+    repeated int32 supported_destinations = 10;
+    // Optional. True if the transport is currently active. Note this is different to mode,
+    // as the transport can be in an active or idle state in different modes.
+    bool active = 11;
+    // Optional. The actor describes an entity that is using the transport.
+    Actor actor = 12;
```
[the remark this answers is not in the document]
deanredfern26: The Location and Payload makes sense. Thinking about Journey in regards to a lift, if people keep getting in and out would that just be represented as a continuous journey?

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+
+    // The current location of the transport.
+    Location current_location = 1;
+    // Optional. The next destination(s) the transport will stop at.
+    repeated Location next_destinations = 2;
+    // Optional. The direction the transport is moving.
+    Direction moving_direction = 3;
+    // Optional. The total amount of load being carried.
+    // The units this is measured in is implementation specific. See TransportSupport
+    float load = 4;
```
mattnathan, line 145: When creating optional fields whose zero value could have some meaning you need to think about how to distinguish these two states (zero and absent). If possible, model the data so that zero means absent, but when that isn't possible specify the field as `optional` in the proto.

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+    // Optional. The status of each the door(s) of the transport. For multiple doors, use TransportSupport to identify multiple doors.
+    repeated DoorStatus door_status = 5;
+    // Optional. The mode the transport is in.
+    Mode mode = 6;
+    // Optional. A list of active fault signals being reported.
+    repeated Fault faults = 7;
+    // Optional. True if the passenger alarm has been activated.
+    bool passenger_alarm = 8;
```
mattnathan, line 153: I think this bool might benefit from being an enum instead.

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+      OPEN = 2;
+      OPENING = 3;
+      CLOSING = 4;
+      // The door is safety locked.
+      SAFETY_LOCKED = 5;
+      // The door remains in a position that is not fully open or closed.
+      LIMITED_OPENED = 6;
+    }
+
+    enum Mode {
```
mattnathan, line 54: Is there any way we can avoid `Mode` as a name here, we already have the `mode` trait and `Electric.Mode` which gets confusing enough as it is. `OperationMode` maybe?

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+      // A comb plate safety switch is activated.
+      COMB_PLATE_FAULT = 9;
+      GENERAL_FAULT = 10;
+      DOOR_FAULT = 11;
+      LEVELLING_FAULT = 12;
+      SAFETY_CIRCUIT_BREAK_FAULT = 13;
+      FAIL_TO_START = 14;
+      ALARM_BUTTON = 15;
+      DOOR_NOT_CLOSING = 16;
+    }
```
mattnathan, line 110: Seeing these faults and knowing we have the Status trait is making me wonder how the two would fit together. I have nothing to change here, but just thinking about it.

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+    Direction moving_direction = 3;
+    // Optional. The total amount of load being carried.
+    // The units this is measured in is implementation specific. See TransportSupport
+    float load = 4;
+    // Optional. The status of each the door(s) of the transport. For multiple doors, use TransportSupport to identify multiple doors.
+    repeated DoorStatus door_status = 5;
+    // Optional. The mode the transport is in.
+    Mode mode = 6;
+    // Optional. A list of active fault signals being reported.
+    repeated Fault faults = 7;
```
mattnathan, line 151: For faults it's useful to know when the fault occurred.

proto/transport.proto
mattnathan, line 153: How is `passenger_alarm` different from `Fault.ALARM_BUTTON`? Or is the latter a fault with the button? This isn't clear.

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+    // Optional. True if the passenger alarm has been activated.
+    bool passenger_alarm = 8;
+    // Optional. The speed of the transport.
+    float speed = 9;
```
mattnathan, line 155: Again here, zero speed is a valid value and currently indistinguishable from an absent speed value.

mattnathan, line 155: I'm not sure how much this will be an issue but speed could have a high change rate during acceleration of a lift. Similar to brightness levels on a light that is ramping. Not sure how to fix that here, but something to think about and maybe add some docs around?

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+    // Optional. The destinations that are supported by the transport.
+    repeated Location supported_destinations = 10;
+    // Optional. True if the transport is currently active. Note this is different to mode,
+    // as the transport can be in an active or idle state in different modes.
+    bool active = 11;
```
mattnathan, line 160: Especially true when the zero state is not the normal state for a field, there's no way to know if the transport is inactive or doesn't support the active property. As before this might benefit from being an enum instead?

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+    // Optional. The mode the transport is in.
+    Mode mode = 6;
+    // Optional. A list of active fault signals being reported.
+    repeated Fault faults = 7;
+    // Optional. True if the passenger alarm has been activated.
+    bool passenger_alarm = 8;
+    // Optional. The speed of the transport.
+    float speed = 9;
+    // Optional. The destinations that are supported by the transport.
+    repeated Location supported_destinations = 10;
```
mattnathan, line 157: Can you write something about the order of destinations having meaning? I don't think we can strictly specify it as bottom->top (as that makes no sense for a conveyor), but saying they have a logical ordering based on the device or something.

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+
+    // Journey describes the journey of the transport or payload.
+    // It will have a source and one or more destinations.
+    message Journey {
+      // The starting point of the journey.
+      Location start = 1;
+      // The destinations of the journey.
+      repeated Location destinations = 2;
+      // Optional. The reason for the journey.
+      string reason = 3;
```
mattnathan, line 134: Do you think a journey could have a start_time and eta?

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+    float speed = 9;
+    // Optional. The destinations that are supported by the transport.
+    repeated Location supported_destinations = 10;
+    // Optional. True if the transport is currently active. Note this is different to mode,
+    // as the transport can be in an active or idle state in different modes.
+    bool active = 11;
+
+    message Payload {
+        // The id of the payload.
+        string id = 1;
```
mattnathan, line 164: Please explain which context the id is useful.

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+        // The description of the payload.
+        string description = 2;
+        // The journey the payload is intending to take.
+        Journey intended_journey = 3;
+        // The actual journey the payload has taken.
+        Journey actual_journey = 4;
```
mattnathan, line 170: As there's no way to get completed payloads, I don't think this will ever include the actual journey, only the actual journey so far.

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+  bool updates_only = 3;
+}
+
+message PullTransportResponse {
+  repeated Change changes = 1;
+
+  message Change {
+    string name = 1;
+    google.protobuf.Timestamp change_time = 2;
+    Transport transport_state = 3;
```
mattnathan, line 222: `transport_state` -> `transport` after name change

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+        string description = 2;
+        // The journey the payload is intending to take.
+        Journey intended_journey = 3;
+        // The actual journey the payload has taken.
+        Journey actual_journey = 4;
+    }
+
+    // Optional. Payloads describes the entities that are currently being transported.
+    repeated Payload payloads = 12;
+}
```
mattnathan, line 175: I can't see any fault or mode that indicates that the transport is overcapacity. Is this not something that our example devices have?

mattnathan (changes requested):

proto/transport.proto
deanredfern26: The latter is a fault with the button, have commented Fault.ALARM_BUTTON to make this more clear

proto/transport.proto
deanredfern26: Strangely enough no, not in the KONE spec or the bacnet that I could see, have added it anyway

proto/transport.proto
```diff
@@ -0,0 +1,224 @@
...
+      // The destinations of the journey.
+      repeated Location destinations = 2;
+      // Optional. The reason for the journey.
+      string reason = 3;
+    }
+
+    // The current location of the transport.
+    Location current_location = 1;
+    // Optional. The next destination(s) the transport will stop at.
+    repeated Location next_destinations = 2;
```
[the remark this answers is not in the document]
deanredfern26: Not sure I quite understand what you mean here, how is this different from any other trait, when after a restart we are waiting for an initial value? You wouldn't be able to render the lift cab until you received current_location, but surely that is true for any device trait, particularly when using pub/sub?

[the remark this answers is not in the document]
mattnathan: I'm saying that the initial value for the Transport isn't enough to render an accurate (enough) representation of what's going on.

I'm assuming that locations are discreet: "floor 2 bay 1" and not continuous: "30% between floor 2 and floor 3"

An example might help: Say a lift has a single cab and that cab is currently travelling between GF and F10 when the app loads and calls GetTransport. The response will look something like this:

```js
{
  "current_location": ?,        // I'm not actually sure what this will be from the docs
  "next_destinations": ["F10"], // I assume this won't include GF
  "moving_direction": UP,
  "supported_destinations": ["GF", "F01", ..., "F10", ...],
  "load", "speed", etc
}
```
Given this response the app has no way to know where to render the cab, assuming it's trying to render a ui like this:

image

All the app knows is that the cab is in some location DOWN (opposite of UP) from F10.

I don't think we should have this trait model exact location and progress between destinations (while in transit), that would just cause PullTransport to be very chatty as the location changes over a journey. But, there are ways we can encode the start and end of a journey leg in a way that an app can use the information to _guess_ where the cab is with some accuracy. A start and end location and a departure time and eta would make this possible. I could see eta being optional and simulated in the app based on leg distance and so on.

Does that help?

proto/location.proto
mattnathan, line 1: I appreciate the effort to introduce common types, but I don't think this is the right time for it.

1. There's a wider effort we need to go through to model a building/site/space and how it relates to things
2. Using id here when everything else in sc uses `name` is confusing.
3. It looks like the location is reinventing `Metadata` with the title, description, and floor properties

For now, let's keep Location local to Transport and the MVP version of it we need to satisfy our use cases.

mattnathan: More importantly, stating that Location.id is unique across the site while also having individual devices responsible for surfacing those locations to callers is just asking for trouble. I could imagine each device/driver assuming that the locations it returns are unique to itself: "bay 2" instead of "LG carpark, EV bank 2, Bay 2"

proto/transport.proto
```diff
@@ -0,0 +1,265 @@
...
+    // i.e. For a linear transport, the first and last items in this list should be the termini.
+    repeated Location supported_destinations = 10;
+    // Optional. The active state of the transport. Note this is different to mode,
+    // as the transport can be in an active or idle state in different modes.
+    Active active = 11;
+
+    message Payload {
+        // The id of the payload. The id must be unique within the transport for the duration of the journey.
+        // i.e. there should not be two payloads with the same id in the same transport at the same time.
+        string id = 1;
```
mattnathan, line 213: Do you think it's worth adding something about consistency of id? I feel it's obvious that a payload with id "foo" will still have id "foo" even if the payloads array is changed, but I'm less clear if "foo" can exist again once it's been removed from Transport.payloads.

1. If "foo" can reappear, how to we cater for infrequent polls to GetTransport potentially missing the fact "foo" disappeared for a bit and is back.
2. If "foo" can't come back, how to deal with id exhaustion

I expect the solution is to say ids will remain consistent for a short time or something like that, but not sure of the exact wording.

proto/transport.proto
```diff
@@ -0,0 +1,265 @@
...
+
+    enum Alarm {
+      ALARM_UNSPECIFIED = 0;
+      UNACTIVATED = 1;
+      ACTIVATED = 2;
+    }
+
+    enum Active {
+      ACTIVE_UNSPECIFIED = 0;
+      INACTIVE = 1;
```
mattnathan, line 93: You use the term `idle` in the docs where Active is used. You should probably stay consistent.

proto/transport.proto
```diff
@@ -0,0 +1,265 @@
...
+    // A zero value means there is no load, an omitted value means the load is unknown.
+    optional float load = 4;
+    // Optional. The status of each the door(s) of the transport.
+    repeated Door doors = 5;
+    // Optional. The operating mode the transport is in.
+    OperatingMode operating_mode = 6;
+    // Optional. A list of active fault signals being reported.
+    repeated Fault faults = 7;
+    // Optional. The status of the passenger alarm button.
+    Alarm passenger_alarm = 8;
```
mattnathan, line 197: As with Fault, it'd be nice to know how long the alarm has been going for.

mattnathan (changes requested):

proto/transport.proto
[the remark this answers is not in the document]
deanredfern26: So, at least in the case of lifts, the current_location will be the floor the lift is currently on, which gets updated as the lift is moving. Is that not what you need to draw this? Also, bear in mind that there is no speed reported by the lift, at least not with KONE, how would you guess the current location without the speed if you only know the start and end (with the end being able to change if someone presses a button)?

proto/transport.proto
deanredfern26: I have added an additional map for external ids that "foo" might have. These can persist after a Journey is complete, but yes, from the Transports point of view,  once "foo" has completed its journey it won't reappear as "foo" .

proto/transport.proto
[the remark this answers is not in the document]
mattnathan: I think _currently on_ is a confusing term when the lift is moving. Based on what you said I'm going to assume it maps to whatever the cab says on the display inside (or in the lobby displays). More like `closest_location` or `display_location` or something like that might be better.

I was writing out an example that used real data and pixels and things but it got complicated and I think the important part is that the rendered cab should arrive at the destination floor around the same time as the real cab arrives at the real floor. For this the app needs to have some kind of ETA, either calculated based on data or returned in the Transport.

A possible calculation approach using the data as described above in Transport would be something like this:

```js
const constantSpeed = 1 / 5; // floors/s: 1 floor distance every 5 seconds
const startDeceleratingFloors = 0.75; // the cab starts decelerating this many floors before the destination floor
const decelerationTime = 3; // seconds. How long does deceleration last
const floorsToTravel = supported_destinations[current_floor : next_destinations[0]].length - 1; // -1, already at current_floor, see Travel Distance below
const timeAtConstantSpeed = (floorsToTravel - startDeceleratingFloors) / constantSpeed;
const estimatedTravelTime = timeAtConstantSpeed + decelerationTime;

// The app then sets up an animation that moves the cab from 
// current_floor to next_destination over estimatedTravelTime
```

Lots of assumptions there and other things that could cause the app to get out of sync with reality. Not a huge issue while moving, but annoying to explain if you've stood on F10 and the app says the cab should be there and it isn't, or even if the cab says its arrived and the doors are open before the app shows its there.

**About travel distance**

When writing this I realised that the travel distance is actually not accurate as a lift might support floors [F1, F2, F10], or F4 might be double height, or something.

proto/transport.proto
```diff
@@ -0,0 +1,288 @@
...
+
+    enum Active {
+      ACTIVE_UNSPECIFIED = 0;
+      INACTIVE = 1;
+      ACTIVE = 2;
+    }
+
+    message Fault {
+      enum FaultType {
+        FAULT_UNSPECIFIED = 0;
```
mattnathan, line 104: I think this should be `FAULT_TYPE_UNSPECIFIED` to match the enum name

proto/transport.proto
```diff
@@ -0,0 +1,288 @@
...
+      FREIGHT = 14;
+      FAILURE = 15;
+      REDUCED_SPEED = 16;
+      STORM_OPERATION = 17;
+      HIGH_WIND_OPERATION = 18;
+    }
+
+    message Alarm {
+      enum AlarmState {
+        ALARM_UNSPECIFIED = 0;
```
mattnathan, line 86: `ALARM_STATE_UNSPECIFIED` to match the enum name

proto/transport.proto
```diff
@@ -0,0 +1,288 @@
...
+        OPENING = 3;
+        CLOSING = 4;
+        // The door is safety locked.
+        SAFETY_LOCKED = 5;
+        // The door remains in a position that is not fully open or closed.
+        LIMITED_OPENED = 6;
+      }
+
+      // The id of the door, must be unique within the transport. This is used as the id of the door in door_status.
+      int32 id = 1;
```
mattnathan, line 175: I feel like this is left over from a previous iteration of the design. I see no `door_status` field.

I wonder if we could say that transports always have a fixed number of doors, I can't think of a situation where the number of doors would change. If the door number is fixed then we can rely on the index in `Transport.doors` as the id and get rid of this field.

proto/transport.proto
```diff
@@ -0,0 +1,288 @@
...
+      // Description of the door; front, rear, etc.
+      string description = 2;
```
mattnathan, line 177: Consider changing this to `display_name` or `title`. Description is more for things like `"Door used to access the upper floor"` whereas Title would be `"Primary"` or something like that.

mattnathan (changes requested):

proto/transport.proto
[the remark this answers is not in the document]
deanredfern26: I have added an optional eta for the next destination. Also have changed the name _current_location_ to _actual_position_. _actual_position_ is the name in the KONE spec and the BACnet spec uses _car_position_ so thought it best to stay similiar to these

mattnathan (approved):

PR #347, opened 2025-02-17
auto/wordpress: support multiple authorization headers in auth config for CDNs

dloveen-v:
Since the staging and production environments can require authentication with a CDN, as well as the actual WordPress plugin, the config for the WordPress automation needs to support multiple Authorization headers; joined using ` , `.

The config should declare these header strings in the order they will be sent and read by a REST server.

Example `Basic` auth secret data file:
```
username:m7Rand0mP@ssw0rd
```
The string in this file will then be Base64 Encoded as per the Basic Authorization header specification.

### DIFF
--- a/cmd/tools/test-wordpress/main.go
+++ b/cmd/tools/test-wordpress/main.go
@@ -92,12 +92,18 @@ const (
   "name":     "wordpress-test",
   "type":     "wordpress",
   "disabled": false,
-  "baseUrl": "https://vanti-plugin-test-com.stackstaging.com/wp-json/recording/v1/create",
+  "baseUrl": "https://app-my-paradise-co-uk.corestaging.co.uk/wp-json/recording/v1/create",
   "site":     "abc-test1",
-  "authentication": {
-	"type":       "Bearer",
-	"secretFile": "./.data/secrets/wordpress"
-  },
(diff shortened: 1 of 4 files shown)


dloveen-v:
Here is a pretty good discussion on the same issue https://github.com/nodejs/node/issues/45699

dloveen-v:
I've marked this as Draft as I have written to the Marketing team asking them to change their Auth header to something other than Authorization.

dloveen-v:
@mattnathan 

I have found articles which reject the header format in this PR. I have also found posts which reference the `Field Order` section and interpret that part of the RFC to mean multiple `Authorization` token strings are permitted.

https://stackoverflow.com/a/74312907/7316335

That being said, they [marketing team] have replied saying that they won't change anything, other than to add certain IP addresses to their allow list in their backend. Allow listing will apparently avoid the need for the primary `Basic` Authorization header. More importantly, only their staging environment uses the `Basic` authorization header.

So, to avoid technicalities of interpreting the RFC for HTTP headers, I am closing this PR, and will send them the IP address OCW uses to issue external connections from the building-controller. Hopefully, this doesn't change during the time they have us using their staging server. Thank you.

PR #348, opened 2025-02-18
ui/space: add air quality widget

mattnathan:
Also support controlling which widgets are shows via the ui config.

localhost_5174_home(QBic Luminen 5) (3)

I expect a future commit will make the ui-config widgets more of a suggestion and then rely on the configured zone to know exactly what this space supports. But as the feature has been deprioritised that can wait.

### DIFF
--- a/example/config/vanti-ugs/space-ui.json
+++ b/example/config/vanti-ugs/space-ui.json
@@ -5,7 +5,7 @@
   "config": {
     "home": "/home",
     "auth": {
-      "disabled": false,
+      "disabled": true,
       "keycloak": {
         "realm": "Smart_Core",
         "url": "http://localhost:8888/",
@@ -21,6 +21,15 @@
           "altText": "Smart Core logo - representing nodes and connections",
           "src": ""
         }
+      }
+    },
+    "pages": {
+      "home": {
+        "widgets": [
+          {"name": "air-quality"},
+          {"name": "lighting"},
+          {"name": "temperature"}
+        ]
       }
     }
   }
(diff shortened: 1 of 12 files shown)


ui/space/src/routes/components/airQuality.js
```diff
@@ -0,0 +1,235 @@
...
+        break;
+      }
+      result[k] = {value: v, status};
+    }
+    return result;
+  });
+
+  const score = computed(() => {
+    const presentScore = presentMetrics.value['score'];
+    const statusToLabel = (status) => {
```
dloveen-v, line 212: This might be cleaner as an enum. I was surprised to find JS doesn't have a native definition of enum, but searching online demonstrates the use of `Symbol` (introduced in ES6) with `Object.Freeze()` may be a slightly error-proneless approach. e.g. https://stackoverflow.com/a/44447975/7316335

mattnathan: done

example/config/vanti-ugs/space-ui.json
deanredfern26, line 1: For my convenience can you add a .env.vanti-ugs for the space UI with the vite config pointing to `vanti-ugs/space-ui.json` rather than `vanti-ugs/ui-config.json`

deanredfern26 (changes requested):
I ran it against the UGS demo and it looks good, nothing much to say. One thing that isn't part of this PR, in the AirTemperatureCard.vue, can you make the step, min and max configurable? We have been asked to change these per site so this would be handy

example/config/vanti-ugs/space-ui.json
mattnathan: This is now in main

mattnathan:
> I ran it against the UGS demo and it looks good, nothing much to say. One thing that isn't part of this PR, in the AirTemperatureCard.vue, can you make the step, min and max configurable? We have been asked to change these per site so this would be handy

@deanredfern26 
[SC-1028](https://vanti.youtrack.cloud/issue/SC-1028) Configurable min/max/step for temp card - created

deanredfern26 (approved):

PR #350, opened 2025-02-19
scripts: support new router/wrapper plugin output

deanredfern26:
Support both versions of router/wrapper -> https://github.com/smart-core-os/sc-golang/pull/18

### DIFF
--- a/scripts/gen-proto.sh
+++ b/scripts/gen-proto.sh
@@ -2,12 +2,26 @@
 
 pushd ../pkg/gen
 go generate
-# Remove imports of the gen package: `gen "github.com/vanti-dev/sc-bos/pkg/gen"`.
-sed -i.bak '/gen "github.com\/vanti-dev\/sc-bos\/pkg\/gen"/d' ../trait/gen/*.pb.go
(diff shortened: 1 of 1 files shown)


mattnathan (changes requested):
This doesn't really support both protoc plugin versions. If I happen to have the old versions then updates to proto files in this repo will end up in the `gen` package, using these will require importing the `gen` package. If instead I happened to have the new plugins installed then things will end up in the `genpb` package and using them will require importing the `genpb` package. This effectively means we can't us the new plugins until the entire repo has been migrated.

I think a better solution would be to have both plugin versions result in files being output to the same package (`pkg/gen`). Should be easily doable as each plugin version outputs to a different place:

```bash
if [ -d ../trait/gen ]; then
  # old version fixes to files
fi
if [ -d ../trait/genpb ]; then
  # new version fixes to files
fi
```

mattnathan (approved):
Tested and it all looks good to me. The mentions of `gen` in the new block surprised me, but I guess they come from the `go_package` option in the proto files :shrug:

PR #352, opened 2025-02-21
scripts: add gen proto script for bacnet rpc

deanredfern26:
Add a gen-proto script for the bacnet rpc and support both versions of the wrapper & router plugins


mattnathan (approved):
Good catch

PR #353, opened 2025-02-21
auto/wordpress: make use of confidence measure in occupancy response

dloveen-v:
Make use of Occupancy trait's confidence value so we don't send bad counts to WordPress.


deanredfern26 (commented):
Is there any reason the acceptable confidence value is 0.2? or is this arbitary?

dloveen-v:
> Is there any reason the acceptable confidence value is 0.2? or is this arbitary?

It's arbitrary. Technically, it should never be set to 0 as this means unknown according to the proto buf docs. So if it can't be 0 and has to be low, then the automation has to choose a random number that isn't 0, or 1 that "feels low enough".

deanredfern26:
Ok that's fine but probably worth a comment in the code just explaining the magic number

deanredfern26 (approved):

PR #354, opened 2025-02-21
ui/ops: add meter/ConsumptionCard widget

mattnathan:
This widget shows the meter consumption between two dates, where one of the dates can be `now`.

A few examples:
- Show usage for yesterday
- Show usage for this week so far
- Show total usage for last year

A few examples on a test dashboard I was using

image

Using this mark up

```html
    <div class="d-flex ga-2 mt-4">
      <consumption-card :name="totalDemandName" period="day" title="Consumption" width="14em" variant="tonal"/>
      <consumption-card :name="totalGeneratedName" period="day" title="Generation" width="14em" variant="text" color="success"/>
      <consumption-card :name="totalDemandName" period="month" width="14em" variant="flat" color="error"/>
      <consumption-card :name="totalGeneratedName" period="year" title="Solar Panels" width="14em" variant="outlined"/>
    </div>
    <div class="d-flex ga-2 mt-4">
      <consumption-card :name="totalDemandName" period="day" offset="1" title="Consumption" width="14em" variant="tonal"/>
      <consumption-card :name="totalDemandName" period="day" offset="2" title="Consumption" width="14em" variant="tonal"/>
      <consumption-card :name="totalDemandName" period="week" offset="1" title="Consumption" width="14em" variant="tonal"/>
      <consumption-card :name="totalDemandName" period="month" offset="1" title="Consumption" width="14em" variant="tonal"/>
    </div>
```

### DIFF
--- a/ui/ops/src/traits/meter/ConsumptionCard.vue
+++ b/ui/ops/src/traits/meter/ConsumptionCard.vue
@@ -1,0 +1,196 @@
+<template>
+  <v-card>
+    <v-tooltip activator="parent" location="bottom">
+      <!-- eslint-disable-next-line vue/no-v-html -->
+      <span v-html="tooltipStr"/>
+    </v-tooltip>
+    <v-card-title v-if="titleStr">{{ titleStr }}</v-card-title>
+    <template v-if="isMeterReset">
+      <v-card-text class="text-h6 font-weight-light">The meter was reset</v-card-text>
+    </template>
+    <template v-else>
+      <v-card-text>
+        <span class="value text-h1 font-weight-light">{{ usageStr }}</span>
+        <span v-if="unit" class="unit text-subtitle-1 font-weight-light">&nbsp;{{ unit }}</span>
+      </v-card-text>
+    </template>
+    <v-card-text class="text-body-1 pt-0">{{ timePeriodStr }}</v-card-text>
+  </v-card>
+</template>
+
+<script setup>
+import {useStartOf} from '@/composables/time.js';
+import {usePullMetadata} from '@/traits/metadata/metadata.js';
+import {
+  useDescribeMeterReading,
+  useMeterReading,
+  useMeterReadingAt,
+  usePullMeterReading
+} from '@/traits/meter/meter.js';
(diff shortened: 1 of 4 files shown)


ui/ops/src/traits/meter/ConsumptionCard.vue
```diff
@@ -0,0 +1,196 @@
...
+
+const readingDiff = computed(() => {
+  const start = readingAtStart.value;
+  const end = readingAtEnd.value;
+  if (isNullOrUndef(start) || isNullOrUndef(end)) {
+    return null;
+  }
+  return {usage: end.usage - start.usage};
+});
+const isMeterReset = computed(() => readingDiff.value?.usage < 0);
```
deanredfern26, line 126: I guess the that generated energy will always have a different dedicated meter, otherwise this will show reset if we have generated more than consumed?

Looking at that I did wonder this before, wow are we modelling energy consumption vs generation at the moment? Physical meters in the real world will have separate readings for this, `TotalActiveImportEnergy` & `TotalActiveExportEnergy` etc. but I don't think we have anything like this yet?

mattnathan: At sites I've configured we have two meters, one on the incomer and one on the PV array. We combine them together in the ui/dashboards.

I agree that we don't have a good way to model meter resets, though the trait does have this. I guess it's worth checking whether the meter readings from generation sources are positive or negative though.

Have created [SC-1029](https://vanti.youtrack.cloud/issue/SC-1029) Model direction for meters to cover this case at the trait level.

deanredfern26: For the meters I have worked with over the years, the readings are always positive and come under 'import' and 'export' depending on the direction of energy flow. MID compliant meters won't be able to have these registers reset. Also, these meters can have loads of registers: import/export active, import/export reactive, power factor, frequency .... Then you can have 3 phase meters which have all these readings for each phase.

This is well outside the scope of this PR though, I just thought I would bring it up as our meter modelling is quite basic currently

PR #355, opened 2025-02-24
ui/ops: add page to view waste records

deanredfern26:
Adds a page under the ops section to view waste records. Functionality is MVP, just has to be able to view records at this point.

Not sure about the icon as it is not really just about recycling, could use mdi-basket-fill or possibly a trash can, but I thought that was too much like delete.

The vanti ugs example is already configured to generate mock waste records too

### DIFF
--- a/example/config/vanti-ugs/ui-config.json
+++ b/example/config/vanti-ugs/ui-config.json
@@ -22,8 +22,6 @@
         "clientId": "scos-opsui"
       }
     },
-    "wasteRecordsSource" : "van/uk/brum/ugs/devices/waste",
-    "wasteRecordUnit" : "kg",
     "ops": {
       "pages": [
         {
@@ -598,7 +596,11 @@
             }
           ]
         }
-      ]
+      ],
+      "waste" : {
+        "source" : "van/uk/brum/ugs/devices/waste",
+        "unit" : "kg"
+      }
     },
     "proxy": true,
     "siteFloorPlans": [
--- a/example/config/vanti-ugs/ui-config.json
+++ b/example/config/vanti-ugs/ui-config.json
@@ -22,6 +22,8 @@
         "clientId": "scos-opsui"
       }
     },
+    "wasteRecordsSource" : "van/uk/brum/ugs/devices/waste",
+    "wasteRecordUnit" : "kg",
     "ops": {
       "pages": [
         {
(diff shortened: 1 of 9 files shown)


ui/ops/src/api/ui/waste.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/**
+ * @param {Partial<PullWasteRecordsRequest.AsObject>} obj
+ * @return {PullWasteRecordsRequest|undefined}
+ */
+function pullWasteRecordsRequestFromObject(obj) {
+  if (!obj) return undefined;
+  const dst = new PullWasteRecordsRequest();
+  setProperties(dst, obj, 'name', 'pageToken', 'pageSize');
```
mattnathan, line 64: don't think pullWasteRecords has pageToken or pageSize, but does have updatesOnly

example/config/vanti-ugs/ui-config.json
```diff
@@ -22,6 +22,8 @@
         "clientId": "scos-opsui"
       }
     },
+    "wasteRecordsSource" : "van/uk/brum/ugs/devices/waste",
+    "wasteRecordUnit" : "kg",
```
mattnathan, line 26: As the waste page is in the ops section in the ui, maybe these properties should exist under `ops.waste` instead of at the top level in this config?

mattnathan (changes requested):

ui/ops/src/routes/ops/waste/WasteTable.vue
```diff
@@ -16,7 +16,7 @@
           {{ timestampToDate(item.wasteCreateTime).toLocaleString() }}
         </template>
         <template #item.weight="{ item }">
-          {{ item.weight.toFixed(2) }} {{ uiConfig.config.wasteRecordUnit ?? "kg" }}
+          {{ item.weight.toFixed(2) }} {{ uiConfig.config.ops.waste.unit ?? "kg" }}
```
mattnathan, line 19: you probably want to use `?.` here otherwise an absent ops or waste object will cause null pointer exceptions:
`uiConfig.config?.ops?.waset?.unit ?? 'kg'`.

Same further down the file

mattnathan (changes requested):

PR #357, opened 2025-02-26
traits/transport: add mock driver support with basic automation which simulates a lift

deanredfern26:


pkg/driver/mock/auto/transport.go
```diff
@@ -0,0 +1,120 @@
...
+					} else { // we are on the move,
+						if t.NextDestinations[0].Floor == "GF" {
+							t.MovingDirection = gen.Transport_DOWN
+							currentFloor--
+							t.ActualPosition.Floor = getFloorName(currentFloor)
+						} else {
+							t.MovingDirection = gen.Transport_UP
+							currentFloor++
+							t.ActualPosition.Floor = getFloorName(currentFloor)
+						}
```
dloveen-v, line 108: I imagine we'll get tripped up on this in future. Should `getFloorName()` be `getFloor()` and return the floor struct with all the right properties set?

dloveen-v (approved):

PR #359, opened 2025-02-26
ui/ops: add card on device sidebar for the transport trait

deanredfern26:
Add a basic card on the device sidebar for the transport trait. This implementation only displays the minimum properties that we would expect from a lift/elevator which implements the Transport trait.


mattnathan (approved):
Simple, but good change 👍

PR #361, opened 2025-03-04
app: add shared postgres storage pools

mattnathan:
Add a new mechanism for configuring storage at the sysconf level and having services (autos, systems, etc) reuse those shared storage clients/connections.

Update all the sc-bos services to use a shared connection if the service config doesn't specify a connection string.

@hexaglow I'm keen to know how this change might be related to your addition of sqlite.

Fixes SC-1033


dloveen-v:
One extension to this PR would be to be able to supply `connect configurations` to the global pool. We were previously initialising database pools with the default min / max connections for instance, which was fine. But it might be an idea to control things like min / max allowed concurrent connections when setting via the global pool to permit fine tuning on site deployments.

mattnathan:
> One extension to this PR would be to be able to supply `connect configurations` to the global pool. We were previously initialising database pools with the default min / max connections for instance, which was fine. But it might be an idea to control things like min / max allowed concurrent connections when setting via the global pool to permit fine tuning on site deployments.

@dloveen-v you should be able to do this via the URI configuration parameters. Unless you meant each user of the global pool specifying their own configuration options?

dloveen-v:
> @dloveen-v you should be able to do this via the URI configuration parameters. Unless you meant each user of the global pool specifying their own configuration options?

I didn't realise jackc/pgx supported this TBH. I thought it was using the default postgres connection string under the hood.

dloveen-v (approved):
LGTM. I haven't yet tested on a production client site as it means changing the config on site. I'll definitely think of a good rollout strategy for this that we can be confident in. Perhaps after upgrading sc-bos, we can turn off direct Postgres connections from the app/system configurations group by group?

pkg/auto/history/factory.go
```diff
@@ -68,7 +72,15 @@ func (a *automation) applyConfig(ctx context.Context, cfg config.Root) error {
 	var store history.Store
 	switch cfg.Storage.Type {
 	case "postgres":
-		pool, err := pgxutil.Connect(ctx, cfg.Storage.ConnectConfig)
+		var pool *pgxpool.Pool
+		var err error
+		if cfg.Storage.ConnectConfig.IsZero() {
+			// use admin pool (for now) as we know it will support create table operations
+			// todo: update store to support r, w, admin pools
```
deanredfern26, line 79: Have you made a ticket for this todo?

deanredfern26 (approved):
Looks good just a comment.

pkg/auto/history/factory.go
mattnathan: [SC-1038](https://vanti.youtrack.cloud/issue/SC-1038) Support r, w, admin connection pools for pgx
[SC-1039](https://vanti.youtrack.cloud/issue/SC-1039) Returns restricted privilege pools from PostgresStore

We do now

PR #366, opened 2025-03-05
ui/ops/energy: add EnergyHistoryCard widget

mattnathan:
The EnergyHistoryCard is a widget that shows stacked energy usage for a number of sources, intended to break down (for example) a building by floors to give a broader overview of energy use where ancestry is known.

This change also updates the ConsumptionCard widget as additional time utils have been written that this component can make use of.

I've added the chart to a new dashboard page for the ugs example so it should be easy to test with.

image

image

image

image

image

**Known Issues/Additions**

1. The chart is a little jumpy as data is loaded into it. This could be solved with a loading indicator of some kind but I feel like that can be added as a future change
2. There's no way to download the data in the chart. I intend for this to download all meter data over the selected range, rather than just the consumption you see in the charts. This requires updates to the history download feature to support a list of device names instead of a query object.
3. The chart can perform a lot of requests at the same time: num sources * edges (bars + 1) * 2, for a day showing 5 floors that would be 5 * 25 * 2 = 250 requests to the server. When testing locally this seems to be fine but as a future addition we might want to revisit how dashboard data is computed, cached, and delivered to the ui
4. If meter readings decrease things can get a little strange. While this shouldn't happen frequently on site, it's still something that could cause strange data for a short period. I'm not sure what the correct solution is for this so am treating it as _fix it when it happens_.
5. I would like to add a _compare to previous_ option so you can compare energy use between today and yesterday, or this month and last month (for example). This would double the requests but not increase other resource usage significantly (I believe). I think this can be added as a future addition though.
6. I would prefer weekly views to show Mon, Tue, Wed instead of 5th, 6th, 7th but this turned out to be harder to do that I was willing to spend. We can add it as a future addition.
7. When the x-axis crosses a period boundary (3rd-4th or 2024-2025, etc) I would prefer a tick that shows this crossing. Instead the current chart just shows 20:00, 23:00, 2:00, 5:00. Again this ended up being harder than I anticipated so am pushing this to a future addition. If a tick _happens_ to be rendered at this boundary it does the expected thing: 20:00, 22:00, 3rd Mar, 2:00, 4:00, etc

### DIFF
--- a/ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
+++ b/ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
@@ -2,7 +2,7 @@
   <v-card class="px-6 py-4">
     <v-toolbar class="chart-header" color="transparent">
       <slot name="title">
-        <v-toolbar-title class="pa-0 mr-auto">{{ props.title }}</v-toolbar-title>
+        <v-toolbar-title class="text-h4 pa-0 mr-auto">{{ props.title }}</v-toolbar-title>
       </slot>
       <v-btn
           icon="mdi-dots-vertical"
--- a/ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
+++ b/ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
@@ -1,0 +1,245 @@
+<template>
+  <v-card class="px-6 py-4">
+    <v-toolbar class="chart-header" color="transparent">
+      <slot name="title">
+        <v-toolbar-title class="pa-0 mr-auto">{{ props.title }}</v-toolbar-title>
+      </slot>
+      <v-btn
+          icon="mdi-dots-vertical"
+          size="small"
+          variant="text">
+        <v-icon size="24"/>
+        <v-menu activator="parent" location="bottom right" offset="8" :close-on-content-click="false">
+          <v-card min-width="24em">
+            <v-list density="compact">
+              <v-list-subheader title="Sources"/>
+              <v-list-item
+                  v-for="(item, index) in legendItems"
+                  :key="index"
+                  @click="item.onClick(item.hidden)"
+                  :title="item.text">
+                <template #prepend>
+                  <v-list-item-action start>
+                    <v-checkbox-btn :model-value="!item.hidden" readonly :color="item.bgColor" density="compact"/>
+                  </v-list-item-action>
+                </template>
+              </v-list-item>
+            </v-list>
+            <v-list>
+              <v-list-subheader title="Data"/>
+              <period-chooser-rows v-model:start="_start" v-model:end="_end" v-model:offset="_offset"/>
+            </v-list>
+          </v-card>
+        </v-menu>
+      </v-btn>
+    </v-toolbar>
+    <v-card-text>
+      <div class="chart__container">
+        <bar :options="chartOptions" :data="chartData" :plugins="[vueLegendPlugin, themeColorPlugin]"/>
+      </div>
+    </v-card-text>
+    <energy-tooltip :data="tooltipData" :edges="edges" :tick-unit="tickUnit" :unit="unit"/>
+  </v-card>
+</template>
+
+<script setup>
+import {
+  computeDatasets,
+  useExternalTooltip,
+  useThemeColorPlugin,
+  useVueLegendPlugin
+} from '@/dynamic/widgets/energy/chart.js';
+import {useDateScale} from '@/dynamic/widgets/energy/date.js';
+import EnergyTooltip from '@/dynamic/widgets/energy/EnergyTooltip.vue';
+import PeriodChooserRows from '@/dynamic/widgets/energy/PeriodChooserRows.vue';
+import {useDescribeMeterReading} from '@/traits/meter/meter.js';
+import {isNullOrUndef} from '@/util/types.js';
+import {BarElement, Chart as ChartJS, Legend, LinearScale, TimeScale, Title, Tooltip} from 'chart.js'
+import {startOfDay, startOfYear} from 'date-fns';
+import {computed, ref, toRef, toValue, watch} from 'vue';
+import {Bar} from 'vue-chartjs';
+import 'chartjs-adapter-date-fns';
+import {useMeterConsumption, useMetersConsumption} from './consumption.js';
+
+ChartJS.register(Title, Tooltip, BarElement, LinearScale, TimeScale, Legend);
+
+const props = defineProps({
+  title: {
+    type: String,
+    default: 'Energy Usage'
+  },
+  totalConsumptionName: {
+    type: String,
+    default: undefined,
+  },
+  totalProductionName: {
+    type: String,
+    default: undefined,
+  },
+  subConsumptionNames: {
+    type: [Array],
+    default: () => [],
+  },
+  subProductionNames: {
+    type: [Array],
+    default: () => [],
+  },
+  start: {
+    type: [String, Number, Date],
+    default: 'day', // 'month', 'day', etc. meaning 'start of <day>' or a Date-like object
+  },
+  end: {
+    type: [String, Number, Date],
+    default: 'day' // 'month', 'day', etc. meaning 'end of <day>' or a Date-like object
+  },
+  offset: {
+    type: [Number, String],
+    default: 0, // when start/End is 'month', 'day', etc. offset that value into the past, like 'last month'
+  }
+});
+
+// we assume here that all the meters share the same unit, so asking about any will be enough.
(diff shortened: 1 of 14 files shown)


mattnathan:
After an initial review I will create tickets for all those future addition / issues bits I mentioned. No point adding them if the review requires I fix them before the PR lands.

mattnathan:
There seems to be a bug where the chart isn't updating as time progresses, I'm looking into it

ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
```diff
@@ -0,0 +1,245 @@
...
+    plugins: {
+      tooltip: {
+        enabled: false,
+        external: tooltipExternal
+      },
+      legend: {
+        display: false, // we use a custom legend plugin and vue for this
+      }
+    },
+    scales: {
```
deanredfern26, line 165: On the y-axis, can we get a label for the units?

ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
```diff
@@ -0,0 +1,245 @@
...
+import {Bar} from 'vue-chartjs';
+import 'chartjs-adapter-date-fns';
+import {useMeterConsumption, useMetersConsumption} from './consumption.js';
+
+ChartJS.register(Title, Tooltip, BarElement, LinearScale, TimeScale, Legend);
+
+const props = defineProps({
+  title: {
+    type: String,
+    default: 'Energy Usage'
```
deanredfern26, line 69: The style for this header does not match the one for the Power widget, this one is much thinner and looks wierd.

ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
```diff
@@ -0,0 +1,245 @@
...
+      tooltip: {
+        enabled: false,
+        external: tooltipExternal
+      },
+      legend: {
+        display: false, // we use a custom legend plugin and vue for this
+      }
+    },
+    scales: {
+      y: {
```
deanredfern26, line 166: Can you either add a 'consumed' and 'produced' label on the top and bottom of the y-axis, or alternatively add a minus symbol for energy produced like the tooltop has? It might not be obvious to some people that consumption is above

deanredfern26 (changes requested):
Not sure if you can change this or if it is the graphing library, but the starts and end points don't match the bar. I've added a screenshot, the tooltip says 12:00 - 13:00 but the bar is drawn before 12. Looks a bit wierd.
image

Also, when all the values shown are way above 1,000 kWh, might it be better to convert the unit to MWh? image

mattnathan:
> Not sure if you can change this or if it is the graphing library, but the starts and end points don't match the bar. I've added a screenshot, the tooltip says 12:00 - 13:00 but the bar is drawn before 12. Looks a bit wierd.

This was intentional. The bar is drawn centred on the tick that most prominently represent the time range. I felt this was better than having the bars accurately span the time range on the x-axis.

deanredfern26:
> > Not sure if you can change this or if it is the graphing library, but the starts and end points don't match the bar. I've added a screenshot, the tooltip says 12:00 - 13:00 but the bar is drawn before 12. Looks a bit wierd.
> 
> This was intentional. The bar is drawn centred on the tick that most prominently represent the time range. I felt this was better than having the bars accurately span the time range on the x-axis.

OK, I would like a third opinion on this as to me this looks odd but it might very well be me who is the odd one

ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
mattnathan: I was hoping that the tooltip (which does have units) would be enough, in order to keep the noise to a minimum. I'll see if I can come up with a way to do it without sacrificing all the space we do for the current power-history chart.

ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
mattnathan: I think the minus in the tooltip is an issue with meters that go down instead of up. The tooltip should really only have positive numbers in it.

I'm not sure about adding labels for produced/consumed - the information is available in the tooltip and in the overflow menu and it's only the first time you see the chart that this is needed.

mattnathan:
> Also, when all the values shown are way above 1,000 kWh, might it be better to convert the unit to MWh?

I did consider this, but I think it's actually quite a lot of work to do this because the unit is a string (from the info trait) and any way I can think to convert the scale involves hacky string matching.

deanredfern26 (approved):
The comment I made on the position of the bars on the graph might just be matter of opinion, unless Loveen has a strong opinion lets just if there's any user feedbaclk. This is a very nice feature to have, nice work

ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
mattnathan: I imagine it's because the seeder starts at 0 each time sc-bos is restarted.

dloveen-v:
> The comment I made on the position of the bars on the graph might just be matter of opinion, unless Loveen has a strong opinion lets just if there's any user feedbaclk. This is a very nice feature to have, nice work

IMO, if we want to support downsampling the time series generated at https://github.com/vanti-dev/sc-bos/pull/366/files#diff-eaf0a267ff220f1b4499e2b58bf5be9f1889c0d21223f4cb0f6b19f82ce3b88aR29 we should create a ticket and prioritise once we have gathered user feedback. We don't know if users will compare this chart to values in CSV exports or assume the rendering is just an approximation.

mattnathan:
> > The comment I made on the position of the bars on the graph might just be matter of opinion, unless Loveen has a strong opinion lets just if there's any user feedbaclk. This is a very nice feature to have, nice work
> 
> IMO, if we want to support downsampling the time series generated at https://github.com/vanti-dev/sc-bos/pull/366/files#diff-eaf0a267ff220f1b4499e2b58bf5be9f1889c0d21223f4cb0f6b19f82ce3b88aR29 we should create a ticket and prioritise once we have gathered user feedback. We don't know if users will compare this chart to values in CSV exports or assume the rendering is just an approximation.

I'm not sure what downsampling you're referring to.

ui/ops/src/dynamic/widgets/energy/PeriodChooserRows.vue
```diff
@@ -0,0 +1,103 @@
...
+ * @typedef {Object} PeriodSelection
+ * @property {string} [text] - the text to display on the button
+ * @property {string} [value] - the value to set when selected
+ * @property {string} [icon] - the icon to display on the button
+ * @property {string} [tooltip] - the tooltip to display when hovering over the button
+ */
+
+/** @type {PeriodSelection[]} */
+const periodSelections = [
+  {text: 'D', value: 'day', tooltip: 'Day'},
```
dloveen-v, line 70: This is missing hour, I've just tried to add:
```
  {text: 'H', value: 'hour', tooltip: 'Hour'},
```

and it is supported.
Screenshot 2025-03-07 at 14 09 29

dloveen-v:
> > > The comment I made on the position of the bars on the graph might just be matter of opinion, unless Loveen has a strong opinion lets just if there's any user feedbaclk. This is a very nice feature to have, nice work
> > 
> > 
> > IMO, if we want to support downsampling the time series generated at https://github.com/vanti-dev/sc-bos/pull/366/files#diff-eaf0a267ff220f1b4499e2b58bf5be9f1889c0d21223f4cb0f6b19f82ce3b88aR29 we should create a ticket and prioritise once we have gathered user feedback. We don't know if users will compare this chart to values in CSV exports or assume the rendering is just an approximation.
> 
> I'm not sure what downsampling you're referring to.

I was referring to irregular meter readings sometimes needing smoothing (i.e. downsampling) to render.

If we get 1 meter reading between 12:00 -> 13:00, but 15 meter readings between 13:00 -> 14:00, I didn't know if the ui would be able to handle cases such as this.

I have reviewed and it is clear that the timestamps are being used to bin the meter readings into the interval to render so it's fine.

I now realise this was a separate concern and isn't the cause of rendering in the middle of the time tick vs at the start. I think this is just a preference thing and happy to see this go ahead in this case.

ui/ops/src/dynamic/widgets/energy/PeriodChooserRows.vue
mattnathan: I don't think hour is a range people are going to want for the chart, though I have added support in some of the utils as they are more broadly applicable than just this chart.

PR #367, opened 2025-03-06
devices: add string_in(_fold) query conditions

mattnathan:
The new condition can be used for queries like _get where name is one of foo or bar_ which will help with using the devices api (and the download api) in more places.


internal/manage/devices/server.go
```diff
@@ -258,3 +268,22 @@ func subMask(mask *fieldmaskpb.FieldMask, prefix string) *fieldmaskpb.FieldMask
...
+	}
+	for _, c := range q.Conditions {
+		if in := c.GetStringIn(); in != nil {
+			if len(in.Strings) > 100 {
+				return status.Errorf(codes.InvalidArgument, "condition string_in len > 100")
+			}
+		}
+		if in := c.GetStringInFold(); in != nil {
+			if len(in.Strings) > 100 {
+				return status.Errorf(codes.InvalidArgument, "condition string_in_fold len > 100")
```
dloveen-v, line 284: I'm curious why we chose 100.

mattnathan: I had to pick something and 100 seems like it would cover the use cases and not use too much memory. Can change later as needed. I was careful not to specify the exact limit in the proto file for that reason.

ui/ui-gen/proto/transport_pb.js
```diff
@@ -360,24 +360,24 @@ proto.smartcore.bos.Transport.prototype.toObject = function(opt_includeInstance)
  */
 proto.smartcore.bos.Transport.toObject = function(includeInstance, msg) {
   var f, obj = {
-actualPosition: (f = msg.getActualPosition()) && proto.smartcore.bos.Transport.Location.toObject(includeInstance, f),
-nextDestinationsList: jspb.Message.toObjectList(msg.getNextDestinationsList(),
+    actualPosition: (f = msg.getActualPosition()) && proto.smartcore.bos.Transport.Location.toObject(includeInstance, f),
```
dloveen-v, line 363: A bit nitpicky. but for visibility, @deanredfern26 and I are trying to figure out why these tab spaces appear when regenerating JS files here: https://github.com/smart-core-os/sc-api/pull/26#issuecomment-2703950440

dloveen-v (approved):

PR #370, opened 2025-03-07
ui/ops/power: use devices download api for the power history chart

mattnathan:
Previously it was computing the CSV content locally and providing the download that way. Now its reusing the DevicesApi DownloadLink feature to get all meter readings for the displayed period.

Blocked on #369, but can be reviewed now as the general approach won't change. I'll change the base once that PR lands


mattnathan:
I've rebased onto #371 as it includes a nice util we can use here

deanredfern26 (approved):
Tested it and it is working. Was a little slow when I tried, I think it would benefit from the 'your download will begin shortly' feedback like the other recent history download had. But as far as moving the work from the js to the device api this is acheived

mattnathan:
@deanredfern26 [SC-1050](https://vanti.youtrack.cloud/issue/SC-1050) 'Your download will start shortly' feedback to cover your feedback

PR #371, opened 2025-03-07
ui/ops/energy: add Export CSV... button to energy chart

mattnathan:
The energy chart now has a button that allows you to download the historical data that was used to compose the chart you are looking at.

I've split the PR into three commits:
1. API changes
2. Utility additions - these could also be used by #370 depending if this lands first
3. Adding the button

They are all in one PR to make reviewing a bit easier as it gets harder to test or understand context without the whole picture. I can split them out if desired.


ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
```diff
@@ -229,6 +233,52 @@ const chartData = computed(() => {
     ]
   };
 });
+
+// download CSV...
+const visibleNames = () => {
+  const names = [];
+  const namesByTitle = {
+    'Other Consumption': props.totalConsumptionName,
```
deanredfern26, line 241: Is this intentional that total and other both show the same thing?

deanredfern26 (approved):
Looks good, just the one comment.

Also created the below ticket as the data generated by the DB seeder is too crazily high. If we use this to demo smart core to people we need to change this
[SC-1048](https://vanti.youtrack.cloud/issue/SC-1048) Update DB seeder to make the meter data more realistic

ui/ops/src/dynamic/widgets/energy/EnergyHistoryCard.vue
mattnathan: Yeah, the chart data only ever shows Total OR Other, depending on whether there are sub consumptions available. This object just makes sure that our download includes that device name in both cases.

PR #372, opened 2025-03-10
Implement a modbus driver that doesn't support conversion to trait values (yet)

dloveen-v:
The modbus driver allows:
 - defining many devices each with their own connection type and connection (IP or serial)
 - each device connection can implement: status, emergency and energyStorage traits
 - clean up after tear down
 - configurable poll intervals

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
 	github.com/eclipse/paho.mqtt.golang v1.4.2
 	github.com/go-jose/go-jose/v4 v4.0.5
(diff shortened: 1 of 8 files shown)


deanredfern26 (changes requested):
@dloveen-v We have a datasheet for the generator here https://vanti.sharepoint.com/:b:/s/client-kir/EUuHzDkKrEhKhmF3Ts8C5RgBafIslCTTMD6-wGHHQtUJbw?e=JDa4he

I have also added a link to the ticket [19C-11](https://vanti.youtrack.cloud/issue/19C-11) Driver for power generator. Use this to develop the modbus driver against, we need to get rid of those TODOs before this driver is actually usable

PR #373, opened 2025-03-11
Modbus implementation

dloveen-v:
The bitwise operations are defined in the modbus specification for a particular device implementation. It should not be considered generic and may have to be extended in future to support different modbus device implementations.

Note: this was previously at https://github.com/vanti-dev/sc-bos/pull/372 but since I change the branch name (due to us supporting more than modbus over IP), GitHub wants a new pull request created.

### DIFF
--- a/pkg/driver/modbus/bitwise.go
+++ b/pkg/driver/modbus/bitwise.go
@@ -1,0 +1,107 @@
+package modbus
+
+import (
+	"github.com/smart-core-os/sc-api/go/traits"
+	"github.com/vanti-dev/sc-bos/pkg/gen"
+)
+
+/** TODO:
+This should be configurable per modbus device implementation
+since we don't have **many** modbus device specifications
+I will assume this is compatible with **all** modbus devices, for now
+*/
+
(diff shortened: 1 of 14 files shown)


pkg/driver/modbus/bitwise.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"github.com/smart-core-os/sc-api/go/traits"
+	"github.com/vanti-dev/sc-bos/pkg/gen"
+)
+
+/** TODO:
+This should be configurable per modbus device implementation
+since we don't have **many** modbus device specifications
+I will assume this is compatible with **all** modbus devices, for now
```
deanredfern26, line 11: This won't be compatible with all modbus devices. For example some have 32 bit values accross 2 registers, some use these 32bit values to encode a floating point etc. Can you move this driver from sc-bos to the 19 cornwall repo while it is being developed?

We can take a look at whether it is worth having a generic modbus driver, I know I wrote it in the ticket but thinking about it I am unsure as to whether it is actually worth it. It might be that we just have this driver specific for this device

PR #374, opened 2025-03-12
traits: add energy produced field to meter

deanredfern26:
A solar power inverter gives us both the energy consumed and energy produced by the device, we need to be able to model this using the meter trait.


proto/meter.proto
```diff
@@ -31,6 +31,10 @@ message MeterReading {
   // The end of the period usage is recorded relative to, i.e. the time the reading was taken.
   // This time might not be now if the device has a low resolution for taking readings.
   google.protobuf.Timestamp end_time = 3;
+  // Produced records the current energy produced value of the meter.
+  // The unit is unspecified, use device documentation or MeterInfo to discover it.
+  // This value is a total recorded between the start and end times.
+  float produced = 4;
```
dloveen-v, line 37: I don't trust that `produced` readings will **always** use the same unit as `usage` readings: e.g. kWh `usage` but `Wh` produced?

deanredfern26: Fair point it could happen

dloveen-v:
You may be able to catch all the bad usage of the old `unit` field with `go vet ./...`

deanredfern26:
> You may be able to catch all the bad usage of the old `unit` field with `go vet ./...`

Wow thanks Loveen, I did not know this existed!

deanredfern26:
There are vulnerability fails, I will fix them in a separate branch and merge that one first

dloveen-v:
Yes, please do also check the ops ui / space ui for any usage of `unit` too. It'll be hard to catch as the component will simply throw errors (I think)

ui/ops/src/traits/meter/meter.js
```diff
@@ -105,7 +105,7 @@ export function useMeterReading(value, support = null) {
   const _s = computed(() => toValue(support));
 
   const unit = computed(() => {
-    return _s.value?.unit ?? '';
+    return _s.value?.usageUnit ?? '';
   });
 
   const usage = computed(() => {
```
dloveen-v, line 111: Should this one be changed too?

deanredfern26: I did not change the name in the MeterReading only in the Describe object

dloveen-v (approved):
I have ran this locally and can see an unrelated JS error being thrown

PR #376, opened 2025-03-13
SC-1056: Fix ListRoles error when Role has no permissions

hexaglow:
When ListRoles encounters a Role with an empty permissions list, an internal error is returned.

Expected behavior is to return the Role with an empty permissions list.

Also fix bug with incorrect error code being returned by CreateRole and UpdateRole when the supplied `display_name` for create/update conflicts with an existing role.

### DIFF
--- a/internal/account/store.go
+++ b/internal/account/store.go
@@ -165,6 +165,27 @@ func (tx *Tx) GenerateServiceCredential(ctx context.Context, create queries.Serv
 	}, nil
 }
 
+func (tx *Tx) ListRolesAndPermissions(ctx context.Context, params queries.ListRolesAndPermissionsParams) ([]RoleAndPermissions, error) {
+	data, err := tx.Queries.ListRolesAndPermissions(ctx, params)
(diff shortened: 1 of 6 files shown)


internal/account/store.go
```diff
@@ -165,6 +165,27 @@ func (tx *Tx) GenerateServiceCredential(ctx context.Context, create queries.Serv
...
+		}
+	}
+	return roleAndPermissions, nil
+}
+
+type RoleAndPermissions struct {
+	Role          queries.Role
+	PermissionIDs []string
+}
+
```
mattnathan, line 188: This looks unrelated to the rest of the PR?

mattnathan (changes requested):

internal/account/store.go
mattnathan: Ah, you overloaded the default Queries function. I was surprised there were no changes in the Server based on this change

mattnathan (approved):

PR #378, opened 2025-03-18
driver/gallagher: add gallagher driver

deanredfern26:
This promotes the gallagher driver from the ocw project to sc-bos as it now needed for other projects.
I have edited the readme to remove the project specific stuff and also added an example.json file to help people use this driver in their projects.

### DIFF
--- a/pkg/driver/gallagher/alarms.go
+++ b/pkg/driver/gallagher/alarms.go
@@ -54,23 +54,21 @@ type Alarm struct {
 type SecurityEventController struct {
 	gen.UnimplementedSecurityEventApiServer
 
-	cleanupDuration time.Duration
-	client          *Client
-	logger          *zap.Logger
-	mu              sync.Mutex
+	client *Client
+	logger *zap.Logger
+	mu     sync.Mutex
 	// security events is a circular buffer, it always points to the oldest security event
 	lastEventTime  time.Time  // *gen.SecurityEvent
 	securityEvents *ring.Ring // *gen.SecurityEvent
 	updates        minibus.Bus[*gen.PullSecurityEventsResponse_Change]
 }
 
-func newSecurityEventController(client *Client, logger *zap.Logger, c time.Duration, n int) *SecurityEventController {
+func newSecurityEventController(client *Client, logger *zap.Logger, n int) *SecurityEventController {
 	return &SecurityEventController{
-		cleanupDuration: c,
-		client:          client,
-		logger:          logger,
-		lastEventTime:   time.Now().Add(-24 * time.Hour),
-		securityEvents:  ring.New(n),
+		client:         client,
+		logger:         logger,
+		lastEventTime:  time.Now().Add(-24 * time.Hour),
+		securityEvents: ring.New(n),
 	}
 }
 
--- a/pkg/driver/gallagher/alarms.go
+++ b/pkg/driver/gallagher/alarms.go
@@ -229,7 +229,9 @@ func (sc *SecurityEventController) ListSecurityEvents(_ context.Context, req *ge
 	// get the most recent ones first as that is what the UI expects
 	for i := start; i >= 0; i-- {
 		se := sc.securityEvents.Move(i)
-		response.SecurityEvents = append(response.SecurityEvents, se.Value.(*gen.SecurityEvent))
+		if se.Value != nil {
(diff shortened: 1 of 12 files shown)


pkg/driver/gallagher/driver.go
```diff
@@ -0,0 +1,134 @@
...
+		return dc.run(ctx, cfg.RefreshDoors, announcer, cfg.ScNamePrefix)
+	})
+
+	sc := newSecurityEventController(client, d.logger, cfg.SecurityEventExpiry.Duration)
+	announcer.Announce(cfg.ScNamePrefix, node.HasTrait(securityevent.TraitName, node.WithClients(gen.WrapSecurityEventApi(sc))))
+	grp.Go(func() error {
+		return sc.run(ctx, cfg.RefreshAlarms)
+	})
+
+	occupancyCtrl := newOccupancyEventController(client, d.logger, cfg.RefreshOccupancyInterval.Or(defaultOccupancyRefreshInterval))
```
dloveen-v, line 103: I would've put this announcement behind a config flag if I knew it would be promoted.

I think it should be behind a flag as the parameters sent and parsing is unique to some sites.

pkg/driver/gallagher/alarms.go
```diff
@@ -0,0 +1,276 @@
...
+	if req.PageSize > 1000 {
+		req.PageSize = 1000
+	}
+
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+	var response gen.ListSecurityEventsResponse
+	// get the most recent ones first as that is what the UI expects
+	for i := start; i >= 0; i-- {
+		response.SecurityEvents = append(response.SecurityEvents, sc.securityEvents[i])
```
dloveen-v, line 254: Just a nitpicking note:

I found a slightly more idiomatic way to do this (hold x number of events and replace the oldest on add).

https://pkg.go.dev/container/ring 
I wanted to hear your opinions on this implementation.

deanredfern26: Is the suggestion to use a ring instead of a slice for the field `sc.securityEvents`?

pkg/driver/gallagher/alarms.go
```diff
@@ -0,0 +1,256 @@
...
+		req.PageSize = 1000
+	}
+
+	sc.mu.Lock()
+	defer sc.mu.Unlock()
+	var response gen.ListSecurityEventsResponse
+	// get the most recent ones first as that is what the UI expects
+	for i := start; i >= 0; i-- {
+		se := sc.securityEvents.Move(i)
+		response.SecurityEvents = append(response.SecurityEvents, se.Value.(*gen.SecurityEvent))
```
dloveen-v, line 234: You may want to check `se.Value` is not nil. I know the events are already initialized at the point of calling, but if the Gallagher API takes a while to resolve this could panic.

pkg/gentrait/securityevent/model.go
```diff
@@ -73,15 +76,19 @@ func (m *Model) GenerateSecurityEvent(time *timestamppb.Timestamp) (*gen.Securit
...
 func (m *Model) ListSecurityEvents(start, count int) []*gen.SecurityEvent {
 
 	var events []*gen.SecurityEvent
 	// reverse to retrieve the latest events first
+
+	e := m.allSecurityEvents.Move(start)
+	events = append(events, e.Value.(*gen.SecurityEvent))
 	for i := start - 1; i >= 0; i-- {
-		events = append(events, m.allSecurityEvents[i])
+		e.Next()
```
dloveen-v, line 90: This should be `e = e.Next()` I believe

pkg/driver/gallagher/alarms.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+type Alarm struct {
+	AlarmPayload
+}
+
+type SecurityEventController struct {
+	gen.UnimplementedSecurityEventApiServer
+
+	cleanupDuration time.Duration
```
dloveen-v, line 57: This needs removing

pkg/gentrait/securityevent/model.go
deanredfern26: Yes the test client reported 100 values were retrieved, but didn't actually print out the contents, so this got missed, thank you. I have updated the test client to catch this kind of thing if it is used to test against other SecurityEvent implementors

PR #380, opened 2025-03-20
ui/ops: fix useCollection race

mattnathan:
These changes combine to fix the race we have in useCollection that causes rapid request changes to show old results.

The fix comes in the form of a new `useAction` utility that _correctly_ handles the request change vs async request race and converting the list part of useCollection to use this utility. Future changes will make use of `useAction` too, which is why it's been factored out.

While the refresh func isn't required for this fix, it is part of a wider change for which this fix has been extracted. The change was small enough that I thought it better to include it here than to create dependent PRs. Especially as the code changes once the actual fix lands.

Fixes SC-985

### DIFF
--- a/ui/ops/src/composables/action.js
+++ b/ui/ops/src/composables/action.js
@@ -1,0 +1,64 @@
+import {newActionTracker} from '@/api/resource.js';
+import {reactive, ref, toRefs, toValue, watch} from 'vue';
+
+/**
+ * @typedef {import('@/api/resource.js').ActionTracker<Res>} UseActionResponse
+ * @property {function(): void} refresh - re-executes the action in the background
+ * @template Res
+ */
+
+/**
+ * Calls action whenever request changes, returning a reactive tracker to monitor the action.
+ * If request is null or undefined, the action is not called and the trackers response is reset.
+ * If request changes frequently, only the latest request is processed.
+ *
+ * @param {import('vue').MaybeRefOrGetter<Partial<Req> | null | undefined>} request
+ * @param {function(Req, ActionTracker<Res>): Promise<Res>} action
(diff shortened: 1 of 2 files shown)


ui/ops/src/composables/action.js
```diff
@@ -0,0 +1,64 @@
...
+  const tracker = reactive(/** @type {ActionTracker<Res>} */ newActionTracker());
+
+  // What follows is a race safe way to perform requests whenever request changes.
+  // We queue up tasks (calls to action) and only process the latest one.
+  const activeTask = ref(null); // the task we are awaiting
+  const nextTask = ref(null); // our queue, except we only care about the tip
+  watch(() => toValue(request), (request) => {
+    nextTask.value = {request};
+  }, {immediate: true});
+  watch(nextTask, (task) => {
```
dloveen-v, line 31: I am thinking the `nextTask` watcher should be before the previous `watch` block? As the previous is `immediate`, would the previous watch block's callback execute immediately or would it bind the `nextTask` watch block first?

deanredfern26 (approved):

ui/ops/src/composables/action.js
mattnathan: As both watchers are immediate the order doesn't really matter from a semantic pov, by the time the first callback of the nextTask watcher is called the request watcher callback has already run to fill nextTask.value.

Swapping the order would allow the nextTask watcher to remove the immediate option but at the cost of reading the function out of data-flow order. Data flows from request to nextTask to activeTask, which I think should be reflected in the read order of the watchers.

mattnathan: I guess this is also equivalent:

```js
const nextTask = ref({request: toValue(request)});
watch(() => toValue(request), (request) => {
  nextTask.value = {request};
}); // not immediate
```

But to me felt like duplicated code

PR #381, opened 2025-03-20
SC-1060: restructure gRPC API for Service Account handling

hexaglow:
Instead of a separate ServiceCredential resource which presented headaches for permission scoping, put details of the service account's credentials into the Account resource.
It is also desirable to more clearly encode that service account credentials have one primary secret, and the previous secret value which may remain valid for a period of time after a new secret is generated, to facilitate zero-downtime rotation.

Change Account structure to put fields that are specific to one type of account into their own `oneof` variant. 

The database schema has been updated to better match this new resource structure. I have also added a test to show the database migration preserves data properly.


mattnathan (changes requested):

mattnathan (approved):

hexaglow:
Just need to figure out why the tests are failing in CI but not locally

hexaglow:
Works fine after rebasing, not sure why but I'm not complaining

PR #383, opened 2025-03-20
traits: add service ticket trait

deanredfern26:
Adds a new 'service ticket' trait for integrating with external systems which have the concept of ticketing, such as a CAFM.
I have restricted this to have create and update functionality for tickets. We are required to be able to raise tickets in external systems, we do not want to be trying to display and manage all the tickets in an external system via smart core.


proto/service_ticket.proto
```diff
@@ -0,0 +1,70 @@
...
+// CreateTicketRequest is the request to create a ticket in the external system.
+message CreateTicketRequest {
+  string name = 1;
+  Ticket ticket = 2;
+}
+
+// UpdateTicketRequest is the request to update a ticket in the external system. The ticket ID must be set.
+message UpdateTicketRequest {
+  string name = 1;
+  Ticket ticket = 2;
```
dloveen-v, line 68: Do we want a `status` field? i.e. update ticket to status = fulfilled.

proto/service_ticket.proto
```diff
@@ -0,0 +1,70 @@
...
+  enum Severity {
+      SEVERITY_UNSPECIFIED = 0;
+      EMERGENCY = 1;
+      CRITICAL = 2;
+      URGENT = 3;
+      HIGH = 4;
+      MEDIUM = 5;
+      LOW = 6;
+  }
+  Severity severity = 6;
```
dloveen-v, line 56: Maybe additional details url field is also required? For posting images or videos?

dloveen-v: I mean, I am just trying to avoid having to parse descriptions and details in future.

proto/service_ticket.proto
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Ticket represents a service ticket in a third party system.
+message Ticket {
+  // Id is blank when creating a ticket, and is filled in by the external system. The ID is then used to update the ticket.
+  string id = 1;
+  // Short description of the issue being reported.
+  string description = 2;
+  // Full details on the issue being reported. This should include all the available information to help resolve the issue.
+  string details = 3;
```
mattnathan, line 24: Can we adjust these slightly. I think `summary` and `description` (instead of `description` and `details`) would align better with other uses of description in our apis.

proto/service_ticket.proto
```diff
@@ -0,0 +1,70 @@
...
+    NON_EMERGENCY_REPAIR = 6;
+    PLUMBING = 7;
+    WASTE = 8;
+    RECYCLING = 9;
+    SECURITY = 10;
+    URGENT_REPAIR = 11;
+    OBSERVATION = 12;
+    PORTERAGE = 13;
+    SPACE_PREPARATION = 14;
+  }
```
mattnathan, line 44: Considering the generic nature of the rest of the Ticket type (reporter_name for example), I'm surprised by the specificity of this list of classifications. Where did this list come from?

proto/service_ticket.proto
mattnathan: _If_ this is added, it's better to represent state changes as rpc methods instead of Update({state=done})

proto/service_ticket.proto
mattnathan: I like this idea, I think a `string external_url` (or something) would be useful.

proto/service_ticket.proto
mattnathan: Thinking of ticketing systems in general, I expect this list to be adjustable by the user of that system so an enum is probably not the correct choice for this information.

proto/service_ticket.proto
mattnathan, line 56: Considering my comment about the classification data, I also expect severity to be adjustable in the external system. Maybe an enum isn't correct here either.

proto/service_ticket.proto
deanredfern26: I am deliberately leaving out status information at this stage and the ability to manage tickets, I think the tickets should be managed via the CAFM and not smartcore. I am worried that going down this road will give the wrong impression of smart core as a product, I don't believe we want to be a CAFM. I might be wrong though

mattnathan (changes requested):

proto/service_ticket.proto
deanredfern26: This list came from the list of possible ticket classifications from the Invida API we are integrating with: 
https://api-40lh-demo.invida.co.uk/api/integration/v1_4/Parameter/Classification/List

proto/service_ticket.proto
deanredfern26: Again this has come from the available `Impact` we can use for the Invida system https://api-40lh-demo.invida.co.uk/api/integration/v1_4/Parameter/Impact/List

proto/service_ticket.proto
mattnathan: Can we make it part of an info service or add ListCategories and ListSeverities to the api service or something. Either that or remove any doubt that this trait isn't a trait and instead is a system specific service

deanredfern26: Yeah it is better suited to an info service really, I adapted these from the Invida list to make them to be more generic tickets that buildings would have, but makes it more extensible to put this an Severity in an info service

dloveen-v (approved):
LGTM. I am conscious we have omitted any awareness of time or dates as a concept in the ticketing system at the smart-core side. Correct me if I'm wrong - I believe these fields are intentionally omitted.

deanredfern26:
> LGTM. I am conscious we have omitted any awareness of time or dates as a concept in the ticketing system at the smart-core side. Correct me if I'm wrong - I believe these fields are intentionally omitted.

Yes, the ticket created time I intend to be populated by the backend when sending the API request to the implementing system.

I did actually add a `reported_time` field but I decided against it so removed it.
For automations, `reported_time` will always be the same as `created_time`, for manual entries this could be different, but again differentiating between these things feels a little out of scope

mattnathan (approved):

PR #384, opened 2025-03-21
auto/exporthttp: swap to cron schedules instead of time intervals

dloveen-v:
Since time intervals may keep resetting if the controllers are restarted, using cron schedules is a more reliable way to ensure "at least 1 execution per schedule". Intervals are reset multiple times before the subsequent execution schedule meaning the automation reporting is unreliable.

I have also buffered the job chan in case an execution takes longer than the smallest unit schedule interval - by executing the job in a go-routine and buffering the multiplexed job chan, we can ensure executions which are scheduled around the same time execute concurrently.

### DIFF
--- a/cmd/tools/test-exporthttp/main.go
+++ b/cmd/tools/test-exporthttp/main.go
@@ -83,7 +83,7 @@ func main() {
 	}()
 
 	// wait for all automations in exporthttp to finish
-	time.Sleep(15 * time.Second)
+	time.Sleep(3 * time.Minute)
 }
 
 const (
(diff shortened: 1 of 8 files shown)


deanredfern26 (approved):
LGTM. Not sure why you are using those non-standard 0/1 schedules instead of just `* * * * *` but they work and are only in the test app

dloveen-v:
> LGTM. Not sure why you are using those non-standard 0/1 schedules instead of just `* * * * *` but they work and are only in the test app

That's exactly what I was testing :) non-standard syntax

PR #385, opened 2025-03-24
ui/ops: add account management pages

mattnathan:
Add new role and account management pages to the Ops UI. These pages allow the creation, modification, and deletion of roles and their associated permissions, and accounts and their assigned roles.

image

The account management feature set isn't 100% complete, here is the feature set included in this PR:

1. [x] View existing roles
2. [x] Create new roles
3. [x] Edit the description of a role
4. [x] Edit the permissions a role has
5. [x] See details about a permission
6. [x] Delete a role
7. [x] View existing accounts
8. [x] See service accounts and user accounts
9. [x] Create new accounts of both types
10. [x] Edit the username of user accounts
11. [x] Edit the description for service accounts
12. [x] Assign new roles to accounts - both individually and in bulk
13. [x] Remove roles from accounts
14. [x] See more details about roles an account has assigned
15. [x] Delete accounts
16. [x] Rotate service account secrets
17. [x] Specify an expiry for old service account secrets
18. [ ] Change a users password
19. [ ] Change your own users password
20. [ ] Edit your own users display name
21. [ ] Hide pages/elements based on the current users permissions

### DIFF
--- a/example/config/vanti-ugs/app.conf.json
+++ b/example/config/vanti-ugs/app.conf.json
@@ -1,5 +1,10 @@
 {
   "name": "van/uk/brum/ugs/eg-ac-01",
+  "metadata": {
+    "appearance": {
+      "title": "Vanti UGS 01"
+    }
+  },
   "drivers": [
     {
       "name": "van/uk/brum/ugs/eg-ac-01/driver/basement",
--- a/example/config/vanti-ugs/ui-config.json
+++ b/example/config/vanti-ugs/ui-config.json
@@ -2,7 +2,8 @@
   "features": {
     "auth": {
       "third-party": true,
-      "accounts": true
+      "accounts": true,
+      "roles": true
     },
     "device": true,
     "devices": true,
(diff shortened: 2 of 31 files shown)


hexaglow (changes requested):
There doesn't appear to be a way to edit the display name of a role. Could you implement or add to the backlog?

The plain link that shows the assigned roles in the account list and account sidebar doesn't look good. The scope type, scope value and role are concatenated without separators. Could something similar to the chips from the 'Grant Role' popup be used instead here?

mattnathan:
Screenshot 2025-06-12 at 10 01 18
 It's inspired by the YT groups list

hexaglow:
I understand there's a space there, but I don't like the way there's no visual distinction between the scope and role name. I don't like it in the YouTrack example either. Personal preference. 
I agree that styling them as links doesn't look the best.

Are you going to do the remaining tasks as part of this PR or separately?

hexaglow:
The visual can be improved at a later date, that's fine.
So the only outstanding point is being able to edit role display names.

hexaglow (approved):
Functionality meets the requirements.
I haven't done a code-level review, Vue is not my area of expertise.

PR #387, opened 2025-03-26
hub: check pgxpool != nil before closing

deanredfern26:


pkg/system/hub/factory.go
```diff
@@ -109,7 +109,9 @@ func (s *System) applyConfig(ctx context.Context, cfg config.Root) error {
 			if err != nil {
```
mattnathan, line 109: I think the correct fix here is `if err == nil {`

mattnathan (changes requested):

PR #391, opened 2025-04-02
traits: add anpr camera trait

deanredfern26:
Adds a trait for ANPR camera devices. Exposes the minimum info we need from the cameras: time of detection, reg plate and also some optional fields which are nice if the ANPR camera supports them.


proto/anpr_camera.proto
```diff
@@ -0,0 +1,71 @@
...
+
+option go_package = "github.com/vanti-dev/sc-bos/pkg/gen";
+
+import "google/protobuf/field_mask.proto";
+import "google/protobuf/timestamp.proto";
+
+
+// AnprCameraApi describes the capability to retrieve detection events from ANPR cameras.
+service AnprCameraApi {
+  rpc GetEvent(GetLastEventRequest) returns (AnprEvent) {}
```
hexaglow, line 13: The name of the RPC and request message don't agree

proto/anpr_camera.proto
```diff
@@ -0,0 +1,71 @@
...
+  // The time the detection events occurred.
+  google.protobuf.Timestamp event_time = 1;
+  // The registration plate detected.
+  string registration_plate = 2;
+  // Country of the detected registration plate.
+  string country = 3;
+  // Optional. The area of the detected registration plate. i.e. for UAE this could be Abu Dhabi, Dubai, etc.
+  optional string area = 4;
+  // Optional. The confidence level of the detection as a percentage. If omitted, means unknown confidence level.
+  optional int32 confidence = 5;
```
hexaglow, line 29: Precent is to use a floating-point type for confidence
https://github.com/smart-core-os/sc-api/blob/5c5bb79965fb358b7ec82475d4f2a0c5404c9288/protobuf/traits/occupancy_sensor.proto#L51
Although I think a double is excessive, a float would be fine here

proto/anpr_camera.proto
```diff
@@ -0,0 +1,71 @@
...
+  // Optional. The type of plate, e.g. standard, personalised, etc.
+  optional string plate_type = 6;
```
hexaglow, line 31: Is this field intended to have a specific set of allowed values, or could it be anything?

hexaglow (changes requested):
The way `optional` is being used isn't consistent with how it's used in the other sc-api services.
 - For message fields, `optional` has no effect and we have been indicating if the field is required in documentation
 - For string fields, we use the default `""` value to signal absence. Using `optional` introduces a distinction between unset and empty, which we don't do in other APIs
 - For numeric fields, we do use `optional` where it's necessary to tell the diference between 0 and absent. Keep using optional in this situation.

I'm conflicted about `GetEvent`. On the one hand, getting the latest event is often a useful operation, but this is a limited operation, compared to a AnprCameraHistory service. I've had a look at other traits that expose event streams (rather than states), to see how they handle it:
- EnterLeave does have GetEnterLeaveEvent, but this is misleading, as this RPC will not return any event details, only totals.
- The Press trait does have `most_recent_gesture` as a property of PressedState, however this is also a state-based API, which can also represent in-progress gestures. 
Do you have a use case for GetEvent?

deanredfern26:
> The way `optional` is being used isn't consistent with how it's used in the other sc-api services.
> 
>     * For message fields, `optional` has no effect and we have been indicating if the field is required in documentation
> 
>     * For string fields, we use the default `""` value to signal absence. Using `optional` introduces a distinction between unset and empty, which we don't do in other APIs
> 
>     * For numeric fields, we do use `optional` where it's necessary to tell the diference between 0 and absent. Keep using optional in this situation.
> 
> 
> I'm conflicted about `GetEvent`. On the one hand, getting the latest event is often a useful operation, but this is a limited operation, compared to a AnprCameraHistory service. I've had a look at other traits that expose event streams (rather than states), to see how they handle it:
> 
>     * EnterLeave does have GetEnterLeaveEvent, but this is misleading, as this RPC will not return any event details, only totals.
> 
>     * The Press trait does have `most_recent_gesture` as a property of PressedState, however this is also a state-based API, which can also represent in-progress gestures.
>       Do you have a use case for GetEvent?

I have changed it from a `get` to `list history`, makes sense as we can cover both cases with this implementation anyway.

proto/anpr_camera.proto
deanredfern26: This could really be anything, depending on the vendor, i.e. the HikVision one has things like 'armed police helicopter licence plate' and 'Hong Kong China/Macao China entrance vehicle' which can be used here. I added it as might be useful if we can use this field to specify if the detected vehicle is an emergency services vehicle or somehting

proto/anpr_camera.proto
```diff
@@ -0,0 +1,89 @@
...
+
+// AnprEvent describes a registration plate detection event from an ANPR camera.
+// Includes registration plate information but also vehicle information if the camera supports it.
+message AnprEvent {
+  // The time the detection events occurred.
+  google.protobuf.Timestamp event_time = 1;
+  // The registration plate detected.
+  string registration_plate = 2;
+  // Country of the detected registration plate.
+  string country = 3;
```
hexaglow, line 25: Could we state that this is e.g. an ISO 3166-1 country code so it's machine-readable?

proto/anpr_camera.proto
```diff
@@ -0,0 +1,89 @@
...
+
+option go_package = "github.com/vanti-dev/sc-bos/pkg/gen";
+
+import "google/protobuf/field_mask.proto";
+import "google/protobuf/timestamp.proto";
+
+
+// AnprCameraApi describes the capability to retrieve detection events from ANPR cameras.
+service AnprCameraApi {
+  rpc ListAnprEvents(ListAnprEventsRequest) returns (ListAnprEventsResponse) {}
```
hexaglow, line 13: This is OK but have a look at e.g. the AirTemperatureHistory service which might be a better thing to replicate.
- It's a separate service so we don't require everything that implements AnprCameraApi to store historical records.
- It supports a specific time slice for history queries.

hexaglow (approved):
I can see opportunities to better specify the allowed values returned by the APIs, which would make it more useful for machine consumers of the info (better normalisation, basically).
At the moment, the API looks useful for driving a user interface, less so for analysis. 
As I don't think that's in scope for now, I'll let you decide if this needs improving. 

Similar thing with the ListAnprEvents API, if you change it a bit then it's more consistent with our other APIs but it's fine as-is.

PR #392, opened 2025-04-04
auto/exporthttp: more reliable interval calculations for filterTime

dloveen-v:
There is a race condition between passing a job that is now scheduled in the channel, and re-assigning the previous execution time field.


dloveen-v:
I have figured out why this bug is happening.

>Operating systems provide both a “wall clock,” which is subject to changes for clock synchronization, and a “monotonic clock,” which is not. The general rule is that the wall clock is for telling time and the monotonic clock is for measuring time. Rather than split the API, in this package the Time returned by [time.Now](https://pkg.go.dev/time#Now) contains both a wall clock reading and a monotonic clock reading; later time-telling operations use the wall clock reading, but later time-measuring operations, specifically comparisons and subtractions, use the monotonic clock reading.

[This line](https://github.com/vanti-dev/sc-bos/pull/392/files#diff-984d1f5816d9ca68e189296ec4bb356adeaa630d3c3181c15d145b1879a625b4R105) can return a `chan time.Time` which omits  _**before**_ the actual expected timestamp. This means the
```
for {
    select {
         ....
   }
}
```
can execute more than once within a single interval. Since the chan omits before the actual expected timestamp, the difference is down to nanoseconds (between previous and current execution timestamps). The `ListMeterReadingHistory` only has a tiny time-window to lookback for searching for meter records leading to 0kWh output for energy.

deanredfern26:
How is this looking on site, are you happy with how it is working with the latest commit?

deanredfern26 (approved):

deanredfern26:
I did already approve this, is the merge option not showing up for you? I will re approve anyway

deanredfern26 (approved):

PR #393, opened 2025-04-07
driver/mock: don't produce UNSPECIFIED access grants

mattnathan:


pkg/driver/mock/auto/access.go
```diff
@@ -14,6 +15,8 @@ import (
 
 func Access(model *accesspb.Model) service.Lifecycle {
 	grants := maps.Values(gen.AccessAttempt_Grant_value)
+	slices.Sort(grants)
```
dloveen-v, line 18: Why don't we just filter on zero-value Grants explicitly? Seems less error prone

mattnathan: I guess the reason the code looks like this was because I had a local change that restricted the grant types to [1..4] to more quickly test a grant based ui. That change was trivial for this code structure, the PR exists because I thought the UNSPECIFIED part of it was generally useful.

PR #395, opened 2025-04-24
util/pull: reduce incorrect "pull now succeeding" messages

mattnathan:
The old logic was overly complicated with the aim of printing the now succeeding message as soon as we could, but had the drawback that it'd get it wrong frequently. This was compounded by the success timer counting the backoff delay too so as attempts increased it became more likely to incorrectly report success.

Aside from the logging, false success messages were also resetting the backoff causing excess churn.

The new code is much simpler and also more reliable at the cost of delaying the success log message, which we can live with.


mattnathan:
FYI I have deployed this to site and it has significantly reduced the strain on the DALI driver and log output but otherwise appears to be working as intended

deanredfern26 (approved):

PR #396, opened 2025-04-25
all: replace grpc.Dial/Context calls with grpc.NewClient

mattnathan:
grpc.Dial/Context has been deprecated, as have options grpc.WithBlock and friends.

Most of the work here is about also removing the grpc.WithBlock and related options which aren't usable with NewClient.

There were some logic changes needed in the hub packages, but I don't think they'll have any noticeable effects on our deployments. Mostly making sure we perform at least one RPC before assuming we have the server cert captured.

The surprising thing was in node/remote.go which required a Connect call to build the resolver before the address could be updated. Without this the code would panic.

Finally, it appears that NewClient doesn't accept an empty target string, where Dial and DialContext did, at least in the case where the Dialler was replaced. In these cases I've used a localhost address just to satisfy, with minimal overhead, the requirements of the resolver. This code is only present in tests.

### DIFF
--- a/pkg/app/servers_test.go
+++ b/pkg/app/servers_test.go
@@ -116,8 +116,7 @@ func TestServeGRPC(t *testing.T) {
 		serveErr <- err
 	}()
 
-	conn, err := grpc.DialContext(clientCtx, addr,
-		grpc.WithBlock(),
+	conn, err := grpc.NewClient(addr,
 		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{
 			InsecureSkipVerify: true,
 		})),
--- a/pkg/system/hub/remote/enroll.go
+++ b/pkg/system/hub/remote/enroll.go
@@ -59,7 +59,8 @@ func Enroll(ctx context.Context, enrollment *gen.Enrollment, authority pki.Sourc
 	client := gen.NewEnrollmentApiClient(conn)
 	// any api call will do to force the connection to be established (or fail)
 	_, err = client.GetEnrollment(ctx, &gen.GetEnrollmentRequest{})
-	if err != nil {
+	if err != nil && status.Code(err) != codes.NotFound {
+		// NotFound is expected if the remote node is not enrolled yet, ignore that error
 		return nil, err
 	}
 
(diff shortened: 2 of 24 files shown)


pkg/app/servers_test.go
```diff
@@ -116,8 +116,7 @@ func TestServeGRPC(t *testing.T) {
 		serveErr <- err
 	}()
 
-	conn, err := grpc.DialContext(clientCtx, addr,
-		grpc.WithBlock(),
```
hexaglow, line 120: The old test had `WithBlock`, should we be using `WaitForReady` here?

pkg/system/hub/remote/enroll.go
```diff
@@ -49,16 +49,20 @@ func Enroll(ctx context.Context, enrollment *gen.Enrollment, authority pki.Sourc
...
-		grpc.WithBlock(),
-		grpc.FailOnNonTempDialError(true),
-		grpc.WithReturnConnectionError(),
 	)
 	if err != nil {
 		return nil, err
 	}
 
+	client := gen.NewEnrollmentApiClient(conn)
+	// any api call will do to force the connection to be established (or fail)
```
hexaglow, line 60: `WaitForReady` would achieve the same and behave more similarly to the previous implementation, no?

hexaglow (changes requested):
The gateway e2e test is now failing.
It appears to work on the previous commit.

pkg/app/servers_test.go
mattnathan: I considered it, but all that would achieve would be surfacing the error earlier, the next thing the test does is an RPC so any connection errors would show up there anyway

pkg/system/hub/remote/enroll.go
mattnathan: Unfortunately there's no way to accurately replicate the old behaviour without using an RPC. The trouble is there's no public way to get the connection error that cause the conn status to enter TransientFailure. This means we can't distinguish between temporary and permanent errors, or even know what the error is.

I figured it would be safer to use the recommended approach (do an RPC) here instead of either masking the underlying error (replace it with client.ErrTransientFailure) or failing too early (because we can't recognise temp/perm errors).

Do you think I need to say that anywhere?

mattnathan:
> The gateway e2e test is now failing.

Hmm, it was working locally. Maybe it was cached... I'll investigate next week

pkg/system/hub/remote/enroll.go
hexaglow: No, I think it's fine

hexaglow (approved):

PR #397, opened 2025-04-25
traits: add sound sensor trait

deanredfern26:
adds a trait to model a sound/noise measurement device.
alternative implemetation to adding this to the microphone trait https://github.com/smart-core-os/sc-api/pull/30
if this one in sc-bos is preferred I will close the PR in sc-api


proto/sound_sensor.proto
```diff
@@ -0,0 +1,59 @@
...
+}
+
+service SoundSensorInfo {
+  rpc DescribeSoundLevel(DescribeSoundLevelRequest) returns (SoundLevelSupport) {}
+}
+
+// SoundLevel represents a sound/noise level.
+message SoundLevel {
+  // The measured sound/noise level, usually in dBA or dB. Use Info service to check.
+  float sound_level = 1;
```
hexaglow, line 24: Best to make this `optional` I'd think, as 0 is a reasonable value in dB and we may need a way to indicate 'value unknown' as we do for other traits.

proto/sound_sensor.proto
```diff
@@ -0,0 +1,59 @@
...
+}
+
+message SoundLevelSupport {
+  // How a named device supports read/write/pull apis
+  smartcore.types.ResourceSupport resource_support = 1;
+  // The unit associated with the sound_level value
```
hexaglow, line 30: Please specify a list of standard values for this and their meanings.

hexaglow (changes requested):

proto/sound_sensor.proto
```diff
@@ -0,0 +1,60 @@
...
+}
+
+service SoundSensorInfo {
+  rpc DescribeSoundLevel(DescribeSoundLevelRequest) returns (SoundLevelSupport) {}
+}
+
+// SoundLevel represents a sound/noise level.
+message SoundLevel {
+  // The measured sound/noise level, usually in dBA or dB. Use Info service to check.
+  // Optional. 0 means no sound or possibly broken sensor. Omitted means unknown.
```
hexaglow, line 24: 0 isn't a good choice to represent a broken sensor or no sound IMO.
0 is a reading you could in theory obtain from a working sound sensor in a very quiet space.
No sound would be whatever the noise floor of the sensor is. Or '-Infinity' I suppose.
Broken sensor should also result in an unset field if we can detect that it's broken.

proto/sound_sensor.proto
```diff
@@ -0,0 +1,60 @@
...
+  optional float sound_pressure_level = 1;
+}
+
+message SoundLevelSupport {
+  // How a named device supports read/write/pull apis
+  smartcore.types.ResourceSupport resource_support = 1;
+  // The unit associated with the sound_level value in dBA or dB.
```
hexaglow, line 31: The referenced property is named `sound_pressure_level`. 
Comment is phrased in an unclear way. Suggestion:

    // The unit associated with the sound_pressure_level value.
    // Known units:
    //   - dBA - decibels SPL, A-weighted
    //   - dB - decibels SPL, unweighted

hexaglow (changes requested):

PR #398, opened 2025-04-25
auto/bms: allow occupancy to be changed on a schedule

mattnathan:
The schedule works either independently or with the occupancy sensors. Without sensors the occupancy mode will be set to on when within the schedule, and off otherwise. When sensors exist, occupied mode is only ever set to on within the schedule using the normal occupancy sensor logic.

I've added tests for all the occupancy cases, both existing and schedule based, just to make sure I didn't introduce any regressions here.

### DIFF
--- a/pkg/auto/bms/config/root.go
+++ b/pkg/auto/bms/config/root.go
@@ -64,8 +64,10 @@ type Root struct {
 	//
 	// When using sensor based occupancy, if any sensor reports occupied then the mode targets are set to on.
 	// If no sensors report occupied then the UnoccupiedDelay is used to determine when to set the mode targets to off.
-	// The mode targets are only ever set to occupied when within the OccupiedSchedule.
-	// If no sensors are configured then the mode targets are set to on when within the OccupiedSchedule, and off otherwise.
+	//
+	// When using schedule based occupancy, the mode targets are set to on when within the OccupiedSchedule, and off otherwise.
+	//
+	// If both sensor and schedule based occupancy are used, the mode targets are set to on if both are true.
 	OccupancyModeTargets []SwitchMode        `json:"occupancyModeTargets,omitempty"` // Defaults: on=occupied, off=unoccupied
 	OccupancySensors     []string            `json:"occupancySensors,omitempty"`     // Sensors whose occupancy is linked with OccupancyModeTargets On mode.
 	UnoccupiedDelay      *jsontypes.Duration `json:"unoccupiedDelay,omitempty"`      // Defaults to 15m.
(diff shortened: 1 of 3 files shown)


pkg/auto/bms/logic.go
```diff
@@ -126,14 +159,14 @@ func analyseOccupancy(now time.Time, state *ReadState) (occupied, total, noRespo
 }
 
 // analyseTimeOfDay works out if we are currently within the deadband period, and when that might change.
```
hexaglow, line 161: Not just used for deadband processing any more, update comment to reflect more generic use

pkg/auto/bms/config/root.go
```diff
@@ -59,9 +59,17 @@ type Root struct {
 	// If this device supports AirTemperatureHistory then it will be used to seed the weighted average.
 	AutoModeOATemp string `json:"oaTemp,omitempty"`
 
-	OccupancySensors     []string            `json:"occupancySensors,omitempty"`     // Sensors whose occupancy is linked with OccupancyModeTargets On mode.
+	// Settings to control how and when we adjust the OccupancyModeTargets between occupied=on and unoccupied=off.
```
hexaglow, line 62: I'm not clear on some details of how occupancy sensors interact. Can you answer these questions, and document them?
- I can see a test that uses occupancy sensors and schedules. Is it intended that they be used together?
- What happens if a schedule says occupied but sensors say unoccupied, or vice versa?
- What happens if all occupancy sensors have no data, and there is a schedule?

pkg/auto/bms/logic.go
```diff
@@ -126,14 +159,14 @@ func analyseOccupancy(now time.Time, state *ReadState) (occupied, total, noRespo
...
-func analyseTimeOfDay(now time.Time, state *ReadState) (on bool, onStart, onEnd time.Time, changesIn time.Duration) {
+func analyseTimeOfDay(now time.Time, schedule []config.Range) (on bool, onStart, onEnd time.Time, changesIn time.Duration) {
```
hexaglow, line 162: I got quite confused looking at this function's `onStart` return values, which returns when the currently active period will *next* start (e.g. tomorrow), instead of when the currently active period started in the past. i.e. `onStart` is after `onEnd`. This doesn't matter when the values are printed with time-of-day only (which they are). Still, this detail was not intuitive to me and I think it's worth a comment.

pkg/auto/bms/config/root.go
hexaglow: After reading through the comment again and your PR description I think I understand now - the logic is by intersection i.e. if there's a schedule the sensor AND the schedule must report occupancy for occupied mode to be selected. Right?

mattnathan: That's correct

hexaglow (approved):

PR #400, opened 2025-05-02
auto/resetbrightness: add an auto to reset brightness after a delay

mattnathan:
See the package docs for the auto for details.

This auto can be used to turn the lights off in a room a bit after someone has turned them on.

I have tested this using the vanti-ugs example and different configuration options.


deanredfern26 (approved):

pkg/auto/resetbrightness/auto.go
```diff
@@ -0,0 +1,240 @@
...
+					return err
+				}
+			}
+		}
+	})
+
+	// reset the brightness of the devices
+	grp.Go(func() error {
+		timer := time.NewTimer(math.MaxInt64)
+		timer.Stop()
```
dloveen-v, line 119: I am sure there is a `defer` missing here? Unless I misunderstand the use of `timer` and `timer.Stop()`

PR #401, opened 2025-05-07
system/history: fix support for order_by clause

mattnathan:
This looks to have been missed when the feature was added and never tested in the context of a cohort.

I've fixed it by publishing the historypb page functionality in a way it can be used by both trait history APIs and the history admin api. The history system now uses this shared pagination logic along with the proto page token types.

As the pagination logic is already tested in the historypb package, I haven't added unit tests in the system/history package. Instead, I've manually tested the HistoryAdminApi using the vanti-ugs example by modifying the sysconf to enable the history system and running manual grpc commands against it.

Sorry for all the files changed, most of it is because of the publication of the PageReader type and methods.

Fixes SC-1074


hexaglow (changes requested):
Implementation looks good.

Although the pagination logic itself has moved into another package, I would appreciate a test in `system/history` to check that the translation of the `order_by` field is happening correctly, as this is custom logic. A handful of test cases of `storeServer.ListHistoryRecords` should be sufficient for this.
This would help catch any regressions.

mattnathan:
Unit test added

hexaglow (approved):

PR #402, opened 2025-05-15
auto/exporthttp: use order by and page size = 1 to reduce the amount of pagination in ListMeterHistory calls to calculate consumption in meters

dloveen-v:
This makes use of this fix https://github.com/vanti-dev/sc-bos/commit/bb723ae88593cc508f1886597484fc7fc979a8a0 which allows clients to calculate consumption between 2 timestamps without paginating through _**every meter history result between those 2 timestamps**_.

I have also added a fix because I don't trust

```
for {
  select {
    case <-time.After(...):
    ...
  }
}
```

won't execute case statements more than once when receiving from a chan dictated by a cron schedule.

### DIFF
--- a/pkg/auto/exporthttp/job/meter.go
+++ b/pkg/auto/exporthttp/job/meter.go
@@ -42,7 +42,7 @@ func getRecordsByTime(ctx context.Context, logger *zap.Logger, historyFn listMet
 	resp, err = historyFn(ctx, &gen.ListMeterReadingHistoryRequest{
 		Name: meter,
 		Period: &sctime.Period{
-			StartTime: timestamppb.New(now),
+			EndTime: timestamppb.New(now),
 		},
 		PageSize: 1,
 		OrderBy:  "record_time DESC",
--- a/pkg/auto/exporthttp/job/meter.go
+++ b/pkg/auto/exporthttp/job/meter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"time"
 
+	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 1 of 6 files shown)


pkg/auto/exporthttp/job/meter.go
```diff
@@ -14,46 +15,50 @@ import (
...
+		logger.Error("no records found in earliest", zap.String("meter", meter), zap.Time("start", start))
+		return earliest, latest, nil
+	}
+
+	earliest = resp.GetMeterReadingRecords()[0]
+
+	resp, err = historyFn(ctx, &gen.ListMeterReadingHistoryRequest{
+		Name: meter,
+		Period: &sctime.Period{
+			StartTime: timestamppb.New(now),
```
deanredfern26, line 45: Shouldn't this be EndTime? Else you are looking for records in the future

deanredfern26: Does this call not resolve to an SQL query something like this?
SELECT *
FROM history
WHERE source==meter AND start_time > now

I might be wrong but if so that won't get any records because the start time would have to be in the future

dloveen-v:
@deanredfern26 the reason I hadn't used `endTime` is because using this field parameter in a site deployment was giving very stale meter readings (from 2024). I have resolved the issue in this particular deployment and can merge this now.

deanredfern26:
> @deanredfern26 the reason I hadn't used `endTime` is because using this field parameter in a site deployment was giving very stale meter readings (from 2024). I have resolved the issue in this particular deployment and can merge this now.

Yeah the end time might not have worked there either, the point was more about not trying to look for records with a start time after `time.now`

deanredfern26:
@dloveen-v LGTM, have you managed to test this? Happy to approve if it has fixed/improved the problem

deanredfern26 (approved):
Based on the screenshot looks like this has been posting more sensible values than previously

PR #405, opened 2025-05-21
driver/mock: add simple automation for air temperature

deanredfern26:
simple automation which randomises a set point and regularly updates the ambient to be +- 2 degrees around the set point

### DIFF
--- a/pkg/driver/mock/auto/airtemperature.go
+++ b/pkg/driver/mock/auto/airtemperature.go
@@ -1,0 +1,51 @@
+package auto
+
+import (
+	"context"
+	"math"
+	"time"
+
+	"golang.org/x/exp/rand"
+
+	"github.com/smart-core-os/sc-api/go/traits"
+	"github.com/smart-core-os/sc-api/go/types"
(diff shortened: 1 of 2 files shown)


pkg/driver/mock/auto/airtemperature.go
```diff
@@ -0,0 +1,51 @@
...
+			}
+			_, _ = model.UpdateAirTemperature(state)
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+					state.AmbientTemperature = &types.Temperature{
+						// update the ambient to be +- 2 degrees from the set point
+						ValueCelsius: setPoint + (rand.Float64()*4 - 2),
```
mattnathan, line 39: While we're here, do you think you could use the live setpoint value instead of the initial set point for this randomness? I think that'd be more useful because you'd be able to change the set point via the Ops UI and the ambient temp would continue to wobble around this value. It would also mean automations and things don't get overridden.

mattnathan (changes requested):

mattnathan (approved):

PR #407, opened 2025-05-22
auto/history: refactor to allow pollOnSchedule lifecycle defined by a cron schedule in config

dloveen-v:
The default behaviour will be as before: without a cron schedule for the history automation, the pull.Fetcher will determine whether polling or pull operations are used to fetch changes from a device trait.

However, if a cron schedule is provided, the history automation will use a poll cycle to fetch values on a given schedule, and will ignore duplicate consecutive changes to device traits.

This is effort taken to reduce the storage required for a history automation.

### DIFF
--- a/pkg/auto/history/airqualitysensor.go
+++ b/pkg/auto/history/airqualitysensor.go
@@ -2,14 +2,13 @@ package history
 
 import (
 	"context"
-	"errors"
 
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/smart-core-os/sc-api/go/traits"
+	"github.com/smart-core-os/sc-golang/pkg/cmp"
 	"github.com/vanti-dev/sc-bos/pkg/auto/history/config"
-	"github.com/vanti-dev/sc-bos/pkg/util/pull"
 )
 
 func (a *automation) collectAirQualityChanges(ctx context.Context, source config.Source, payloads chan<- []byte) {
@@ -18,6 +17,8 @@ func (a *automation) collectAirQualityChanges(ctx context.Context, source config
 		a.logger.Error("collection aborted", zap.Error(err))
 		return
 	}
+
+	last := newDeduper[*traits.AirQuality](cmp.Equal(cmp.FloatValueApprox(0, 0.0001)))
 
 	pullFn := func(ctx context.Context, changes chan<- []byte) error {
 		stream, err := client.PullAirQuality(ctx, &traits.PullAirQualityRequest{Name: source.Name, UpdatesOnly: true, ReadMask: source.ReadMask.PB()})
@@ -30,10 +31,15 @@ func (a *automation) collectAirQualityChanges(ctx context.Context, source config
 				return err
 			}
 			for _, change := range msg.Changes {
-				payload, err := proto.Marshal(change.AirQuality)
+				if !last.Changed(change.GetAirQuality()) {
+					continue
+				}
+
+				payload, err := proto.Marshal(change.GetAirQuality())
 				if err != nil {
 					return err
 				}
+
 				select {
 				case <-ctx.Done():
 					return ctx.Err()
@@ -43,14 +49,20 @@ func (a *automation) collectAirQualityChanges(ctx context.Context, source config
 		}
 	}
 	pollFn := func(ctx context.Context, changes chan<- []byte) error {
-		demand, err := client.GetAirQuality(ctx, &traits.GetAirQualityRequest{Name: source.Name, ReadMask: source.ReadMask.PB()})
+		resp, err := client.GetAirQuality(ctx, &traits.GetAirQualityRequest{Name: source.Name, ReadMask: source.ReadMask.PB()})
 		if err != nil {
 			return err
 		}
-		payload, err := proto.Marshal(demand)
+
+		if !last.Changed(resp) {
+			return nil
+		}
+
+		payload, err := proto.Marshal(resp)
 		if err != nil {
 			return err
 		}
+
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
@@ -59,11 +71,7 @@ func (a *automation) collectAirQualityChanges(ctx context.Context, source config
 		return nil
 	}
 
-	err := pull.Changes(ctx, pull.NewFetcher(pullFn, pollFn), payloads, pull.WithLogger(a.logger))
-	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-		return
-	}
-	if err != nil {
+	if err := collectChanges(ctx, source, pullFn, pollFn, payloads, a.logger); err != nil {
 		a.logger.Warn("collection aborted", zap.Error(err))
 	}
 }
--- a/pkg/auto/history/collect.go
+++ b/pkg/auto/history/collect.go
@@ -1,0 +1,44 @@
+package history
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/vanti-dev/sc-bos/pkg/auto/history/config"
+	"github.com/vanti-dev/sc-bos/pkg/util/pull"
+)
+
+type collector func(ctx context.Context, source config.Source, payloads chan<- []byte)
+type getter func(context.Context, chan<- []byte) error
+
+func collectChanges(ctx context.Context, cfg config.Source, pullFn, pollFn getter, changes chan<- []byte, logger *zap.Logger) error {
+	if cfg.PollingSchedule != nil {
+		var t time.Time
+
+		for {
+			t = time.Now()
+			select {
+			case <-ctx.Done():
+				return nil
+			case <-time.After(time.Until(cfg.PollingSchedule.Schedule.Next(t))):
+				err := pollFn(ctx, changes)
+				if err != nil {
+					logger.Warn("poll aborted", zap.Error(err))
+				}
+			}
+		}
+	}
+
+	err := pull.Changes(ctx, pull.NewFetcher[[]byte](pullFn, pollFn), changes, pull.WithLogger(logger))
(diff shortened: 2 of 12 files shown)


pkg/auto/history/airqualitysensor.go
```diff
@@ -67,3 +68,48 @@ func (a *automation) collectAirQualityChanges(ctx context.Context, source config
...
+			if err != nil {
+				a.logger.Warn("sample aborted", zap.Error(err))
+				continue
+			}
+
+			select {
+			case <-ctx.Done():
+				return
+			case payloads <- payload:
+			}
```
mattnathan, line 112: Almost all this code is a duplicate, with the pattern being repeated between all the traits can you get rid of the duplicated code please.

pkg/auto/history/airqualitysensor.go
```diff
@@ -67,3 +68,48 @@ func (a *automation) collectAirQualityChanges(ctx context.Context, source config
...
+
+			t = time.Now()
+			if err != nil {
+				a.logger.Warn("sample aborted", zap.Error(err))
+				continue
+			}
+
+			if proto.Equal(prev, resp) {
+				continue
+			}
```
mattnathan, line 98: sc-golang has a `cmp` package that would be a better fit here. Would allow you to ignore timestamps and do float proximity comparisons.

pkg/auto/history/airqualitysensor.go
```diff
@@ -67,3 +68,48 @@ func (a *automation) collectAirQualityChanges(ctx context.Context, source config
...
+	var prev *traits.AirQuality
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(time.Until(source.Sample.Schedule.Next(t))):
+			resp, err := client.GetAirQuality(ctx, &traits.GetAirQualityRequest{Name: source.Name, ReadMask: source.ReadMask.PB()})
+
+			t = time.Now()
```
mattnathan, line 90: I think this would be better just after the `for {` statement. Would give you the same back pressure support without breaking the flow of the request, check error lines. It'd also handle back pressure from the chan send too.

pkg/auto/history/config/root.go
```diff
@@ -23,6 +24,13 @@ type Source struct {
 	Trait trait.Name `json:"trait,omitempty"`
 	// ReadMask instructs the history service to only read the specified fields
 	ReadMask *FieldMask `json:"readMask,omitempty"`
+	// Sample controls whether periodic downsampling is applied to the data source.
+	// It is intended for high-frequency sources where storing every data point is excessive.
+	// When Sample is set, the source is polled using a Cron schedule, and only values
+	// that differ from the previous sample are recorded — reducing storage by skipping redundant data.
+	// If Sample is not set (the default) or disabled, the source will be treated as event-driven:
+	// all changes will be recorded at their native frequency, with no sampling applied.
+	Sample *Sampling `json:"sample,omitempty"`
```
mattnathan, line 33: Can we not use the term Sampling here. Sampling usually means to selectively choose records from a stream of them. That's not what you're doing here. It's worse in this case, too, because the normal behaviour _is_ a stream.

pkg/auto/history/config/root.go
```diff
@@ -46,6 +54,14 @@ func (f *FieldMask) MarshalJSON() ([]byte, error) {
 	return protojson.Marshal((*fieldmaskpb.FieldMask)(f))
 }
 
+type Sampling struct {
+	// Schedule is the time schedule between samples being recorded.
+	// The default is every 1 hour at minute 0 < 0 * * * * >.
+	Schedule *jsontypes.Schedule `json:"schedule,omitempty"`
+	// Disabled is the flag used to disable sampling.
+	Disabled bool `json:"disabled,omitempty"`
+}
```
mattnathan, line 63: This is a fairly big breaking change for all deployed sc-bos systems. You're changing the default behaviour from Pull to Schedule(1h) which could have large effects on a lot of different apps. The Ops UI, for one example, relies on occupancy being more frequent than 20mins, which this would break. I don't see a reason to change the default here.

If the schedule is no longer the default then you can get rid of the Disabled property, and the Sampling struct entirely. Enabling a schedule based collection can be turned on with the presence of the schedule in the json config. ParseConfig can also go away.

pkg/auto/history/factory.go
```diff
@@ -157,26 +158,56 @@ func (a *automation) applyConfig(ctx context.Context, cfg config.Root) error {
...
 	switch cfg.Source.Trait {
 	case trait.AirQualitySensor:
 		serverClient = gen.WrapAirQualitySensorHistory(historypb.NewAirQualitySensorServer(store))
-		collect = a.collectAirQualityChanges
+
+		if cfg.Source.Sample != nil && !cfg.Source.Sample.Disabled {
+			collect = a.sampleAirQualityChanges
+		} else {
+			collect = a.collectAirQualityChanges
+		}
```
mattnathan, line 170: I don't think this logic belongs here. The collect function is already given the config.Source so can make up its own mind about how to collect things. Maybe as part of the de-dupe change this would become clearer. I'm thinking a package local util func like

```go
func collectChanges(ctx context.Context, cfg config.Source, pullFn, pollFn pull.GetFunc, changes chan-> []byte, logger *zap.Logger) error {
  // if cfg is using schedule
  //    setup poll loop calling pollFn
  // else
  //    call poll.Changes
}
```

mattnathan (changes requested):

pkg/auto/history/airqualitysensor.go
mattnathan: Also see the comment below about the code in factory.go

pkg/auto/history/factory.go
mattnathan: Doesn't have to be nil, could be a function that returns `status.Error(codes.Unsupported, "poll only")`, or something like that. Though I don't think this is _really_ what we want as we'd be using the pull.Changes poll behaviour instead of being in control of it here.

pkg/auto/history/airqualitysensor.go
dloveen-v: I was trying to avoid iterating through protobuf fields as these can be added / removed later on which has high maintenance requirements.

pkg/auto/history/airqualitysensor.go
dloveen-v: There's no guarantee `*jsontypes.Schedule.Next()` only executes once during a single scheduling cycle. I think this is an issue with the underlying library so setting `t = time.Now()` at the start of the for loop isn't ideal.

mattnathan (changes requested):
I get the impression that you're overcomplicating your design with the aim of putting the de-dupe logic into a single codepath. I'm not sure thats necessary or even desirable.

I'd move dedupe code into the `collectFooChanges` methods, shouldn't be too hard, you could even make a util for it:

```go
type deduper[T proto.Message] struct {
  last T
  equal cmp.Message // from github.com/smart-core-os/sc-golang/pkg/cmp
}

func (dd *deduper[T]) Changed(m T) bool {
  if dd.equal(dd.last, m) {
    return false
  }
  dd.last = m
  return true
}
```

Then in (for example) `collectAirQualityChanges` you'd have:

```go
last := deduper[*traits.AirQuality]{equal: cmp.Equal(
  cmp.FloatValueApprox(0, 0.01), // == if within 0.01 of each other
)}
pullFn := func(/* ... */) {
  // ...
  for _, change := msg.Changes {
    next := change.GetAirQuality()
    if !last.Changed(next) {
      continue;
    }
    // proto.Marshal, payloads<-etc
  }
}
pollFn := func(/* ... */) {
  next, err := client.GetAirQuality()
  if !last.Changed(next) {
    return
  }
  // proto.Marshal, payloads<-etc
}
```

I'm sure you can fill in the rest :D

pkg/history/memstore/store.go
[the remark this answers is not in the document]
mattnathan: Changing it in GH I meant
image

pkg/auto/history/collect.go
```diff
@@ -0,0 +1,66 @@
...
+				case <-ctx.Done():
+					return nil
+				case payload := <-payloads:
+					if bytes.Equal(payload, prev) {
+						continue
+					}
+
+					prev = payload
+					changes <- payload
+				}
```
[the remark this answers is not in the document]
mattnathan: You are forgetting that `pollFn` is blocking. By the time it returns it has either sent on `payloads` or never will.

pkg/auto/history/collect.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/vanti-dev/sc-bos/pkg/auto/history/config"
+	"github.com/vanti-dev/sc-bos/pkg/util/pull"
+)
+
+type collector func(ctx context.Context, source config.Source, payloads chan<- []byte)
+type getter func(context.Context, chan<- []byte) error
+
+func collectChanges(ctx context.Context, cfg config.Source, pullFn, pollFn getter, changes chan<- []byte, logger *zap.Logger) error {
+	defer close(changes)
```
mattnathan, line 18: This is surprising, closing channels is usually done right next to the thing that creates them. Having this here will mean we'll get lost in the future trying to find how the chan is closed.

Can you move the close back to the factory go routine.

I guess technically the close should be in a defer in the `collect` go routine as you close chans when no more data will be sent on them. Feels like an unrelated change though, up to you if you want to fix that

mattnathan (changes requested):
Looks much better, not too complicated, and does what it needs to do.

pkg/auto/history/collect.go
dloveen-v: I thought the sender closes the channel, if it is not created by the sender? This is to ensure the part of the code that is sending down the channel doesn't send on a closed chan.

mattnathan: Yeah, the sender closes it. But normally you'd do the close as close to the top of the send stack as you can. So in this case it'd be next to the `collect` call in factory.go.

PR #408, opened 2025-05-23
feat/device monitor auto

deanredfern26:
adds an automation which allows checking air temperature devices for; abnormal temperatures and/or whether the set point is reached in a certain time period. Raises a status alert when not ok conditions are met.

Was written to implement [SC-1046](https://vanti.youtrack.cloud/issue/SC-1046/Detect-abnormal-FCU-temperatures-set-point-vs-actual-discrepancies) but can be extended with other traits to abitarily monitor devices to detect abnormalities

There are some unit tests and you can also add the automation to the ugs example and adjust the FCU temperature to trigger the alerts

### DIFF
--- a/pkg/auto/allautos/factory.go
+++ b/pkg/auto/allautos/factory.go
@@ -4,6 +4,7 @@ import (
 	"github.com/vanti-dev/sc-bos/pkg/auto"
 	"github.com/vanti-dev/sc-bos/pkg/auto/azureiot"
 	"github.com/vanti-dev/sc-bos/pkg/auto/bms"
+	"github.com/vanti-dev/sc-bos/pkg/auto/devicemonitor"
 	"github.com/vanti-dev/sc-bos/pkg/auto/export"
 	"github.com/vanti-dev/sc-bos/pkg/auto/exporthttp"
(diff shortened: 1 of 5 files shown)


mattnathan:
I'm not entirely sure about the approach being used here. If I'm reading correctly each monitored device could result in multiple alerts being created.

Aside from the orphaned alerts that would happen when the auto is restarted while a device is misbehaving, I think the statusalerts auto will be stepping on the toes of this auto as that one assumes a single alert per source.

Having the auto be alert generating instead of status updating might also cause some confusion down the line when you see an alert for a device but the device details say it's all ok.

Thoughts?

deanredfern26:
> I'm not entirely sure about the approach being used here. If I'm reading correctly each monitored device could result in multiple alerts being created.
> 
> Aside from the orphaned alerts that would happen when the auto is restarted while a device is misbehaving, I think the statusalerts auto will be stepping on the toes of this auto as that one assumes a single alert per source.
> 
> Having the auto be alert generating instead of status updating might also cause some confusion down the line when you see an alert for a device but the device details say it's all ok.
> 
> Thoughts?

So instead of this auto creating alerts for devices, instead it sets the status and then the statusalerts auto is the one which then creates the alerts?

PR #410, opened 2025-05-23
driver/mock: add time of day scaling for random values

mattnathan:
Now electric, meter, and occupancy values will change based on the working day.

Also includes more realistic meter calculations based on kWh instead of random increments.

### DIFF
--- a/pkg/driver/mock/auto/Meter.go
+++ b/pkg/driver/mock/auto/Meter.go
@@ -4,9 +4,9 @@ import (
 	"context"
 	"time"
 
-	"golang.org/x/exp/rand"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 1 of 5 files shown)


deanredfern26 (commented):
LGTM. 
Separate point not related to the code, is it worth fixing the casing of Meter & OccupancySensor filenames as part of this while your here? Keeps bugging me

mattnathan:
Can I change the names separately, because of annoying case insensitivity issues?

deanredfern26 (approved):

mattnathan:
> One for the future.. It would be good if we could somehow use this auto for mock data in the DB seeding tools, rather than duplicating the new scaling logic. This is out of scope for this PR though.

Yes, that occurred to me. The scale package was originally in utils but figured we can move it later if we want to, or make a better mock data package that's less tied to the driver impl... something like that. Future PR stuff.

PR #411, opened 2025-05-23
drivers/opcua: add driver for opcua

deanredfern26:
adds a driver for OPC UA devices, essentially a wrapper around gopcua https://github.com/gopcua/opcua which allows implementors to write config to add support for devices & traits.

being used at OCW and now 3CS, need to get this merged as we will use this at 19 Cornwall and 40LH too


go.mod
```diff
@@ -32,14 +33,15 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/timshannon/bolthold v0.0.0-20210913165410-232392fc8a6a
 	github.com/vanti-dev/gobacnet v0.0.0-20250122134204-5b55cc2d13fe
+	github.com/vanti-dev/mepc-3cs v0.0.0-20250523093338-dd77bdcf0150
```
dloveen-v, line 36: Don't think we want this

dloveen-v: It doesn't look like it's an import causing this. The `go.mod` just needs a re-do

pkg/driver/opcua/driver.go
```diff
@@ -49,10 +50,9 @@ type Driver struct {
 
 func (d *Driver) applyConfig(ctx context.Context, cfg config.Root) error {
 
-	config.SetDefaults(&cfg)
-	a, undo := node.AnnounceScope(d.announcer)
+	a := node.NewReplaceAnnouncer(d.announcer).Replace(ctx)
```
mattnathan, line 53: This line needs to be in the `Factory.New` method and a saved field stored in `*Driver`

PR #414, opened 2025-06-03
all: replace node.Clienter usage with node.ClientConner

mattnathan:
The node.Node.Client method was already deprecated, but not the node.Clienter interface nor node.Clienter.Client method so there were still plenty of uses throughout the code. The former is brittle and prone to runtime issues, like I just found with the client credential auth not working in single node mode.

This change replaces all usage of the old mechanism with the new one while keeping the behaviour the same. In fact some code will have got a little more reliable as boot order is no longer important when resolving client instances.

Downstream projects should now see deprecation warnings if they use the old Clienter interface, we'll give them a bit of time to update before we delete node.Clienter and all the associated types, funcs, and packages (like alltraits).

### DIFF
--- a/internal/driver/settings/driver.go
+++ b/internal/driver/settings/driver.go
@@ -24,7 +24,6 @@ type factory struct{}
 func (f factory) New(services driver.Services) service.Lifecycle {
 	d := &Driver{
 		services:  services,
-		clients:   services.Node,
 		announcer: node.NewReplaceAnnouncer(services.Node),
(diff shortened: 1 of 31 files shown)


hexaglow (approved):
A straightforward change with no downsides that I can see

dloveen-v (approved):
LGTM. I have tested and the broken endpoint when running the vanti-ugs docker compose is fixed when ran against this PR

PR #418, opened 2025-06-05
demo/ugs: demo container move and optimisation

mattnathan:
This PR changes how the demo container images are built and deployed for a few reasons:

1. We tend to use podman, so use that instead of docker by default
2. The old images were ~1GB large (and there were two of them), the new ones are ~50-130MB; much better!
3. The old build process needed manual steps to copy config and build ui apps, the new one doesn't
4. The new build is tagged with a version, though I haven't quite worked out the best way to publish compose files for this yet
5. We are trying to move to GH Packages (away from nexus, aws, etc) so I've made that change too.

The crux of the change is to combine the Dockerfile from the root of the repo with the config from examples/config/vanti-ugs, which this does - though there is a bug in Podman that we've had to work around, see the Ugs dockerfile for details.

Hopefully this change will allow us to streamline the publication of demo images going forward, maybe even automate the process around releases.


mattnathan:
> Is the idea that we just run `build.sh` and that does everything for us? The README still looks like the old steps.
> 
> When I run `build.sh` I get a `panic: assignment to entry in nil map` at `STEP 9/16: RUN --mount=type=cache,target=/yarn-cache --mount=type=secret,id=npmrc,target=/root/.npmrc yarn install --frozen-lockfile --check-files`

I'm working on the readme now and testing the deployment steps. Still a draft PR ;)

deanredfern26:
I updated ubuntu for wsl & podman and got rid of the previous error, still have issue with this
`STEP 9/16: --mount=type=cache,target=/yarn-cache     --mount=type=secret,id=npmrc,target=/root/.npmrc     yarn install --frozen-lockfile --check-files`

I get a `401 Unauthorized` for this url ` https://npm.pkg.github.com/download/@smart-core-os/sc-api-grpc-web`, 
I did create a new token and authorize the organisations, `/root/.npmrc` looks fine
Likely just me not doing something right but can you just check the readme in case there is a step missing?

mattnathan:
@deanredfern26 I've updated it to not use the npmrc as we didn't need it anyway.

deanredfern26:
Yeah I still got this when I tried
` Error: https://npm.pkg.github.com/download/@smart-core-os/sc-api-grpc-web/1.0.0-beta.56/010c5da8c16e0906edb214728886fda9c1fc242a: Request failed "401 Unauthorized"`

Have done a hard reset, cleared yarn cache, yarn lock, node modules and rebuilt ops ui, still get the same. Likely just an issue with my machine, I jumped on something else I will take another look now

mattnathan:
Very strange, that package is public so shouldn't be going through auth anyway.

hexaglow (approved):
Verified working

PR #421, opened 2025-06-06
Login to accounts in the local database

hexaglow:
Support logging in to user accounts stored in the local accounts database.

The user can use the local account login on the Ops UI to log in to a user account stored in the database. As the same API is used for login, no changes will be required to the frontend to log in.

The system still uses the old role system & token format, so only assigned roles that have a legacy role mapped will be effective during login. Migrating to the new scoped role assignments with permissions is a following piece of work.

Once the new account management UI is also merged, this will achieve self-managed user accounts from the Ops UI.

### DIFF
--- a/internal/auth/accesstoken/accesstoken_test.go
+++ b/internal/auth/accesstoken/accesstoken_test.go
@@ -8,18 +8,7 @@ import (
 )
 
 func TestTokenSource_createAndVerify(t *testing.T) {
-	key, err := generateKey()
-	if err != nil {
-		t.Fatal(err)
-	}
-	ts := &Source{
-		Key:    key,
-		Issuer: "test",
-		Now:    time.Now,
-	}
-
-	ts.SignatureAlgorithms = []string{string(jose.HS256)}
-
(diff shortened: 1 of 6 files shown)


hexaglow:
I've solved 1 by returning different OAuth2 error codes, based on the spec, to allow the client to distinguish these scenarios:
- When the username/password is incorrect: `invalid_grant`
- When the user has no roles assigned: `unauthorized_client`

> Looks good. Have tested locally and all works fine barring a few quirks:
> 
>     1. cred exchange for tokens when the account has no roles fails with 401 and an identical response to invalid creds. I think we need to distinguish between "login succeeded, you don't have permission" and "login failed"
> 
>     2. deleting an account that exists in users.json then rebooting the BOS has the account show up again - ticket created on backlog to prevent deleting this kind of account.

mattnathan (approved):

hexaglow:
Rebased onto main and squashed fixups.

hexaglow:
Test failure unrelated to changes in this PR (I've created a ticket). Merging anyway.

PR #422, opened 2025-06-10
fix/dont allow nil history payloads

deanredfern26:
prevent records with nil payloads being added to the history table

null payloads were causing issues manifesting in wierd graphs: [SC-1093](https://vanti.youtrack.cloud/issue/SC-1093) 
I don't think it makes sense to record nil payloads in history, at least with the current implementation

### DIFF
--- a/pkg/auto/history/meter.go
+++ b/pkg/auto/history/meter.go
@@ -2,6 +2,7 @@ package history
 
 import (
 	"context"
+	"math"
 	"time"
 
 	"go.uber.org/zap"
@@ -28,6 +29,9 @@ func (a *automation) collectMeterReadingChanges(ctx context.Context, source conf
 				return err
 			}
 			for _, change := range msg.Changes {
+				if change.GetMeterReading() == nil || math.IsNaN(float64(change.GetMeterReading().Usage)) {
+					continue
+				}
 				if !last.Changed(change.GetMeterReading()) {
 					continue
 				}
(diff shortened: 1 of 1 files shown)


deanredfern26:
@mattnathan Is there a reason the history table schema allows NULL for the payload? 
I can change this fix if there is a reason to allow NULL just couldn't see why we would

pkg/history/apistore/store.go
```diff
@@ -31,6 +32,9 @@ type Store struct {
 var _ history.Store = (*Store)(nil)
 
 func (s *Store) Append(ctx context.Context, payload []byte) (history.Record, error) {
+	if payload == nil {
+		return history.Record{}, fmt.Errorf("payload cannot be nil")
+	}
```
mattnathan, line 37: I think this is solving the issue in the wrong place.
Firstly, checking `slice == nil` is bad practice, `len(slice) == 0` is better.
Secondly, an empty slice is a valid payload, it's a serialised proto that has no non-default, non-empty fields.

If we must check a slice for nil then we must comment why we're doing this and why nil is treated differently than an empty slice.

mattnathan (changes requested):
I agree that we shouldn't be storing nil payload data in the database, however I'm not sure what you've got here is solving the correct problem. In Go a nil slice should be treated identically to an empty slice, which admittedly seems to be different to how pgx is handling it.

Somewhere in the code we are attempting to store an empty / nil payload when we think it shouldn't be empty. If this is happening I suspect we have a bug in one of our history subscriptions. Maybe we need to add `UpdatesOnly` or be a bit smarter about which records we store.

What do you think?

PR #423, opened 2025-06-11
all: use range-over-int where we can

mattnathan:
Places found with: `for (\w+) := 0; \1 < ([\w\d]+); \1\+\+ \{`


hexaglow (approved):

dloveen-v:
```
for (\w+) :*= [\S]+; \1 <=* ([\w\d]+); \1\+\+ \{
```

This is more accurate catching : https://github.com/vanti-dev/sc-bos/blob/main/cmd/tools/dbadd-publications-sample/main.go#L104

mattnathan:
> ```
> for (\w+) :*= [\S]+; \1 <=* ([\w\d]+); \1\+\+ \{
> ```
> 
> This is more accurate catching : https://github.com/vanti-dev/sc-bos/blob/main/cmd/tools/dbadd-publications-sample/main.go#L104

@dloveen-v that case can't be replaced by range-over-int without also changing the body of the loop.

PR #424, opened 2025-06-11
auto/lights: hope to make the test more reliable

mattnathan:
This includes a bunch of changes that might help us diagnose flaky tests:

1. Increase the timeout for state changes by an order of magnitude
2. Replace use of emitter with a blocking call to reduce async-iness
3. Hold a lock when signalling the next call should return an error
4. Turn on logging for when the above inevitably fails to fix the issue

### DIFF
--- a/pkg/auto/lights/auto_test.go
+++ b/pkg/auto/lights/auto_test.go
@@ -3,10 +3,14 @@ package lights
 import (
 	"context"
 	"errors"
+	"fmt"
+	"math"
+	"sync"
 	"testing"
 	"time"
 
 	"go.uber.org/zap"
+	"golang.org/x/exp/rand"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/smart-core-os/sc-api/go/traits"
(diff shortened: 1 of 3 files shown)


hexaglow (changes requested):
Still failing intermittently unfortunately. No failure under the race detector. I think it's a sequencing problem rather than a violation of the concurrency rules.

Logs attached.
[lights_test.txt](https://github.com/user-attachments/files/20689770/lights_test.txt)

pkg/auto/lights/auto_test.go
```diff
@@ -30,59 +33,69 @@ func TestPirsTurnLightsOn(t *testing.T) {
...
 	testActions := newTestActions(t)
 
-	now := time.Unix(0, 0)
+	clock := newTestClock(t)
 
-	automation := PirsTurnLightsOn(rootNode, zap.NewNop())
+	loggerConfig := zap.NewDevelopmentConfig()
+	loggerConfig.DisableStacktrace = true
+	logger, _ := loggerConfig.Build()                                          // the test is sometimes failing, this might help with debugging
+	logger = logger.Named(fmt.Sprintf("%x", math.Float64bits(rand.Float64()))) // helps with --count>1 tests
```
hexaglow, line 41: What's the motivation for this random-float-to-int conversion?
Seems like it just gets a random integer with a certain limited range?
Also, `x/exp/rand` is used here which is deprecated.
Couldn't [rand.Uint64](https://pkg.go.dev/math/rand/v2#Uint64) be used instead?

hexaglow (changes requested):
I've tested this and I can no longer trigger the intermittent failures at all, so looks to have fixed the issue.

pkg/auto/lights/auto_test.go
mattnathan: Good catch, and you are correct. I went down the road of thinking _I want 32 or 64 bits of random data_ and got caught in the wrong are of code.

hexaglow (approved):

PR #425, opened 2025-06-11
ui/ops: don't hang if there are too many graphic elements

mattnathan:
We don't have a good solution for this, but hanging is definitely a bad solution. At least with this option other parts of the app, like sc status and the sidebar, continue to work.


deanredfern26 (approved):
This is much better

PR #426, opened 2025-06-12
ui/ops: format airQualityCard values to 1 decimal point

dloveen-v:
Small change to fix a numerical text overflow problem:

Before:
Screenshot 2025-06-13 at 11 47 02

After:
Screenshot 2025-06-13 at 11 46 12

### DIFF
--- a/ui/ops/src/traits/airQuality/AirQualityCard.vue
+++ b/ui/ops/src/traits/airQuality/AirQualityCard.vue
@@ -9,7 +9,7 @@
       <v-list-item v-for="(val, key) of presentMetrics" :key="key" class="py-1">
         <v-list-item-title class="text-body-small">{{ key }}</v-list-item-title>
         <template #append>
-          <v-list-item-subtitle class="font-weight-medium text-body-1">{{ val.value }}</v-list-item-subtitle>
+          <v-list-item-subtitle class="font-weight-medium text-body-1">{{ intlNumbers.format(val.value ?? 0) }}</v-list-item-subtitle>
         </template>
       </v-list-item>
     </v-list>
@@ -42,6 +42,8 @@ const props = defineProps({
     required: true
   }
 });
+
+const intlNumbers = new Intl.NumberFormat();
 
 const {score, presentMetrics} = useAirQuality(() => props.value);
 const scoreColor = useStatusColor(score);


ui/ops/src/traits/airQuality/AirQualityCard.vue
```diff
@@ -9,7 +9,7 @@
       <v-list-item v-for="(val, key) of presentMetrics" :key="key" class="py-1">
         <v-list-item-title class="text-body-small">{{ key }}</v-list-item-title>
         <template #append>
-          <v-list-item-subtitle class="font-weight-medium text-body-1">{{ val.value }}</v-list-item-subtitle>
+          <v-list-item-subtitle class="font-weight-medium text-body-1">{{ (val.value ?? 0).toFixed(3) }}</v-list-item-subtitle>
```
mattnathan, line 12: And now any metric that is in the `123` range will be shown to way more points than is needed wasting space. I think we have a util somewhere that formats reasonably, maybe this: https://github.com/vanti-dev/event-anticipate-2024/blob/9a6213bd27a3a5493313d11fa0affcad16f856ac/ui/dashboards/src/util/number.js#L1

Al alternative is using `Intl.NumberFormat` for this which has support for significant digits

mattnathan (changes requested):

ui/ops/src/traits/airQuality/AirQualityCard.vue
dloveen-v: I don't have access to that repository. I will look at `Intl`

dloveen-v:
Incidentally, I have found this quote:
>Every time toLocaleString is called, it has to perform a search in a big database of localization strings, which is potentially inefficient. When the method is called many times with the same arguments, it is better to create a Intl.DateTimeFormat  object and use its format()  method, because a DateTimeFormat object remembers the arguments passed to it and may decide to cache a slice of the database, so future format calls can search for localization strings within a more constrained context.

This is also true for `Intl.NumberFormat`.

I tested it and 1,000,000 `toLocaleString()` calls takes ~37s on my browser. With a `Formatter` singleton, it takes <1s. Probably a performance improvement we can make across the ops ui with a fairly simple change.

ui/ops/src/traits/airQuality/AirQualityCard.vue
dloveen-v: Incidentally, I have found this quote:

Every time toLocaleString is called, it has to perform a search in a big database of localization strings, which is potentially inefficient. When the method is called many times with the same arguments, it is better to create a Intl.DateTimeFormat object and use its format() method, because a DateTimeFormat object remembers the arguments passed to it and may decide to cache a slice of the database, so future format calls can search for localization strings within a more constrained context.

This is also true for Intl.NumberFormat.

I tested it and 1,000,000 toLocaleString() calls takes ~37s on my browser. With a Formatter singleton, it takes <1s. Probably a performance improvement we can make across the ops ui with a fairly simple change.

mattnathan (approved):

PR #432, opened 2025-06-17
ui/ops/accounts: add a users page

mattnathan:
This page is a copy of the accounts page but with all the service account and role scopes removed.
When we are ready to release the new auth system these pages should be deleted.

As the page will eventually be deleted, I felt copy-paste was the best option, though did parameterise some components where it made sense.


mattnathan:
> @mattnathan How hard would it be to use a feature flag to selectively disable the parts we don't want yet? I'd prefer this approach. Temporary things have a habit of becoming less temporary that you expect, after all

The approach taken is more or less a feature flag already. The users page _is_ the accounts page with the features we don't want taken out/adjusted, though I admit that a copy isn't quite the same as a parameterised component.

My rational behind this was that to make the accounts page configurable enough to support both modes would have made it way more complicated and harder to maintain. Where it made sense to do this (for the sub components) I did, but the top level page would have had to compose all those individual parameters. We could probably make it work, but the end would look very similar to what we have now but all muddled together. Just my opinion on it

PR #433, opened 2025-06-18
driver/bacnet: support light trait

dloveen-v:
Although some BacNet light points may support granular brightness intensity levels, I have opted to stick to scenes only for now as these are usually paired with interactions with the PIRs which breaks or requires additional logic involving timeouts and such when setting granular intensities.

### DIFF
--- a/pkg/driver/bacnet/merge/light.go
+++ b/pkg/driver/bacnet/merge/light.go
@@ -165,7 +165,7 @@ func (l *light) pollPeer(ctx context.Context) (*traits.Brightness, error) {
 			scene, err := l.findSceneByValue(int(value))
 
 			if err != nil {
-				return comm.ErrPropNotFound
+				return err
 			}
 
 			data.Preset.Name = scene.Name
--- a/pkg/driver/bacnet/merge/light.go
+++ b/pkg/driver/bacnet/merge/light.go
@@ -1,0 +1,211 @@
+package merge
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+	"google.golang.org/grpc/codes"
+	grpcStatus "google.golang.org/grpc/status"
+
+	"github.com/smart-core-os/sc-api/go/traits"
+	"github.com/smart-core-os/sc-golang/pkg/cmp"
+	"github.com/smart-core-os/sc-golang/pkg/resource"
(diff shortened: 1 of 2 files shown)


pkg/driver/bacnet/merge/light.go
```diff
@@ -0,0 +1,211 @@
...
+
+			scene, err := l.findSceneByValue(int(value))
+
+			if err != nil {
+				return comm.ErrPropNotFound
+			}
+
+			data.Preset.Name = scene.Name
+			data.Preset.Title = scene.Title
+			data.LevelPercent = scene.Brightness
```
deanredfern26, line 170: This `data.LevelPercent = scene.Brightness` should go before the error check on `l.findSceneByValue(int(value))`
We are getting a brightness level back but we just can't map it to a scene.

Also not sure `comm.ErrPropNotFound` is the correct error to return here, we do find the property just not the scene

deanredfern26 (changes requested):

pkg/driver/bacnet/merge/light.go
dloveen-v: I changed the error in a later commit. 

Regarding the first paragraph I don't think the BacNet point is reporting any brightness levels. It looks from the only documented implementation I have that the point is reporting a scene number which we have to convert to a brightness internally.

pkg/driver/bacnet/merge/light.go
```diff
@@ -0,0 +1,211 @@
...
+	presetName := request.GetBrightness().GetPreset().GetName()
+
+	scene, err := l.findSceneByName(presetName)
+
+	if err != nil {
+		return nil, grpcStatus.Error(codes.InvalidArgument, "preset not found")
+	}
+
+	if l.config.Point != nil {
+		err := comm.WriteProperty(ctx, l.client, l.known, *l.config.Point, scene.SetValue, 0)
```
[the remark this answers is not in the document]
dloveen-v: Will take a look thanks

pkg/driver/bacnet/merge/light.go
dloveen-v: I mean, technically you are right as the documentation says points values 0-100 are brightness values. But I have chosen to only use values 101-108 for now as in my experience, setting granular brightness values breaks the PIR automation in the background.

deanredfern26 (approved):

PR #439, opened 2025-06-24
ui/space: swap to infinite-scroll to render setup page; fix layout when airQuality gauge is missing; prevent nullPointer exceptions in airTemperature

dloveen-v:
The use of v-intersect was broken - if only a single zone was returned, the v-intersect would never leave the viewport causing the load more zone functionality to be permanently disabled until someone resized the viewport so the v-intersect leaves the viewport. The v-infinite-scroll component is a more reliable way to render zones and fetch more on scroll-to-bottom.

I have also fixed the airQuality component which would render the metric success/error status outside of the component if the airQuality gauge data was missing from the backend due to device limitations.

I also prevent null pointer exceptions when the devices which support airTemperature trait don't return a value celcius reading.


dloveen-v:
Before:
https://github.com/user-attachments/assets/1ab14e2d-2473-4aeb-899b-8e2c7143607e

After:

https://github.com/user-attachments/assets/7ef07f7c-59e7-4cc1-a8bd-75dfacf8c7b7

deanredfern26 (approved):
LGTM

PR #442, opened 2025-06-25
driver/helvarnet: driver for Helvarnet lighting systems

deanredfern26:
Driver for lighting systems which communicate via the Helvarnet lighting protocol
Tested this on site and lighting control is working with fast response
We are using this as Island too so promoting it now it has actually been visually tested


pkg/driver/helvarnet/client.go
```diff
@@ -0,0 +1,157 @@
...
+
+// make 3 attempts to try to send pkt and receive response. want is the prefix of the response we are expecting
+// if the response is not what we expect, we return an error
+func (c *tcpClient) sendAndReceive(pkt string, want string) (string, error) {
+
+	for tries := 3; tries > 0; tries-- {
+		var err error
+		err = c.sendPacket(pkt)
+		if err != nil {
+			c.logger.Error("failed to send packet", zap.Error(err))
```
dloveen-v, line 129: Same here, is it worth returning an `err`?

pkg/driver/helvarnet/client.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		c.close()
+		return "", err
+	}
+	return string(buf[:n]), nil
+}
+
+// make 3 attempts to try to send pkt and receive response. want is the prefix of the response we are expecting
+// if the response is not what we expect, we return an error
+func (c *tcpClient) sendAndReceive(pkt string, want string) (string, error) {
```
dloveen-v, line 123: I tend to lock sendAndReceive cycles with a thread-safe lock (like a mutex). The reason is if concurrent sendAndReceive calls are made, the send may race with the subsequent receive between calls. What's your opinion?

deanredfern26:
@dloveen-v no worries if you are busy with project stuff but if you have time to look at this hopefully I can get it done before planning tomorrow

deanredfern26:
[SC-1136](https://vanti.youtrack.cloud/issue/SC-1136) helvarnet driver: check context.Context in sendAndReceive()

PR #445, opened 2025-06-30
fix/ui-build-warnings

deanredfern26:
2 commits which fix 2 separate things:

- make imports of PlaceholderCard.vue dynamic to remove `yarn build` warnings
- fix casing of `AccountsSideBar` import which was breaking on certain systems

### DIFF
--- a/ui/ops/src/dynamic/layout/LayoutGrid.vue
+++ b/ui/ops/src/dynamic/layout/LayoutGrid.vue
@@ -11,8 +11,9 @@
 </template>
 
 <script setup>
+import {defineAsyncComponent} from 'vue';
 import useSignage from '@/composables/signage.js';
-import PlaceholderCard from '@/dynamic/widgets/general/PlaceholderCard.vue';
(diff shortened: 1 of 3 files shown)


ui/ops/src/routes/ops/overview/pageConfig.js
```diff
@@ -1,11 +1,11 @@
 import {builtinLayouts} from '@/dynamic/layout/pallet.js';
-import PlaceholderCard from '@/dynamic/widgets/general/PlaceholderCard.vue';
+const PlaceholderCard = defineAsyncComponent(() => import('@/dynamic/widgets/general/PlaceholderCard.vue'));
```
dloveen-v, line 2: Is this the right order? I think it would be best to move this below the import for `defineAsyncComponent`. I know `import` will evaluate first but it just doesn't feel as readable.

ui/ops/src/dynamic/layout/LayoutGrid.vue
```diff
@@ -11,8 +11,9 @@
 </template>
 
 <script setup>
+import {defineAsyncComponent} from 'vue';
```
dloveen-v, line 14: duplicate `from 'vue';`

PR #450, opened 2025-07-08
ui/ops: consistently format electrical and meter usage

mattnathan:
Fixes an issue where small values would show as `-0.00`, now they show as `~0`


deanredfern26 (commented):
Looks fine but just a thought is is worth added a little tooltip in the ElectricDemandCell for what the tilde means when we show it? Just think it might be a question that gets asked but we can wait to see if it does

PR #453, opened 2025-07-08
ui/ops/ops: support child-only dashboard pages

mattnathan:
For any page that is configured with children but no layout of itself, configure the nav to treat this as a non-navigable page. This means clicking on the nav item will not navigate, but will instead directly expand the child items.

Fixes SC-1087

image

The UGS example has been updated to group the energy and water pages like this.


deanredfern26 (approved):
Looking good at 3CS
image

PR #457, opened 2025-07-14
proto/traits: add report trait

deanredfern26:
Adds a Report trait, which is designed to allow listing and downloading of reports, such as monthly meter reports etc., via the OPs UI. The idea is to use this trait to implement [SC-1129](https://vanti.youtrack.cloud/issue/SC-1129).

There is minimal functionality here, the idea is we can implement this on the monthly meter report automation and add a simple page on the UI to list & download reports ready for next month, instead of emailing, to increase traffic to Smart Core.

### DIFF
--- a/proto/report.proto
+++ b/proto/report.proto
@@ -53,7 +53,7 @@ message Report {
   string description = 3;
 
   // The timestamp when the report was created.
-  google.protobuf.Timestamp created_at = 4;
+  google.protobuf.Timestamp create_time = 4;
 
   // The media type of the report, e.g., "application/pdf", "text/csv".
   string media_type = 5;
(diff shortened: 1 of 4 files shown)


proto/report.proto
```diff
@@ -0,0 +1,69 @@
...
+option go_package = "github.com/vanti-dev/sc-bos/pkg/gen";
+
+import "google/protobuf/field_mask.proto";
+import "google/protobuf/timestamp.proto";
+
+// ReportApi provides methods to manage and retrieve reports. Reports are typically created by automations, such as
+// monthly metering reports, monthly fault reports etc. This API allows automations to provide these reports on demand.
+service ReportApi {
+  rpc ListReports(ListReportsRequest) returns (ListReportsResponse);
+  rpc DownloadReport(DownloadReportRequest) returns (stream ReportChunk);
```
mattnathan, line 14: I'm not sure this download api will work in the browser, don't think streaming local file download is a thing. Other download apis we have use a GetDownloadURL rpc instead.

mattnathan: Ah, ok. We should attempt to avoid any API that forces the browser to download and process the entire file at once, especially for this kind of thing. That's what the URL versions were made to do.

PR #458, opened 2025-07-16
proto: regen proto files with version 5.29.3

deanredfern26:


dloveen-v (approved):
LGTM. I am however, slightly worried for the files with bigger changes in the bytes output there is some risk of loss of compatibility. Do we have a risk-free upgrade path for updating sc-bos in on-site deployments?

PR #464, opened 2025-07-25
util/slices: fix out of order sorted slice

mattnathan:
Add tests and fix an issue where inserted (or deleted, updated, etc) items would not be placed in the correct order in the set, in fact it was reversed, which would result in duplicate items in the set.


pkg/util/slices/sorted.go
```diff
@@ -124,3 +118,9 @@ func (s *Sorted[T]) Copy() *Sorted[T] {
 		cmp:   s.cmp,
 	}
 }
+
+func (s *Sorted[T]) find(item T) (int, bool) {
+	return sort.Find(s.Len(), func(i int) int {
```
hexaglow, line 123: Nitpick: `slices.BinarySearchFunc(s.items, item, s.cmp)` is a little neater, given that we're using generics.

hexaglow (approved):

PR #466, opened 2025-07-25
system/gateway: have the gateway support proxying legacy names (again)

mattnathan:
This was removed (incorrectly, imo) when we refactored to the new scan+announce model. We didn't pick up on it because it didn't actually break anything in the test, but did cause a lot of extra loops.

### DIFF
--- a/pkg/system/gateway/scan.go
+++ b/pkg/system/gateway/scan.go
@@ -130,6 +130,9 @@ func (s *System) pullSystems(ctx context.Context, node *remoteNode) (task.Next, 
 			if id == "" {
 				continue // not sure what happened, all services should have an id
 			}
+			if id == LegacyName {
+				id = Name // legacy support for the old system name
+			}
 			switch id {
 			case Name:
 				systems.gateway = c.GetNewValue()


mattnathan:
> * Would this only affect systems that are using mixed versions of sc-bos, announcing the `proxy` name?

Yes, specifically when we have (at least) one gateway and (at least) one proxy - this is how the e2e tests are set up right now.

> * What's the extra loops you refer to? Is it useless announcements similar to [systems/gateway: wait for gateway checks before announcing #467](https://github.com/vanti-dev/sc-bos/pull/467) ?

Because the gateway didn't recognise that another gateway was a gateway, it would try to announce names from it, which would fail because they'd already been announced by the owner node. This would cause the entire announcement to fail and retry, which would try to announce them again. The loop is supposed to be there to cover other issues we might face, but this PR fixes this known issue.

mattnathan:
> What's the extra loops you refer to?

If you run the e2e tests with logging enabled you see a lot of name announced ... name unannounced ... name announced, etc. This is the loop

hexaglow (approved):

PR #467, opened 2025-07-25
systems/gateway: wait for gateway checks before announcing

mattnathan:
See commit messages for details. TL;DR wait a little before announcing a remote node to give the scan a chance to get a response from the node about whether it's a gateway or not.

This is an optimisation, it should not change behaviour from before.

### DIFF
--- a/pkg/system/gateway/announce.go
+++ b/pkg/system/gateway/announce.go
@@ -5,6 +5,7 @@ import (
 	"iter"
 	"strings"
 	"sync"
+	"time"
 
 	"go.uber.org/zap"
 	reflectionpb "google.golang.org/grpc/reflection/grpc_reflection_v1"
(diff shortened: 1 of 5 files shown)


hexaglow:
@mattnathan could you provide a bit more info about what the problem is without the patch?
What happens and when does it occur?

hexaglow (approved):

mattnathan:
> @mattnathan could you provide a bit more info about what the problem is without the patch? What happens and when does it occur?

The comment here outlines the issue: https://github.com/vanti-dev/sc-bos/pull/467/files#diff-c91581da2b567540f8b8f158c19577773c643ac18cc94883a2a28bd478a4388aR172-R174
Similarly this commit message says the same thing: https://github.com/vanti-dev/sc-bos/pull/467/commits/92f32c30799e750b6394e94f6a02f516138d25c1

Basically, when gateway A proxies gateway B there's a period of time when A doesn't know B is a gateway yet and so tries to add all of B's devices to it's own routing table, which can emit events on pulls to A, some of which might be from B causing a loop. Eventually A finds out B is a gateway and so undoes all the announcements it made for B, so all that extra work is unneeded. This PR is an attempt to avoid the extra work at the cost of delaying announcement of non-gateway nodes that are slow to respond to Service API requests.

PR #469, opened 2025-07-25
system/gateway/internal/rx: make writes block waiting for listeners

mattnathan:
The previous behaviour was causing inconsistencies in the Sub methods, the value and chan did not agree on the state they returned causing duplicates in the best case, and potential data loss in the worst case.

Making the write synchronous makes deadlocks (caused by bad code) more likely, but we should be able to keep on top it.


hexaglow:
@mattnathan it sometimes works and sometimes fails. The latest CI run has an example.

hexaglow (approved):
I'm satisfied the hanging test issue is present elsewhere and not affected by this code.

PR #471, opened 2025-07-29
driver/helvanet: set confidence of pir sensors to 1 all the time

dloveen-v:
Confidence of 0 are filtered by some automations. This guarantees the occupancy state is used during exports.


mattnathan (approved):
A confidence of 0 indicates it is unknown, autos should not be ignoring Occupancy with a 0 confidence. Having said that, this change seems reasonable.

PR #476, opened 2025-08-06
driver/helvarnet: add support for emergency light tests

deanredfern26:
adds support to the Helvarnet driver to run emergency light tests and get the results. Uses the DALI proto API, while Helvarnet itself is not DALI, it is basically just a wrapper around DALI lights.

Tested on invoking function tests and querying the results. Testing duration tests is a bit more invloved, I will need to arrange this but the logic is the same.

At the moment this does not use the GetTestTime commands, see the todo, support for these will follow once I have added it into the DALI proto

### DIFF
--- a/pkg/driver/helvarnet/light.go
+++ b/pkg/driver/helvarnet/light.go
@@ -22,6 +22,9 @@ import (
 type Light struct {
 	gen.UnimplementedStatusApiServer
 	traits.UnimplementedLightApiServer
+	// Helvarnet is not really DALI, but it is really just a wrapper around DALI devices
+	// the DALI API has the stuff in for emergency lighting tests, so just use it for now.
+	gen.UnimplementedDaliApiServer
 
 	brightness *resource.Value // *traits.Brightness
 	client     *tcpClient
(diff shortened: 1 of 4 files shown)


pkg/driver/helvarnet/light.go
```diff
@@ -22,6 +22,9 @@ import (
 type Light struct {
 	gen.UnimplementedStatusApiServer
 	traits.UnimplementedLightApiServer
+	// Helvarnet is not really DALI, but it is really just a wrapper around DALI devices
```
dloveen-v, line 25: This feels like a hanging piece of work that isn't resolved by the comment alone. Can a ticket be created for this please?

pkg/driver/helvarnet/light.go
```diff
@@ -249,3 +252,187 @@ func (l *Light) runHealthCheck(ctx context.Context, t time.Duration) error {
...
+		if err != nil {
+			l.logger.Error("Failed to get duration test result", zap.String("name", l.conf.Name), zap.Error(err))
+			return nil, status.Error(codes.Internal, "failed to get duration test result")
+		}
+	default:
+		l.logger.Error("Unsupported test type requested", zap.String("name", l.conf.Name), zap.String("test", req.Test.String()))
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("unsupported test type %s", req.Test.String()))
+	}
+
+	pass := false
```
dloveen-v, line 430: nit picking: `pass := eState != nil && *eState == Pass`

PR #480, opened 2025-08-11
node: store device information in a devicespb.Collection

mattnathan:
The devices announced on a Node are presented to clients, both network and in-process, in three ways. Via children of the ParentApi when using the nodes name, via the MetadataApi, and via the DevicesApi. The MetadataApi and DevicesApi were both backed by a metadatapb.Collection, the ParentApi was backed by a parentpb.Model.

This change moves the backing store for all these into a (new) devicespb.Collection adjusting or introducing api servers that are now backed directly by this type.

For the DevicesApi the server got a bit simpler as conversion from a metadata collection to devices collection no longer needs to happen. As a side effect I've also adjusted the headers for the devices download feature to change "md.name" to just "name" now that it's backed with a Device instead of purely based on Metadata.

For the MetadataApi and ParentApi I've introduced some internal service impls that back off to various apis the Node now has for accessing devices.

Finally, I've updated usage of Node.ListAllMetadata to use the new ListDevices api instead of deprecating it. I couldn't find any use in dependent projects.


hexaglow (changes requested):
I'm not a fan of the tiny packages `metadatadevices`, `page` and `parentdevices`. As far as I can tell, these are only used from the `node` package. I think it would be better to make them part of the node package with private types. We've already got a lot of packages and deep nesting and I think it would be good to rein this in a little.

None of the types are public anyway so we can easily refactor in the future if they need to be used from multiple places.

mattnathan:
> I'm not a fan of the tiny packages `metadatadevices`, `page` and `parentdevices`. As far as I can tell, these are only used from the `node` package. I think it would be better to make them part of the node package with private types. We've already got a lot of packages and deep nesting and I think it would be good to rein this in a little.
> 
> None of the types are public anyway so we can easily refactor in the future if they need to be used from multiple places.

I'd like to discuss this a little more as I disagree, my use of these tiny packages was very deliberate in this case (with one exception which I'll mention at the end).

I've been getting confused lately with some of our more complex packages (node, internal/manage/device, etc). I think my issue is when trying to distinguish between second and third level functionality (see **aside** for what I mean by _level_). My confusion when all these levels are in the same package is that there's no (built-in) way to identify which symbols of each level the level above should be using, and arguably more importantly which it should _not_ be using.

**aside** I've just made up the _levels_ terminology for this discussion, but imagine exported package APIs being first level functionality, the types/funcs used to implement the first level is the second level, the third level is used to implement the second, and so on. Using this PR as an example: `Node.New` (and the registered services) is first level, `parentdevices.Server` is second level, and `parentdevices.deviceChangeToProto` is third level. In general, if there are multiple second levels they tend to have independent sets of third levels. **/aside**

In our existing packages we've been (inconsistently) using naming conventions to split the independent level functionality. Using this PR again, this would look like a `type parentDevicesServer` and `func deviceChangeToChildrenResponseProto` each placed into a `parentdevices.go` file.

Another solution we've also used is to pull the second level (or even third level) functionality into a new independent package, sometimes in (top level) `internal` sometimes in `pkg/util`. I did consider this for this PR, the functionality of _a ParentApi server backed by a devicespb.Collection_ is independent of node and might be useful in other places.

Both approaches have their drawbacks: 
- Using the same package relies on naming convention and introduces a lot of extra typing/reading to add the name-spacing for the independent bits. This also has a tendency to become inconsistent as there's nothing to enforce the naming convention.
- An independent package adds weight in proportion to its visibility, both to the new package as it now needs to be designed for wider use, and to the repo as a whole. It's not obvious when looking at the new package where it _should_ be used.

For these reasons I decided to introduce node-internal packages for these second-level features so that each feature could export the parts of that feature the first level node functionality should be using, while also being able to hide away third level types/funcs without any significant custom name spacing.

_The Exception_: The `internal/page` package, as it stands in this PR, should really be part of the `internal/parentdevices` package, this is the only package that uses it. It just so happens that I wrote the `page` package alongside the `parentdevices` package under the impression that I'd need it for the `metadatadevices` pacakge, mis-remembering that the MetadataApi doesn't have paging. I left it in because there wasn't a significant difference in my head between the two options (`page/page.go` and `parentdevices/page.go`) to warrant me spending time refactoring.

hexaglow (changes requested):
Thanks for the explanation. Could you tidy this bit up?

> The internal/page package, as it stands in this PR, should really be part of the internal/parentdevices package, this is the only package that uses it.

hexaglow (approved):

PR #485, opened 2025-08-14
feat/emergency light trait

deanredfern26:
Adds an emergency light trait

### DIFF
--- a/pkg/gentrait/emergencylightpb/model.go
+++ b/pkg/gentrait/emergencylightpb/model.go
@@ -3,13 +3,13 @@ package emergencylightpb
 import (
 	"context"
 	"math/rand"
-	"time"
 
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/smart-core-os/sc-golang/pkg/resource"
 	"github.com/vanti-dev/sc-bos/pkg/gen"
+	"github.com/vanti-dev/sc-bos/pkg/util/resources"
 )
 
 type Model struct {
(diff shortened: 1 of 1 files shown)


deanredfern26:
Yeah fair enough I think it was trying to do too much at once. I will just do the emergency-light trait for now and keep lighting test in so EW can carry on using it until I can get round to migrating it over there. As for historical data I wasn't planning on keeping it, is this a requirement? I thought they would just go in and download the report when they needed it

proto/emergency_light.proto
```diff
@@ -0,0 +1,100 @@
...
+    NO_TEST = 1;
+    FUNCTION_TEST = 2;
+    DURATION_TEST = 3;
+  }
+
+  string name = 1;
+  // Specify the test type to get results for. If omitted, results for all test types will be returned.
+  optional TestType test = 2;
+  // If an implementing driver supports caching of test results, this field can be set to true to use the cached results.
+  bool use_cache = 3;
```
mattnathan, line 83: I really don't like this property. I expect there's a specific reason stemming from the driver you're implementing, but it feels very out of place here.

If we do provide a way for the client to affect caching of data, the default should give power to the server to implement it how they want. This doesn't do that, it implies no caching will be performed unless asked for. Caching like this usually looks like no_cache or etag or max-age related properties, etc.

proto/emergency_light.proto
```diff
@@ -0,0 +1,100 @@
...
+}
+
+message PullTestResultRequest {
+  string name = 1;
+  google.protobuf.FieldMask read_mask = 2;
```
mattnathan, line 88: Whatever solution is decided on for the TestType comment above, it should also be applied here.

mattnathan (changes requested):

proto/emergency_light.proto
```diff
@@ -0,0 +1,98 @@
...
+    TEST_PASSED = 1;
+    CIRCUIT_FAILURE = 2;
+    BATTERY_DURATION_FAILURE = 3;
+    BATTERY_FAILURE = 4;
+    LAMP_FAILURE = 5;
+    TEST_FAILED = 6;
+    LIGHT_FAULTY = 7;
+    COMMUNICATION_FAILURE = 8;
+    OTHER_FAULT = 9;
+    TEST_RESULT_PENDING = 10;
```
mattnathan, line 44: This is a minor thing, but I find it useful to know that enum values over a certain ordinal are classed together. For example

```go
if t.Result.Number() > 2 {
  return err
}
```

It means you can check for errors and still react correctly even if new error states are added in the future. We can't do that with this enum ordering.

Consider putting PENDING near the top, maybe at `1` and move all the others down?

mattnathan (changes requested):
Looks pretty good now. Update the enum and it's good to merge.

mattnathan (approved):

PR #486, opened 2025-08-14
proto/valve: generate a valve proto definition with go and js implementations

dloveen-v:


proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+  // openness of the valve sensor, 0-100%
+  // read_only
+  float pumpage_sensor = 3;
+  // openness of the valve's pump set point, 0-100%
+  float pumpage_set_point = 4;
+  // temperature sensor e.g. C
+  // read_only
+  float temperature_sensor = 5;
+  // temperature set point e.g. C
+  float temperature_set_point = 6;
```
mattnathan, line 31: We already have a Temperature trait the valve device can implement

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+  // flow rate set point e.g. L/s
+  float flow_rate_set_point = 1;
+  // flow rate sensor e.g. L/s
+  // read_only
+  float flow_rate_sensor = 2;
+  // openness of the valve sensor, 0-100%
+  // read_only
+  float pumpage_sensor = 3;
+  // openness of the valve's pump set point, 0-100%
+  float pumpage_set_point = 4;
```
mattnathan, line 26: what is pumpage?

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+  rpc PullValve(PullValveRequest) returns (stream PullValveResponse) {}
+  rpc UpdateValve(UpdateValveRequest) returns (Valve) {}
+}
+
+message Valve {
+  // flow rate set point e.g. L/s
+  float flow_rate_set_point = 1;
+  // flow rate sensor e.g. L/s
+  // read_only
+  float flow_rate_sensor = 2;
```
mattnathan, line 21: `_sensor` is not required for any of these properties

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+  // pressure sensor e.g. bar
+  // read_only
+  float pressure_sensor = 7;
+  // pressure set point e.g. bar
+  float pressure_set_point = 8;
+  // drive frequency sensor e.g. Hz
+  // read_only
+  float drive_frequency_sensor = 9;
+  // drive frequency set point e.g. Hz
+  float drive_frequency_set_point = 10;
```
mattnathan, line 41: what is drive frequency? I can't think how that's related to a valve

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+
+  enum ValveMode {
+    VALVE_MODE_UNKNOWN = 0;
```
mattnathan, line 44: `_UNSPECIFIED`

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+    VALVE_MODE_FLOW = 1;
+    VALVE_MODE_RETURN = 2;
+    VALVE_MODE_BLOCKED = 3;
```
mattnathan, line 47: Don't need `VALVE_MODE_` when using nested enums

mattnathan, line 47: BLOCKED isn't a mode, it's a state, along with OPEN, STUCK, etc.

mattnathan, line 47: What does blocked mean? that the valve can't operate or the fluid can't flow?

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+import "google/protobuf/field_mask.proto";
+import "google/protobuf/timestamp.proto";
+
+service ValveApi {
+  rpc GetValve(GetValveRequest) returns (Valve) {}
+  rpc PullValve(PullValveRequest) returns (stream PullValveResponse) {}
+  rpc UpdateValve(UpdateValveRequest) returns (Valve) {}
+}
+
+message Valve {
```
mattnathan, line 16: A valve is a type of device, I'm not convinced it describes a trait. Maybe `Flow` is more apt as a trait. A valve controls the _flow_ of something. Flow has a rate and direction, it can have pressure, it can be open/closed, etc.

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
+syntax = "proto3";
+
+package smartcore.bos;
+
+option go_package = "github.com/vanti-dev/sc-bos/pkg/gen";
+
+import "google/protobuf/field_mask.proto";
+import "google/protobuf/timestamp.proto";
+
+service ValveApi {
```
mattnathan, line 10: All services, especially traits need docs explaining what they model and what kinds of device they apply to. There's a good chance your trait is not a trait if you can only think of one kind of device it applies to.

proto/valve.proto
mattnathan, line 21: Zero values for non-optional fields should always be documented

mattnathan, line 21: Docs should be proper english (Capitalised sentences, punctuation, not snake_case) and should explain what the field represents. These aren't any of those things.

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+  rpc GetValve(GetValveRequest) returns (Valve) {}
+  rpc PullValve(PullValveRequest) returns (stream PullValveResponse) {}
+  rpc UpdateValve(UpdateValveRequest) returns (Valve) {}
+}
+
+message Valve {
+  // flow rate set point e.g. L/s
+  float flow_rate_set_point = 1;
+  // flow rate sensor e.g. L/s
+  // read_only
```
mattnathan, line 20: We try to use _output only_

proto/valve.proto
```diff
@@ -0,0 +1,83 @@
...
+  // temperature sensor e.g. C
+  // read_only
+  float temperature_sensor = 5;
+  // temperature set point e.g. C
+  float temperature_set_point = 6;
+  // pressure sensor e.g. bar
+  // read_only
+  float pressure_sensor = 7;
+  // pressure set point e.g. bar
+  float pressure_set_point = 8;
```
mattnathan, line 36: Maybe pressure is a trait that more than valves can implement?

proto.mod
mattnathan, line 1: This change is not related to any of the other changes in this PR.

proto/valve.proto
dloveen-v: How often the valve pumps I guess. It was a property in on-site documentation.

proto/valve.proto
dloveen-v: Ahh this should read `BLOCKING`.

proto.mod
dloveen-v: It is required to generate the proto unless I am using the wrong script.

PR #488, opened 2025-08-14
Support container image builds from any commit

hexaglow:
The existing action `docker-package` uses prebuilt artefacts from releases to built the containers.

This means that it can only build for releases, and because it uses a different Dockerfile to the one at the root of the repo, that creates potential drift.

Switch to building the container image using Docker Buildx.

Images are tagged automatically based on Git tags (if present) and commit hash. A custom image tag can be specified if the workflow is manually dispatched.

As future work, we could enable automatic nightly builds.


mattnathan (approved):
Looks good, nice to have a standard way to do this

PR #489, opened 2025-08-15
util/page: new package for paging though lists of items

mattnathan:
I've updated places that can use this new util, which were fewer than I thought but still more than one.


deanredfern26 (approved):
Looks fine, is it worth making the check in `capPageSize` to be <= 0 instead of ==?

PR #492, opened 2025-08-20
auto/statusemail;auto/statusalerts;auto/udmi: change readPaths top level message validation type to *gen.Device

dloveen-v:
I cannot find any references to where the top level message struct name needs to be included in the fieldmask path.

I am pretty sure the string path shouldn't include the top level message name and have found the documented example doesn't either. Although, the documentation does mislead us by suggesting the top level message name is part of the field mask.

// As an example, consider the following message declarations:
//
//     message Profile {
//       User user = 1;
//       Photo photo = 2;
//     }
//     message User {
//       string display_name = 1;
//       string address = 2;
//     }
//
// In proto a field mask for `Profile` may look as such:
//
//     mask {
//       paths: "user.display_name"
//       paths: "photo"
//     }

Where it misleads us:
// `FieldMask` represents a set of symbolic field paths, for example:
//
//     paths: "f.a"
//     paths: "f.b.d"
//
// Here `f` represents a field in some root message, `a` and `b`
// fields in the message found in `f`, and `d` a field found in the
// message in `f.b`.

I am surprised this was working for so long. I will try deploying this branch to the site experiencing the panic to see if it resolves the issue.

### DIFF
--- a/pkg/auto/statusalerts/statusalerts.go
+++ b/pkg/auto/statusalerts/statusalerts.go
@@ -8,7 +8,6 @@ import (
 
 	"go.uber.org/zap"
 
-	"github.com/smart-core-os/sc-api/go/traits"
 	"github.com/smart-core-os/sc-golang/pkg/resource"
 	"github.com/vanti-dev/sc-bos/pkg/auto"
 	"github.com/vanti-dev/sc-bos/pkg/auto/statusalerts/config"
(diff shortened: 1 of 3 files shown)


dloveen-v:
I have deployed to the building-controller and area-controllers on-site and the offending automations are running without panics now.

mattnathan (changes requested):
The issue here is the top level message type is `*gen.Device` not `*traits.Metadata`. The DevicesApi should receive read paths relative to this type.

~~I am also surprised this has worked, but the paths themselves are correct, just not the type they are _supposed to be_ validated against in this `resource.WithReadPaths` func.~~

I accidentally introduced this issue in the recent https://github.com/vanti-dev/sc-bos/pull/480 - sorry.

dloveen-v:
Right. I think me reading the documentation mislead me to believing the documentation mislead the initial implementation. I have changed it but let me test on site first.

mattnathan (approved):

PR #493, opened 2025-08-21
auto/exporthttp: check bolt db for previous stored execution time to prevent unnecessary immediate executions

dloveen-v:


pkg/auto/exporthttp/job/job_test.go
```diff
@@ -0,0 +1,106 @@
...
+			previous: makeTime("09:00"),
+			expected: false,
+		},
+		{
+			name:     "nil previous execution",
+			schedule: jsontypes.MustParseSchedule("0 9 * * *"),
+			now:      makeTime("09:00"),
+			previous: time.Time{},
+			expected: true,
+		},
```
mattnathan, line 87: There are no tests that check changes to the schedule, or I guess if `previous` doesn't align with the schedule.

dloveen-v: The idea is the check for immediate execution is done on application start, not when the automation is started. You raise a good point though, I didn't think about someone starting the automation via the SmartCore API.

pkg/auto/exporthttp/job/job.go
```diff
@@ -28,15 +29,21 @@ var (
...
 // Job represents an exporthttp automation task that executes Do to send a POST request
 type Job interface {
 	GetName() string
 	GetUrl() string
 	GetSite() string
-	GetExecutionAfter(t time.Time) <-chan time.Time
-	SetPreviousExecution(t time.Time)
+	GetExecutionAfter(autoName, scName, jobName string) <-chan time.Time
+	SetPreviousExecution(autoName, scName, jobName string, t time.Time)
+	GetSchedule() *jsontypes.Schedule
```
mattnathan, line 46: Why pass all these args into the func? Can't the job just know these things? The job is created once during the config processing after all, so it knows these at that time.

pkg/auto/exporthttp/job/job.go
```diff
@@ -56,7 +63,7 @@ func (m *Mulpx) WaitForDone() {
 }
 
 // Multiplex receives multiple jobs and fans all the jobs into a single chan
-func Multiplex(ctx context.Context, jobs ...Job) *Mulpx {
+func Multiplex(ctx context.Context, autoName, scName string, jobs ...Job) *Mulpx {
```
mattnathan, line 66: See previous comment for why these new string args aren't needed.

pkg/auto/exporthttp/job/job_test.go
```diff
@@ -0,0 +1,120 @@
...
+		schedule *jsontypes.Schedule
+		now      time.Time
+		previous time.Time
+		expected bool
+	}{
+		{
+			name:     "now is behind the next execution",
+			schedule: jsontypes.MustParseSchedule("0 9 * * *"),
+			now:      makeTime("08:59").Add(24 * time.Hour),
+			previous: makeTime("09:00"),
```
mattnathan, line 22: I still think these tests will be clearer if previous was in the past instead of now being in the future.

mattnathan (changes requested):

pkg/auto/exporthttp/job/job_test.go
dloveen-v: Yes, I avoided doing this because the times are parsed to zero time so going further in the past doesn't make sense numerically to me. I'll give it a go though

pkg/auto/exporthttp/job/job.go
```diff
@@ -99,22 +124,53 @@ func (b *BaseJob) GetUrl() string {
...
+
+func (b *BaseJob) GetScName() string {
+	return b.ScName
+}
+
+func (b *BaseJob) GetExecutionAfter(jobName string) <-chan time.Time {
+	t := time.Now().UTC()
+	// check the previous execution timestamp
+	previous := time.Time{}
+	key := fmt.Sprintf(boltKeyTemplate, b.GetAutoName(), b.GetScName(), jobName)
```
mattnathan, line 139: If you use the field `b.AutoName` and `b.ScName` here you can remove the methods from the Job interface.

pkg/auto/exporthttp/job/job.go
```diff
@@ -68,13 +78,11 @@ func Multiplex(ctx context.Context, jobs ...Job) *Mulpx {
...
 		out.group.Go(func() error {
-			out.C <- j
-			current := time.Now().UTC()
 			for {
 				select {
-				case <-j.GetExecutionAfter(current):
+				case <-j.GetExecutionAfter(j.GetName()):
 					out.C <- j
-					current = time.Now().UTC()
+					j.SetPreviousExecution(j.GetName(), time.Now())
```
mattnathan, line 85: The job already knows its own name, no need to pass it in.

mattnathan (changes requested):
Also, your PR title is not a _summary_ of what is changing

pkg/auto/exporthttp/job/job.go
dloveen-v: The name is defined in the `Job` implementation struct, not the `BaseJob` anonymous struct.

dloveen-v: Ahh I see what you mean now. I have amended.

pkg/auto/exporthttp/job/job.go
```diff
@@ -28,15 +29,22 @@ var (
...
 
 // Job represents an exporthttp automation task that executes Do to send a POST request
 type Job interface {
-	GetName() string
 	GetUrl() string
 	GetSite() string
-	GetExecutionAfter(t time.Time) <-chan time.Time
+	GetName() string
+
+	GetExecutionAfter() <-chan time.Time
```
mattnathan, line 45: The name of this method no longer makes sense, _after_ what?

pkg/auto/exporthttp/job/job.go
```diff
@@ -28,15 +29,22 @@ var (
...
 	SetPreviousExecution(t time.Time)
+	GetSchedule() *jsontypes.Schedule
```
mattnathan, line 47: I think this method can also be removed, it's only ever used from within other methods on Job

mattnathan (changes requested):

PR #494, opened 2025-08-22
auth: fix tenant (client credentials) auth

mattnathan:
Using it in a cohort would crash the gateway, using it generally wouldn't work due to enum comparison issues.


hexaglow (approved):
Discussed this on Friday

PR #499, opened 2025-08-29
ui/ops: refactor emergency light page to use emergency-light trait

deanredfern26:
refactor the emergency light page to use the new trait, add a few emergency lights to the ugs example and fix the broken functions in the model


deanredfern26:
[SCB-1184](https://vanti.youtrack.cloud/issue/SCB-1184) Add ability to filter emergency lights by floor

PR #502, opened 2025-09-05
driver/bacnet: support transport trait

dloveen-v:


pkg/driver/bacnet/merge/transport.go
```diff
@@ -0,0 +1,476 @@
...
+		}
+	}
+	status.UpdatePollErrorStatus(t.statuses, t.config.Name, "TransportInfo", requestNames, errs)
+	if len(errs) > 0 {
+		return multierr.Combine(errs...)
+	}
+	return nil
+}
+
+func (t *transport) pollPeer(ctx context.Context) (*gen.Transport, error) {
```
deanredfern26, line 143: I know that this function is just following the pattern from other traits, but it is already very long and will only get bigger. Is it worth changing resProcessors to `[]func(response any, data *gen.Transport)` and pulling the processors out of this function?

Going forward this `pollPeer` is going to grow really long, not just for this trait, I think it will help to do something about it sooner

PR #504, opened 2025-09-05
history: add SQLite3-based history storage

hexaglow:
Add a `history.Store` implementation using SQLite3. Features:
- Local storage, no server required
- Source names normalised to reduce storage space
- ID and timestamp stored combined for small record size and fast queries

Limitations:
- Max of 1,000,000 records with the same timestamp (millisecond precision) due to how IDs and timestamps are handled.

### DIFF
--- a/pkg/app/stores/stores.go
+++ b/pkg/app/stores/stores.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
 	"sync"
 
 	"github.com/jackc/pgx/v4/pgxpool"
 	"go.uber.org/multierr"
+	"go.uber.org/zap"
 
 	"github.com/vanti-dev/sc-bos/internal/util/pgxutil"
+	"github.com/vanti-dev/sc-bos/pkg/history/sqlitestore"
 )
 
 var (
@@ -20,29 +23,46 @@ var (
 // Config configures shared storage (dbs) for systems on this node.
 type Config struct {
 	Postgres *PostgresConfig `json:"postgres,omitempty"`
+	// Local directory for storing database files.
+	DataDir string      `json:"-"`
+	Logger  *zap.Logger `json:"-"`
 }
 
 type PostgresConfig struct {
 	pgxutil.ConnectConfig
 }
 
+// New creates a new Stores instance based on cfg, which must be non-nil.
 func New(cfg *Config) *Stores {
-	s := &Stores{}
-	if cfg != nil && cfg.Postgres != nil {
+	logger := cfg.Logger
+	if logger == nil {
+		logger = zap.NewNop()
+	}
+
+	s := &Stores{
+		sqliteHistoryStore: sqliteHistoryStore{
+			path: filepath.Join(cfg.DataDir, defaultSqliteHistoryPath),
+		},
+	}
+	if cfg.Postgres != nil {
 		s.postgresStore.cfg = cfg.Postgres
 	}
 	return s
 }
 
+const defaultSqliteHistoryPath = "data/history.sqlite3"
+
 // Stores provides access to shared storage connections/clients.
 type Stores struct {
 	postgresStore
+	sqliteHistoryStore
 }
 
 // Close closes all stores.
 func (s *Stores) Close() error {
 	return multierr.Combine(
 		s.postgresStore.close(),
+		s.sqliteHistoryStore.close(),
 	)
 }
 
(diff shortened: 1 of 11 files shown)


pkg/history/sqlitestore/store.go
```diff
@@ -0,0 +1,645 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	// Query the ID of the source
+	var id int64
+	err = sa.queryStmt.QueryRowContext(ctx, source).Scan(&id)
+	if err != nil {
+		return 0, err
+	}
```
mattnathan, line 582: I'm curious why you've gone with `INSERT IGNORE` + `SELECT` instead of `SELECT` + `if absent { INSERT RETURNING ID }`. I feel like the latter is more idiomatic and potentially more performant as you could skip a statement.

pkg/history/sqlitestore/store_test.go
```diff
@@ -0,0 +1,470 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestDatabase_Insert(t *testing.T) {
+	db := newTestDB(t)
+	ctx := context.Background()
```
mattnathan, line 20: `ctx := t.Context()`?

pkg/history/sqlitestore/store_test.go
```diff
@@ -0,0 +1,470 @@
...
+		}
+
+		diff := cmp.Diff(expected[i], retrievedRecords[i],
+			cmpopts.IgnoreFields(Record{}, "ID"),
+			cmpopts.EquateApproxTime(time.Millisecond),
+		)
+		if diff != "" {
+			t.Errorf("record %d: data mismatch (-want +got):\n%s", i, diff)
+		}
+	}
```
mattnathan, line 88: You could probably do this entire comparison using `cmp.Diff` which might help with added/deleted comparisons too

pkg/history/sqlitestore/store_test.go
```diff
@@ -0,0 +1,470 @@
...
+	records := []Record{
+		{Source: "duplicate-source", CreateTime: time.Now(), Payload: []byte("payload-1")},
+		{Source: "duplicate-source", CreateTime: time.Now(), Payload: []byte("payload-2")},
+	}
+
+	err := db.InsertBulk(ctx, records)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
```
mattnathan, line 125: I feel like there should be more to this test, maybe a `verifyRecords` call?

pkg/history/sqlitestore/store_test.go
```diff
@@ -0,0 +1,470 @@
...
+	if size <= 0 {
+		t.Errorf("expected database size to be greater than 0, got %d", size)
+	}
+}
+
+func TestDatabase_LargeScale(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping large scale test in short mode")
+	}
```
mattnathan, line 331: For this kind of test I think it'd be better to look for an env var (or something) instead of using `-short`. This isn't really testing anything, it's more about discovery and manual checking of performance.

pkg/history/sqlitestore/store.go
hexaglow: Changed the implementation as suggested.

pkg/app/stores/stores.go
```diff
@@ -20,29 +23,46 @@ var (
...
-	s := &Stores{}
-	if cfg != nil && cfg.Postgres != nil {
+	logger := cfg.Logger
+	if logger == nil {
+		logger = zap.NewNop()
+	}
+
+	s := &Stores{
+		sqliteHistoryStore: sqliteHistoryStore{
+			path: filepath.Join(cfg.DataDir, defaultSqliteHistoryPath),
```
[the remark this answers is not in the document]
hexaglow: I can't find a package called `paths`, where's it located?

pkg/history/sqlitestore/store.go
```diff
@@ -0,0 +1,645 @@
...
+	return RecordID(int64(id) + 1), true
+}
+
+func (id RecordID) String() string {
+	return fmt.Sprintf("%016X", int64(id))
+}
+
+type sourceAllocator struct {
+	sources    map[string]int64 // source name to ID mapping
```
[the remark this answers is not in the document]
hexaglow: It seems applicable but also a premature optimisation? My intuition is that the map maintenance is a very small part of the overall transaction time.

pkg/history/sqlitestore/store_test.go
```diff
@@ -505,3 +561,160 @@ func newTestDB(tb testing.TB) *Database {
...
+	if err != nil {
+		t.Fatalf("unexpected Append error: %v", err)
+	}
+	diff := cmp.Diff(history.Record{
+		CreateTime: time.Now(),
+		Payload:    []byte("test-payload"),
+	}, record,
+		cmpopts.IgnoreFields(history.Record{}, "ID"),
+		// we can't control exactly what time is used, but it shouldn't vary much from now
+		cmpopts.EquateApproxTime(time.Second),
```
mattnathan, line 711: Would this get better with synctest?

mattnathan (approved):

pkg/history/sqlitestore/store_test.go
hexaglow: Yes, because synctest has a deterministic clock so we'd know exactly what time to expect.

PR #507, opened 2025-09-10
ui/ops: add electric demand history widget

mattnathan:
Similar to the energy history meter, but based on the electric trait instead. See the readme for a little more detail and a screen shot.


deanredfern26 (changes requested):
I think as part of this PR it would be nice to add `electric` support to the db seeder tool.
Shouldn't be too much more work and will help me test it out

PR #519, opened 2025-09-24
feat/zone on off & ui on/off support

deanredfern26:
adds on/off support to zones, the bacnet driver and also a on/off card for the ops ui


pkg/zone/feature/onoff/config/root.go
```diff
@@ -0,0 +1,11 @@
+package config
+
+import (
+	"github.com/vanti-dev/sc-bos/pkg/zone"
+)
+
+type Root struct {
+	zone.Config
+
+	OnOff []string `json:"onOff,omitempty"`
```
mattnathan, line 10: This should be plural to match the other features.

pkg/zone/feature/onoff/group.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, status.Errorf(codes.FailedPrecondition, "no onoff devices in group")
+	case 1:
+		return all[0], nil
+	default:
+		state := all[0].State
+		var err error
+		// either they are all off, all on or it is an error
+		// this should really return 'Mixed' or something I think, but we don't have that in the trait
+		for _, v := range all[1:] {
+			if v == nil || v.State != state {
```
mattnathan, line 178: I think you should be able to ignore unspecified responses here, they shouldn't trigger the error.

pkg/zone/feature/onoff/group.go
```diff
@@ -0,0 +1,210 @@
...
+				break
+			}
+		}
+		return &traits.OnOff{
+			State: state,
+		}, err
+	}
+}
+
+func mergeOnOffMap(all map[string]*traits.OnOff, v *traits.OnOff) (*traits.OnOff, error) {
```
mattnathan, line 190: I'd like to see these two merge funcs combined, maybe using an `iter.Seq[*traits.OnOff]` as the arg, so we don't have to fix issues in two places.

mattnathan (changes requested):

pkg/zone/feature/onoff/group.go
deanredfern26: Ah thank you! I looked for something like this in the docs but couldn't find it.

pkg/zone/feature/onoff/group.go
```diff
@@ -0,0 +1,192 @@
...
+		if v.State != state {
+			state = traits.OnOff_STATE_UNSPECIFIED
+			err = multierr.Append(err, status.Errorf(codes.FailedPrecondition, "not all onoff devices have the same state"))
+			break
+		}
+	}
+	return &traits.OnOff{
+		State: state,
+	}, err
+}
```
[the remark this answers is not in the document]
mattnathan: You're still using Pull.

Something like this:

```go
var state traits.OnOff_State
var got bool
for v := range seq {
  got = true
  switch {
  case v.State == UNSPECIFIED:
    // skip
  case state == UNSPECIFIED:
    state = v.State
  case state != v.State:
    return nil, status.Errorf(codes.FailedPrecondition, "not all onoff devices have the same state")
  }
}
```

mattnathan (changes requested):

pkg/zone/feature/onoff/group.go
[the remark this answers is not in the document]
mattnathan: That's not what I see on GH

Image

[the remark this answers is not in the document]
deanredfern26: OK you mean don't use pull at all I will change it

mattnathan (approved):

PR #521, opened 2025-09-25
Fix out-of-bounds crash inside SQLite3

hexaglow:
go-sqlite3 runs inside WASM using wazero.

Older versions of wazero had an incompatibility with Go 1.24+ that could cause a crash with error "panic: wasm error: out of bounds memory access"

Pull in newer go-sqlite3 and therefore also wazero.


deanredfern26:
The test job is now failing, can you check if this is a result of the various package upgrades?

hexaglow:
@deanredfern26 the problem has been fixed, and it's ready for review

PR #523, opened 2025-09-26
ui/ops: fix total count for collections drifting after pull

mattnathan:
The total count was not taking the items added or deleted after the initial list request into account when reporting the total item count. Now it does.

### DIFF
--- a/ui/ops/src/composables/collection.js
+++ b/ui/ops/src/composables/collection.js
@@ -294,7 +294,7 @@ export default function useCollection(request, client, options) {
   function processChanges() {
     const _items = items.value;
     const _changes = unprocessedChanges.value;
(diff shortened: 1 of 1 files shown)


ui/ops/src/composables/collection.js
```diff
@@ -283,6 +288,8 @@ export default function useCollection(request, client, options) {
    * unprocessedChanges will be empty after this call.
    *
    * If options.cmp is not defined, items created via pull will not be added to the list.
+   *
+   * @return {number} delta of items added/removed from the items list
```
deanredfern26, line 292: Am getting a lint warning for this return value

286:3  warning  JSDoc @return declaration present but return expression not available in function  jsdoc/require-returns-check

PR #526, opened 2025-09-29
fix/energy graph performance

deanredfern26:
Make interpolation optional, this massively speeds up graph rendering.
If interpolation is required for more accurate graphs in the future, a start time will need to be provided for `getReadingBefore` to work fast enough.
This is a bit of a quick fix as I don't want to make massive changes to the graph rendering stuff here.

### DIFF
--- a/ui/ops/src/traits/meter/meter.js
+++ b/ui/ops/src/traits/meter/meter.js
@@ -137,9 +137,10 @@ export function useMeterReading(value, support = null) {
  *
  * @param {import('vue').MaybeRefOrGetter<string>} name
  * @param {import('vue').MaybeRefOrGetter<Date>} t
+ * @param {boolean} interpolate - whether to interpolate between readings before and after t, or just use the next reading after t
  * @return {import('vue').ComputedRef<null | MeterReading.AsObject>}
  */
-export function useMeterReadingAt(name, t) {
+export function useMeterReadingAt(name, t, interpolate = false) {
   const usageAtT = ref(/** @type {null|number} */ null);
   const readingAtT = computed(() => {
     if (usageAtT.value === null) return null;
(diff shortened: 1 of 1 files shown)


ui/ops/src/traits/meter/meter.js
```diff
@@ -169,10 +170,21 @@ export function useMeterReadingAt(name, t) {
     // see: https://github.com/grpc/grpc-web/issues/946
     fetching.value = {t, cancel: () => {}};
     try {
-      const [before, after] = await Promise.all([
-        getReadingBefore(name, t),
-        getReadingOnOrAfter(name, t)
-      ]);
+      let before, after;
+      // trying to interpolate with the reading before can really slow things down as startTime is not set in getReadingBefore
+      // only do this if we care about the graph being super accurate. for most cases this won't matter
```
dloveen-v, line 175: Is this true? Accuracy for meter graphs doesn't matter for most cases depends on who the user is and not which UI is rendering the graph.

I took a look at the PostgreSQL specification: 
> From the [PostgreSQL 15 documentation on aggregate functions (https://www.postgresql.org/docs/current/functions-aggregate.html#FUNCTIONS-AGGREGATE-TABLE)
:

> Aggregate functions such as MAX() and MIN() compute the result over all input rows. They generally require scanning all rows that satisfy the WHERE clause, unless an index-only plan can be applied.

> Key point: MAX(id) cannot directly jump to the highest value unless the query can use an index that matches both the filter and the ordering.



This means at https://github.com/vanti-dev/sc-bos/blob/main/pkg/history/pgxstore/store.go#L279 there is a chance that replacing `id <= (select max(id) ...` with 

```
id <= ( select ... ORDER BY id DESC limit 1)
```

has the potential to reduce the query time (the need for a full table or index scan) when only passing `endtime`. The disclaimer is I cannot get pgadmin to connect to the OCW db instance over a vpn so this is an untested theory. I think it is worth trying though?

dloveen-v: Of course this will mean changing the query builder to use that query format if the theory is true.

deanredfern26: There is no point adding a button with the way this currently works. Enabling interpolation is simply unusable at OCW as trying to use it takes minutes for the graph to render

dloveen-v (approved):
Although this is necessary to get a "fix" out there on sites. I do think we need to revisit how this is displayed and what options are made available to the end user. We potentially could be resolving this with the db migration step as mentioned in another place.

deanredfern26:
100% agree this is not a fix, just a workaround to get the graphs rendering within an acceptable time

PR #529, opened 2025-10-01
zone/meter: use the latest reading from history table if enabled in config and an error is returned on initial meter read

dloveen-v:
If a meter name in a zone fails, the current behaviour is to fail the entire zone even if other names may resolve to a valid reading in the zone.

This change aims to mitigate failures of the entire zone by using the latest reading from the history table. The reason the latest reading is needed is to prevent the total meter reading from falling (which may break widgets downstream that expect the reading to only increase over time).


pkg/zone/feature/meter/group.go
```diff
@@ -77,6 +80,23 @@ func (g *Group) GetMeterReading(ctx context.Context, request *gen.GetMeterReadin
 		i := i
 		fns[i] = func() {
 			res, err := g.apiClient.GetMeterReading(ctx, request)
+			if err != nil && g.useHistoryBackupOnErr {
+				// try to get the latest backup reading from the history table if the main read fails
+				latest, historyErr := g.historyApiClient.ListMeterReadingHistory(ctx, &gen.ListMeterReadingHistoryRequest{
```
deanredfern26, line 85: I think at the very least, we should only use data if it is within a certain time frame. Using a reading from a few hours / a day or so ago for a zone approximation I think is fine. But if the latest history reading is older than some value, then we throw the error. Also, if more than x% of meters are throwing errors getting data, then we should raise the error. 

Mitigating temporary issues with a few meters is not a bad idea IMO, however there has to be some point where we just have to say that we can't show the zone/building data as there is a major issue with the underlying data. This error needs to be surfaced on the graph though, just having an empty graph with no data and no explanation of why the data isn't there is bad.

deanredfern26: Yes more just me thinking and typing for the UI stuff, have created [SCB-1231](https://vanti.youtrack.cloud/issue/SCB-1231) Show error message on energy graph when no data is available

deanredfern26 (approved):

PR #539, opened 2025-10-08
driver/hikcentral: implement access grants rpcs in hikcentral

dloveen-v:


pkg/driver/hikcentral/ring.go
```diff
@@ -0,0 +1,80 @@
...
+		values = append(values, a.(T))
+	})
+
+	return values
+}
+
+func (r *Ring[T]) Len() int {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+	return r.Len()
```
deanredfern26, line 52: This call is recursive, will cause stack overflow

pkg/driver/hikcentral/ring.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (r *Ring[T]) Find(predicate func(T) bool) T {
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
+	var result T
+	r.head.Do(func(a any) {
+		v := a.(T)
```
deanredfern26, line 60: There is no nil checking here, there probably should be

deanredfern26 (changes requested):

pkg/driver/hikcentral/ring.go
dloveen-v: Thanks @deanredfern26 Missed accessing the `head` field.

deanredfern26:
closing as stale

PR #653, opened 2025-12-22
ui/ops: make table headers fixed when scrolling

deanredfern26:
[SCB-1166](https://vanti.youtrack.cloud/issue/SCB-1166) Fix the table header so we can always see column headers, search bar, filters etc when scrolling down the table.

### DIFF
--- a/ui/ops/src/routes/ops/waste/WasteTable.vue
+++ b/ui/ops/src/routes/ops/waste/WasteTable.vue
@@ -11,6 +11,8 @@
           v-bind="tableAttrs"
           disable-sort
           :items-length="queryTotalCount"
+          :fixed-header="props.fixedHeader"
+          :style="tableStyles"
           class="pt-4">
         <template #item.createTime="{ item }">
           {{ timestampToDate(item.wasteCreateTime).toLocaleString() }}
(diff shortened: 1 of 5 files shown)


mattnathan (changes requested):

ui/ops/src/routes/ops/waste/WasteTable.vue
```diff
@@ -107,6 +117,16 @@ const allHeaders = [
...
 ];
 
+const tableStyles = computed(() => {
+  if (!props.fixedHeader) {
+    return {};
+  }
+  if (props.height) {
+    return {height: typeof props.height === 'number' ? `${props.height}px` : props.height};
+  }
+  return {height: '80vh'};
```
mattnathan: I'm still not sure on this value. It feels like it'll cause issues for us if we ever change the layout of the pages that host this table - or even adjusting the toolbar size of the page or adding a footer or anything like that will cause issues.

What's the intent here? Could you use `position: sticky` instead of the fixed-header prop on the data-table. I _think_ the prop causes the table to control its own scroll, but the position would work across all contexts. I might have misremembered that.

mattnathan (changes requested):

PR #732, opened 2026-03-11
feat/reboot trait

deanredfern26:
Adds a Boot trait (SCB-793 (https://vanti.youtrack.cloud/issue/SCB-793), SCB-794 (https://vanti.youtrack.cloud/issue/SCB-794)) that exposes reboot control and uptime tracking for SC-BOS nodes via gRPC, with a UI for operators on the Cohort card.

image


mattnathan (dismissed):
Proto trait definition needs changes before we review the rest of the implementation. See comment for details.

hexaglow (changes requested):
## Goals

Is this trying to implement SCB-793, SCB-794 or both? They are written to be subtly different:

### SCB-794

> Allow rebooting of nodes via the ops ui

This is a pretty narrow spec, which this PR seems to achieve. If we're going for this use-case, the API can be narrow like this, but would probably fit better as a private Ops UI API under `smartcore.bos.ops.boot` or similar. And you might as well implement the Ops UI feature at the same time as they are two sides of the same coin.

### SCB-794

> A trait that allows you to reboot a device (turn it off, on, etc).

I would read that as a broader thing, allowing you to power a device off, on and reboot it etc. It should be suitable as a general-purpose trait, that drivers could implement for other non-BOS devices. 

## Implementation

`pkg/gentrait` should not exist.

Use of `os.Exit`: this quits the program immediately. It does not wait for other goroutines to stop and does not run any defers. Therefore:

- Buffers/caches are not flushed
- Database transactions may not be completed
- In-progress RPCs stop immediately

Use an alternative implementation that allows for a clean shutdown. Existing clean shutdown scenarios that we already handle:

- Handling of SIGINT (Ctrl-C) using `app.RunUntilInterrupt` to cancel the run context.
- When we need to reboot to install a config update downloaded from the cloud, we use `errgroup`'s automatic context cancellation when a goroutine stops with a (sentinel) error. See `controller.go:629`.

An `os.Exit`-based "force reboot" could be provided, but a clean reboot needs to be the default.

deanredfern26:
have added a `force` flag to allow for both graceful and forced reboots.
this is trying to implement both SCB-793, SCB-794, devices should be able to implement `Reboot` if they support it and it is desirable to expose it.

ui/ops/src/routes/system/components/CohortNodeCard.vue
```diff
@@ -238,17 +241,80 @@
...
+      </template>
     </v-card-text>
   </v-card>
+
+  <v-dialog v-model="showRebootDialog" max-width="400px">
+    <v-card>
+      <v-card-title>Reboot {{ node.name }}?</v-card-title>
+      <v-card-text>
+        <p class="mb-2 text-body-2">
+          The node will exit and be restarted by its supervisor.
```
hexaglow: This will be hard to understand for the end user. We shouldn't expect the user to know or care that it's actually another piece of software that does the restarting - they are interacting with the BOS system here.

ui/ops/src/routes/system/components/CohortNodeCard.vue
```diff
@@ -373,6 +443,58 @@ const utilizationColor = (value) => {
...
+  const d = Math.floor(diff / 86400);
+  const h = Math.floor((diff % 86400) / 3600);
+  const m = Math.floor((diff % 3600) / 60);
+  const s = diff % 60;
+  if (d > 0) return `${d}d ${h}h ${m}m`;
+  if (h > 0) return `${h}h ${m}m ${s}s`;
+  if (m > 0) return `${m}m ${s}s`;
+  return `${s}s`;
+});
+
```
hexaglow, line 482: This is full of magic numbers. Consider:
1. Using `timestampToDate` from `api/convpb.js`
2. Using the duration constants from `util/date.js` instead of magic numbers
3. Factoring the duration calculation into a utility function in `util/date.js` e.g. `decomposeDuration(ms) -> {days, hours, minutes, seconds, milliseconds}`

ui/ops/src/routes/system/components/CohortNodeCard.vue
```diff
@@ -373,6 +443,58 @@ const utilizationColor = (value) => {
...
+      pullBootState(req, bootResource);
+      return () => closeResource(bootResource);
+    }
+);
+const bootState = computed(() => bootResource.value);
+
+const now = ref(Date.now());
+const uptimeInterval = setInterval(() => { now.value = Date.now(); }, 1000);
+onScopeDispose(() => clearInterval(uptimeInterval));
+
```
hexaglow: We have a utility for reactive now.
Use `useNow` from `components/now.js`.

ui/ops/src/routes/system/components/CohortNodeCard.vue
```diff
@@ -238,17 +241,80 @@
...
+          </div>
+          <div v-if="bootState.lastRebootActor?.displayName" class="resource-row">
+            <span class="resource-label">Actor</span>
+            <span class="resource-value text-truncate" style="width: auto"
+                  :title="bootState.lastRebootActor.displayName">
+              {{ bootState.lastRebootActor.displayName }}
+            </span>
+          </div>
+        </div>
+      </template>
```
hexaglow, line 276: The labels for the Reason and Actor aren't super clear. Mostly because the section heading is 'Last Boot'. The uptime is always just time since boot, so that makes sense. But not all boots are reboots etc. UI should make it clear that Reason is 'the reason why BOS most recently rebooted' and Actor is 'who triggered it'.

pkg/system/boot/factory.go
```diff
@@ -0,0 +1,206 @@
...
+
+	// On graceful shutdown, write a clean-exit marker so a planned stop isn't treated as a crash.
+	go func() {
+		<-ctx.Done()
+		mu.Lock()
+		defer mu.Unlock()
+		if !rebootWritten {
+			_ = s.writeStateFile(rebootState{CleanExit: true})
+		}
+	}()
```
hexaglow: There is no synchronisation here to ensure that the file is written before the process exits. This can cause an incorrect detection of an unclean shutdown.

pkg/system/boot/factory.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	s.logger.Info("reboot requested",
+		zap.String("reason", req.Reason),
+		zap.String("actor", req.GetActor().GetDisplayName()),
+	)
+
+	if req.Force || s.requestReboot == nil {
+		go func() {
+			time.Sleep(100 * time.Millisecond)
```
hexaglow, line 143: There should be a comment indicating why there's a sleep here and why 100ms is the right duration. I assume this is to leave enough time for the RPC response before cutting the connection off?

hexaglow (changes requested):
I encountered a UI error when triggering a (normal) reboot:

Image

The reboot reason tracking doesn't integrate with the automatic deployment updates. I would expect to see a reboot reason of "automatic restart to install deployment" or something like that. Also the reboot state file write is currently unreliable.

Idea: Move the state file write to a defer in Controller.Run - this runs (a) when we know we are shutting down, and (b) synchronously (so won't be interrupted). Expand rebootNowError to contain a reason, which can be extracted.

hexaglow:
@deanredfern26 this ready for review?

Edit: oh I see now, I left a comment instead of a review by mistake last time. So it's still waiting for my review. Will look.

pkg/system/boot/factory.go
```diff
@@ -0,0 +1,195 @@
...
+	announcer.Announce(s.nodeName,
+		node.HasServer(bootpb.RegisterBootApiServer, bootpb.BootApiServer(server)),
+		node.HasServer(bootpb.RegisterBootHistoryServer, bootpb.BootHistoryServer(histServer)),
+		node.HasTrait(bootpb.TraitName),
+	)
+
+	// Block until shutdown. The clean-exit state file write is handled by
+	// Controller.Run's defer, which covers both graceful shutdown and deployment
+	// restarts. onReboot handles the Boot RPC path (writing reason + actor).
+	<-ctx.Done()
```
hexaglow: It's not necessary for an `applyConfig` function to block on ctx. Compare `pkg/system/alerts/factory.go`

pkg/system/boot/factory.go
```diff
@@ -0,0 +1,195 @@
+// Package boot implements the Boot trait for the sc-bos process itself.
+// When enabled, it records the process boot time and handles reboot requests by
+// calling os.Exit(0), relying on the process supervisor (e.g. systemd) to restart it.
+package boot
```
hexaglow, line 5: Package comment outdated

hexaglow (changes requested):
Architecture feels right now.

I don't like how there are 2 implementations of the state file writing. Although there are comments indicating they need to be compatible, they should share code. 

Consider exporting the type and a function to write the file from the `boot` package. That does make them public, but they should be pretty stable APIs anyway because it's an on-disk format. But you could create a little internal package for it instead.

hexaglow:
There's something else I noticed looking at this with fresh eyes. I won't ask you to make changes but would like to hear your thoughts:

The `boot` system requires no configuration, and I can't think of a situation where it would be harmful to turn it on. So in this design you need to add `"boot": {}` to your config or it's not available. It could probably be available by default, though. Perhaps it doesn't need to be a system at all.

deanredfern26:
> There's something else I noticed looking at this with fresh eyes. I won't ask you to make changes but would like to hear your thoughts:
> 
> The `boot` system requires no configuration, and I can't think of a situation where it would be harmful to turn it on. So in this design you need to add `"boot": {}` to your config or it's not available. It could probably be available by default, though. Perhaps it doesn't need to be a system at all.

Agree it makes sense for it to be included by default for all nodes. Is it possible that we might add config for it at some point? Something like "cannot reboot between certain times". If we need to manage the config for it via Connect, it might just be easier for it to follow the same pattern as other systems. 

I think having default config file inclusions for area controllers, edge gateways etc is a thing we should have that would make certain features available. Probably would just end up existing when creating nodes in connect is fully formed.

hexaglow (approved):

PR #748, opened 2026-03-20
proto/dataretentionpb: add dataretention trait and wire into controller stores

deanredfern26:
[SCB-1111](https://vanti.youtrack.cloud/issue/SCB-1111)

Add data retention trait                                                                                                                                                                                                                                                                          
                                     
  Adds a new DataRetention BOS trait that exposes storage utilisation stats and administrative operations for the node's configured data backends. The trait is automatically announced on Controller.Run() for each enabled store.                                                                 
                                                                                                                                                                                                                                                                                                    
  What's new                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                    
  Proto / generated code                                                                                                                                                                                                                                                                            
  - proto/smartcore/bos/data_retention/v1/data_retention.proto — defines DataRetentionApi and DataRetentionInfo                                                                                                                                                                                     
  - pkg/proto/dataretentionpb/ — generated Go + UI code, plus Model and ModelServer co-located with the generated files                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                  
  Trait capabilities                                                                                                                                                                                                                                                                                
  - GetDataRetention / PullDataRetention — reads current bytes and item counts (streaming-capable)                                                                                                                                                                                                
  - ClearDataRetention — delete all records                                                                                                                                                                                                                                                         
  - DeleteOldDataRetention — delete records older than a given retention period                                                                                                                                                                                                                   
  - CompactDataRetention — VACUUM / GC / repack (Postgres only)                                                                                                                                                                                                                                     
  - SpringCleanDataRetention — delete old + compact in one RPC (Postgres only)                                                                                                                                                                                                                      
  - DescribeDataRetention — advertises which operations a device supports via bool flags (can_clear, can_delete_old, can_compact, can_spring_clean)                                                                                                                                                 
                                                                                                                                                                                                                                                                                                    
  Backend wiring (pkg/app/stores_data_retention.go)                                                                                                                                                                                                                                                 
  - SQLite history store → announced at {nodeName}/stores/history; supports clear and delete-old; polls disk usage + row count every 30s                                                                                                                                                            
  - Postgres store → announced at {nodeName}/stores/postgres; supports all four operations; row count is cached for 1 hour to avoid expensive COUNT(*) on large tables                                                                                                                              
                                                                                                                                                                                                                                                                                                    
  SQLite additions (pkg/history/sqlitestore/)                                                                                                                                                                                                                                                       
  - TotalCount(ctx) — total row count across all sources                                                                                                                                                                                                                                            
  - Clear(ctx) — deletes all history and history_sources rows in a single transaction                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                    
  Routing                                                                                                                                                                                                                                                                                           
  - Registered in pkg/node/alltraits/services.go so gateway proxying works correctly


mattnathan (dismissed):

hexaglow (changes requested):
This doesn't build currently

deanredfern26:
> This doesn't build currently

oops didn't add the updated test file, is ok now

proto/smartcore/bos/dataretention/v1/data_retention.proto
```diff
@@ -0,0 +1,134 @@
+syntax = "proto3";
+
+package smartcore.bos.data_retention.v1;
```
hexaglow: Use of an underscore in the package name is not consistent with our other gRPC services.

In #809 we added an API under `smartcore.bos.ops.cloud.v1alpha` as a "private API" only for consumption by the Ops UI. The advantage is that it's excluded from compatibility guarantees. If you only forsee this API being used from the Ops UI, consider moving the API there.

proto/smartcore/bos/data_retention/v1/data_retention.proto
```diff
@@ -0,0 +1,134 @@
...
+
+  // Remove all records from this data retention resource.
+  rpc ClearDataRetention(ClearDataRetentionRequest) returns (ClearDataRetentionResponse);
+  // Remove records older than retention_period. retention_period must be positive.
+  rpc DeleteOldDataRetention(DeleteOldDataRetentionRequest) returns (DeleteOldDataRetentionResponse);
+  // Compact/optimise storage (GC, VACUUM, repack). No-op if not supported.
+  rpc CompactDataRetention(CompactDataRetentionRequest) returns (CompactDataRetentionResponse);
+  // Delete old records and compact storage in one operation (spring clean).
+  rpc SpringCleanDataRetention(SpringCleanDataRetentionRequest) returns (SpringCleanDataRetentionResponse);
+}
```
hexaglow: 1. ClearDataRetention and DeleteOldDataRetention could be 1 RPC with an optional parameter for the cutoff date. They are both the same operation fundamentally.
2. Awkward naming `DeleteOldDataRetention` etc. It's not the data retention being deleted (what would that mean?), it's the data. "spring clean" is a bit informal too.
3. Is it necessary to have SpringCleanDataRetention as opposed to just having the client call Delete and Compact in sequence?
4. For operations that could not be completed because the target resource does not support them, prefer to return `UNIMPLEMENTED` or similar. The client can decide if it wants to ignore the error or not (e.g. show message to the user).

proto/smartcore/bos/data_retention/v1/data_retention.proto
```diff
@@ -0,0 +1,134 @@
...
+
+// DataRetentionBytes describes storage consumption measured in bytes.
+message DataRetentionBytes {
+  // Bytes currently in use.
+  optional uint64 used = 1;
+  // Total capacity in bytes. Omit if unlimited or unknown.
+  optional uint64 capacity = 2;
+  // Utilization as a percentage (0–100).
+  optional float utilization = 3;
```
hexaglow: Is it necessary to include this in the API response when it can be trivially calculated by the client from the above fields? (same below in DataRetentionItems)

pkg/proto/dataretentionpb/model_server.go
```diff
@@ -0,0 +1,153 @@
...
+	UnimplementedDataRetentionInfoServer
+
+	model              *Model
+	clearHandler       ClearHandler
+	deleteOldHandler   DeleteOldHandler
+	compactHandler     CompactHandler
+	springCleanHandler SpringCleanHandler
+	support            *DataRetentionSupport
+
+	subscribers atomic.Int32
```
hexaglow: This is never actually used - it is incremented and decremented in PullDataRetention but nothing reads its value.

pkg/app/stores_data_retention.go
```diff
@@ -0,0 +1,325 @@
...
+		retention.Bytes = &dataretentionpb.DataRetentionBytes{Used: &used}
+	}
+
+	// COUNT(*) is expensive; use the cached value if still fresh.
+	if n, ok := cache.get(); ok {
+		used := uint64(n)
+		retention.Items = &dataretentionpb.DataRetentionItems{Used: &used}
+	} else {
+		var n int64
+		if err = r.QueryRow(ctx, "SELECT COUNT(*) FROM history").Scan(&n); err != nil {
```
hexaglow: `COUNT(*)` is very slow in Postgres! It performs a full table scan. See https://stackoverflow.com/a/7945274. Performing this every 30 seconds is a bad idea.

pkg/proto/dataretentionpb/model_server.go
```diff
@@ -0,0 +1,153 @@
...
+
+// DeleteOldHandler is called by ModelServer to delete records older than the given retention period.
+type DeleteOldHandler func(ctx context.Context, req *DeleteOldDataRetentionRequest) (*DeleteOldDataRetentionResponse, error)
+
+// CompactHandler is called by ModelServer to compact/optimise storage.
+type CompactHandler func(ctx context.Context, req *CompactDataRetentionRequest) (*CompactDataRetentionResponse, error)
+
+// SpringCleanHandler is called by ModelServer to delete old records and compact storage in one operation.
+type SpringCleanHandler func(ctx context.Context, req *SpringCleanDataRetentionRequest) (*SpringCleanDataRetentionResponse, error)
+
```
hexaglow: Do these handlers need to accept / return gRPC types? Introduces coupling between database layer and RPC layer.

These would make more sense as methods on an interface. I can't see a case where you'd have a ModelServer that sent a Clear request to one place and a Compact request somewhere totally different - they will be directed at the same DB in the end.

hexaglow (changes requested):
The `app` package isn't the right place for implementing database functionality e.g. SQL statements. It is for initialisation and gluing together functionality implemented elsewhere.

Performance: Polling the size of the databases every 30s may be problematic for large databases containing millions of rows. In particular `COUNT(*)` in Postgres is known to be slow. Please perform some benchmarks on large tables to demonstrate that this won't drag down the performance of BOS. Consider:

1. If PullDataRetention is even necessary to implement for the intended use case
2. If polling can be supressed while nobody is fetching the data

pkg/app/stores_data_retention.go
deanredfern26: The query result does get cached, so it should only be getting refreshed once an hour, but to be fair that's probably still too frequent for what we need. I'll increase the interval and work through the other comments

deanredfern26: I have replaced `COUNT` with `n_live_tup`, it looks like it is sufficient for what we are trying to achieve

pkg/proto/dataretentionpb/data_retention.go
```diff
@@ -0,0 +1,5 @@
+package dataretentionpb
+
+import "github.com/smart-core-os/sc-bos/pkg/trait"
+
+const TraitName trait.Name = "smartcore.bos.data_retention.DataRetention"
```
hexaglow: Hasn't been updated to reflect the new path

pkg/system/dataretention/sqlite.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/smart-core-os/sc-bos/pkg/app/stores"
+	"github.com/smart-core-os/sc-bos/pkg/node"
+	"github.com/smart-core-os/sc-bos/pkg/proto/dataretentionpb"
+)
+
+func announceSqlite(ctx context.Context, n *node.Node, name string, s *stores.Stores, dataDir string, logger *zap.Logger) node.Undo {
+	model := dataretentionpb.NewModel()
+	server := dataretentionpb.NewModelServer(model, &sqliteBackend{stores: s},
+		dataretentionpb.WithItemName("record"),
```
hexaglow: Is there a reason this is called 'record' and the Postgres one is called 'row'?

hexaglow (changes requested):
New measurement approach and RPC structure seems appropriate.

`pkg/system/dataretention` is not the right package path, because `dataretention` is not a system (does not provide a `system.Factory`).

(the below aren't new but I didn't spot it the first time, sorry)

There's a layering problem with the postgres handling. `pgxstore.Store` is intended to represent the storage for a single source (it's provided as a construction parameter). However, the new methods `TotalSize`, `ApproxCount`, `Delete` and `Vacuum` ignore this and operate table-wide, across all sources. That's surprising to me - especially that calling `Delete` on a store which has a source value could delete data from other sources - breaks the abstraction. There's a lot of dummy source values (`pgxstore.NewStoreFromPool("", ...)` introduced, suggestsingthat these methods don't belong on `Store`. They could be free functions taking a pool instead. 

Refactoring opportunity (optional): `announceSqlite` and `announcePostgres` are very similar, differing slightly in initialisation and which update function is called. They could be combined by e.g. passing in a (package-private) interface which could fetch and return the statistics.

(edit) Currently the API uses the default trait permissions, so anyone with trait:write permission (e.g. Operator) can call all these APIs, including deleting data. This needs a Rego file to restrict destructive operations to admin.

hexaglow (approved):

PR #750, opened 2026-03-20
feat/audit logs

deanredfern26:
[SCB-422](https://vanti.youtrack.cloud/issue/SCB-422) Audit logging for write operations
                                                                                                                                                                                                                                                                                                    
Adds an audit log that records every mutating gRPC and HTTP request processed by a controller — capturing who made the request, what they did, and whether it was allowed or denied. Audit entries are written to a rotating JSON file and exposed via the existing Log trait.                                                                                                                                                                                                                                                        
 
For each write operation (any gRPC method not prefixed with Get, Pull, Describe, or List; any HTTP method other than GET/HEAD/OPTIONS):                                                                                                                                                           
                  
  - Outcome: allowed or denied                                                                                                                                                                                                                                                                      
  - gRPC service and method name (or HTTP path and method)
  - Peer address                                                                                                                                                                                                                                                                                    
  - Token subject and display name (from Keycloak preferred_username or file account name)                                                                                                                                                                                                          
  - Whether a client certificate was presented and its subject
  
  Configuration                                                                                                                                                                                                                                                                                     
                  
  Audit logging is opt-in via the new auditLog key in system.conf.json:                                                                                                                                                                                                                             
   
  "auditLog": {                                                                                                                                                                                                                                                                                     
    "filename": ".data/my-controller/audit.log",
    "maxSizeMB": 50,                                                                                                                                                                                                                                                                                
    "maxAgeDays": 90,
    "maxBackups": 10,                                                                                                                                                                                                                                                                               
    "compress": true                                                                                                                                                                                                                                                                                
  }
                                                                                                                                                                                                                                                                                                    
  When auditLog is present, the controller announces an audit-log device (e.g. my-controller/audit-log) with the Log trait, making audit log metadata and signed download URLs available via the standard Log API.


pkg/system/log/service.go
```diff
@@ -118,6 +119,137 @@ func (s *System) applyConfig(ctx context.Context, cfg config.Root) error {
 	return nil
 }
 
+// --------------------------------------------------------------------------
+// Audit file logger construction
+// --------------------------------------------------------------------------
+
+// AuditSetup holds an initialised audit log subsystem for use by Bootstrap.
+// Create it with NewAuditSetup; always call Close on shutdown.
+type AuditSetup struct {
```
hexaglow: Although they are both log related, `AuditSetup` isn't part of the log *system*, so I'm not sure it belongs in this package. Bear in mind this is in a public package.

pkg/auth/policy/interceptor.go
```diff
@@ -185,6 +190,17 @@ func (i *Interceptor) checkPolicyGrpc(ctx context.Context, creds *verifiedCreds,
 			zap.Strings("queries", queries),
 		)
 	}
+	// Only audit once per RPC, not for every message on an open client/bidirectional stream.
+	if isWriteMethod(method) && !stream.Open {
+		outcome := "allowed"
+		if err != nil {
+			outcome = "denied"
+		}
+		i.writeAuditEntry(outcome, creds, addr,
```
hexaglow, line 233: This is called synchronously before the actual gRPC handler runs, and because it uses a file-based logger, that involves a file write syscall for every RPC call we log. Looks like it will slow down RPC handling. Could we log asynchronously via a buffered channel instead?

pkg/system/log/service.go
```diff
@@ -118,6 +119,137 @@ func (s *System) applyConfig(ctx context.Context, cfg config.Root) error {
...
+	// Logger writes structured JSON audit entries to a rotating file.
+	// Nil when no filename was configured.
+	Logger *zap.Logger
+	// Model holds the in-memory ring buffer of audit entries. Always non-nil.
+	Model *logpb.Model
+
```
hexaglow: Every user of AuditSetup is supposed to write exactly the same messages to the Logger and to the Model, otherwise they end up inconsistent. This isn't documented, and is a fragile invariant. 

A different design which made sure that the same logs always went to both would be preferable.

pkg/app/controller.go
```diff
@@ -454,6 +502,17 @@ func Bootstrap(ctx context.Context, config sysconf.Config) (*Controller, error)
...
 	c.Defer(closeHealthStore)
+	if cloudDataRoot != nil {
+		c.Defer(cloudDataRoot.Close)
+	}
+	if auditSetup != nil {
+		c.Defer(func() error {
+			if auditSetup.Logger != nil {
+				_ = auditSetup.Logger.Sync()
+			}
+			return auditSetup.Close()
```
hexaglow: This should happen inside AuditSetup.Close

pkg/system/log/service.go
```diff
@@ -118,6 +119,137 @@ func (s *System) applyConfig(ctx context.Context, cfg config.Root) error {
...
+	a.Logger = logger
+	a.closer = closer
+	a.downloadKey = newHMACKey()
+	a.filename = filename
+	return a, nil
+}
+
+// RegisterHTTP installs the audit-log file download handler on mux at dlPath.
+// Does nothing when the AuditSetup has no file backing.
+func (a *AuditSetup) RegisterHTTP(mux *http.ServeMux, dlPath string) {
```
hexaglow: The dependency on `http.ServeMux` is unnecessary I feel, as AuditSetup doesn't actually care about what path it runs from. Could do

```go
func (a *AuditSetup) HTTPHandler() http.Handler { ... }
```

instead and let the caller worry about how routing works. 

I know that `devices.Server` uses this pattern, but most of the mux registrations in `app/controller.go` do not.

pkg/app/controller.go
```diff
@@ -310,11 +313,34 @@ func Bootstrap(ctx context.Context, config sysconf.Config) (*Controller, error)
...
+		}
+	}
+	if pol != nil || auditSetup != nil {
+		if pol == nil {
+			// No access policy — inject allow-all so the interceptor runs for audit logging only.
+			logger.Warn("no access policy configured; running interceptor for audit logging only (all requests allowed)")
+			pol = policy.Func(func(_ context.Context, _ string, _ policy.Attributes) (rego.ResultSet, error) {
+				return rego.ResultSet{{Expressions: []*rego.ExpressionValue{{Value: true, Text: "allow"}}}}, nil
+			})
+		}
```
hexaglow: This could be added to the `policy` package as e.g. `AllowAll`

hexaglow (changes requested):
Couple of design elements stand out to me as not ideal:

1. Logs happen synchronously with RPC processing, so the RPC handler can't start until the OS has handled the log write, slowing down RPC processing

2. The same data is stored in the in-memory buffer, and on disk, and it's actually the caller who's responsible for making sure they are the same. Suggested alteratives: (a) store only on disk and read back in when required; or (b) encapsulate this so that they will always be in sync

`app.Bootstrap` has expanded yet again, it's becoming unwieldy. Not something that necessarily needs addressing in this PR but I think we should refactor this.

deanredfern26:
changes implemented and a ticket made to refactor `app.Bootstrap` [SCB-1355](https://vanti.youtrack.cloud/issue/SCB-1355)

pkg/auth/policy/interceptor.go
```diff
@@ -334,62 +364,75 @@ type auditField struct {
...
-			zap.String("peer", addr),
-			zap.String("subject", subject),
-			zap.String("name", name),
-			zap.Bool("cert", creds.certValid),
-			zap.String("certSubject", certSubjectStr),
-			zap.Bool("token", hasToken),
-		)
-		i.auditLogger.Info("write", fields...)
 	}
+	// Non-blocking send; drop if worker can't keep up rather than slowing RPCs.
```
hexaglow: I think a blocking send would be preferable here, so we don't lose data. In very high load scenarios, this would indeed block. But an audit system is significantly less useful if there might be gaps in the data. The buffering means that in the vast majority of calls there is no blocking. So I think a blocking send into a large buffer is the sweet spot.

In fact, it wouldn't actually be slower because `i.logger.Warn` writes to the console which would take the same time or longer than waiting for a free slot in the channel.

You could consider plumbing the request context through to here, so that if the request is cancelled we can give up the send without blocking. But that would be a very rare scenario so it's not really necessary.

hexaglow (changes requested):
New structure is better 👍 

Does `pkg/app/audit` need to be used by projects dependent on sc-bos? As far as I can tell, only sc-bos needs it. Therefore it should be an internal package e.g. `internal/audit` or `pkg/app/internal/audit`. Avoid increasing the API surface area unnecessarily.

hexaglow (approved):

PR #795, opened 2026-04-09
scfix: remove fixers used for completed migrations

hexaglow:
The `cmd/tools/scfix` tool had a couple of single-purpose fixers that only refactored files inside this repo. That refactoring work has been completed, so the fixers are now redundant. Remove them.


hexaglow:
@mattnathan bump

mattnathan (approved):

PR #798, opened 2026-04-13
Fix flaky Go tests causing spurious CI failures

hexaglow:
There's 2 tests that sometimes fail, causing spurious Go Test fails in Github Actions and locally:
- In `auto/history` - `Test_automation_applyConfig` occasionally failed with an incorrect number of detected changes. Make this deterministic.
- In `internal/router` - `TestStreamHandler/downstream` - fix tests failing with a timeout on some platforms. This was due to accidentally invoking gRPC's name resolution when we're using a fake in-memory dialler.

### DIFF
--- a/internal/router/stream_test.go
+++ b/internal/router/stream_test.go
@@ -282,7 +282,7 @@ func nodeServer(name string) *grpc.Server {
 }
 
 func bufConn(buf *bufconn.Listener) (*grpc.ClientConn, error) {
-	return grpc.NewClient("localhost:0",
+	return grpc.NewClient("passthrough://localhost:0",
 		grpc.WithContextDialer(func(context.Context, string) (net.Conn, error) {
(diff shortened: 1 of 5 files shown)


dloveen-v (approved):

deanredfern26:
I redid the action and there is another failing test, can you have a look at this as well please?
```

--- FAIL: TestApi_PullServices (0.01s)
    api_test.go:69: Receive: Unexpected error context deadline exceeded
FAIL
FAIL	github.com/smart-core-os/sc-bos/pkg/task/serviceapi	0.036s

```

hexaglow:
@deanredfern26 I couldn't reproduce the problem locally, but looking at the test, it's another one with a context timeout that could fail if the runner is slow or pauses for a couple of seconds. I've converted it to use `synctest`.

PR #799, opened 2026-04-13
Enable UI linting in CI

hexaglow:
Automatically lint UI packages when modified. I've combined this with the dependency checking job.

Remove keycloak-login-prototype - it's an old expermient that's no longer relevant.


.github/workflows/ui-lint.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Install dependencies
+        run: yarn --frozen-lockfile
+        working-directory: ui
+
+      - name: Run lint
+        working-directory: ui
+        run: |
+          for dir in $(jq -r '.workspaces[]' package.json); do
+            echo "Linting $dir..."
+            yarn --cwd "$dir" lint:nofix
```
dloveen-v, line 50: Does this need forwarding and parsing for errors?

hexaglow: I'm not sure what you mean? 
If the lint fails, the action run fails.

dloveen-v (approved):

PR #806, opened 2026-04-16
feat/space-ui-updates

deanredfern26:
Adds some updates to the Space UI:
- Update colour scheme and logo to match latest style guidelines
- Add a 'combined' mode, to support switching a panel between single room and joined mode
- Add 'admin' mode, to allow control of any zone without being 'locked' into one room
- Add zone filter to filter out selectable zones based on account 'zones'
- Add zone name under date time
- Update the README to instruct what the new modes do and general info on the Space UI

image


ui/space/README.md
```diff
@@ -1,5 +1,263 @@
...
+When two adjacent rooms are booked together, a "combined room" zone can be configured during setup. The home screen shows a toggle button to switch between the two views.
+
+- **Single view** — shows controls for the panel's own room.
+- **Combined view** — shows controls for the joined zone.
+
+The toggle persists across reloads. If a combined room was not configured during setup, the toggle is not shown.
+
+### Admin Mode
+
+Admin mode is for a roving panel or shared tablet that should not be locked to a single room. Instead of saving a zone, the panel shows the full zone list on every visit. The admin selects a zone, adjusts controls, and can return to the list without anything being permanently saved.
```
dloveen-v, line 222: typo `roving`

dloveen-v (commented):
Code LGTM from initial pass. Have you checked the login flow as restarting edge gateways sometimes logs UIs out which isn't ideal from a maintenance perspective for this high profile UI

ui/space/src/routes/components/AirTemperatureCard.vue
dloveen-v, line 87: Should current Temp fall back to pulling `Temperature` instead of `airTemperature`?

deanredfern26:
> Code LGTM from initial pass. Have you checked the login flow as restarting edge gateways sometimes logs UIs out which isn't ideal from a maintenance perspective for this high profile UI

Good call, restarting the edge gateways does kick the panel out. We can't really have that behaviour, I have added a auth retry which stores the creds in browser storage and then re-attempts. It works and the panel just carries on, but not the most secure, what do you think?

ui/space/src/routes/components/AirTemperatureCard.vue
deanredfern26: These aren't really the same thing I don't think, I think has to be airTemperature or nothing, although I've never really understood why these are 2 distinct traits.

dloveen-v: airTemperature is for control, whereas temperature is for sensing, I believe.

IMO the read value should be equivalent which ever trait is used. Then again sensors aren't always calibrated the same either. It also means that zone name has to support at least 1 of the temp traits, which the changes is checking for currently.

 It's up to you if you want the widget to support both.

dloveen-v:
> > Code LGTM from initial pass. Have you checked the login flow as restarting edge gateways sometimes logs UIs out which isn't ideal from a maintenance perspective for this high profile UI
> 
> Good call, restarting the edge gateways does kick the panel out. We can't really have that behaviour, I have added a auth retry which stores the creds in browser storage and then re-attempts. It works and the panel just carries on, but not the most secure, what do you think?

Not the most secure at all as localStorage is viewable. But I would be very surprised if someone goes through the effort of connecting the Android panel to a computer to read the browser localStorage.

dloveen-v (approved):

dloveen-v:
As an after thought, should `Switch to combine room` be a dynamic text string reading `Combine rooms x and y` instead?

deanredfern26:
> As an after thought, should `Switch to combine room` be a dynamic text string reading `Combine rooms x and y` instead?

It is a little nicer, it would require a 3 way step on the config so each room knows itself, the other zone and the combined zone. Maybe this one is a future addition as there is already a fair bit being added here and I am deploying this tomorrow :D

deanredfern26:
> > > Code LGTM from initial pass. Have you checked the login flow as restarting edge gateways sometimes logs UIs out which isn't ideal from a maintenance perspective for this high profile UI
> > 
> > 
> > Good call, restarting the edge gateways does kick the panel out. We can't really have that behaviour, I have added a auth retry which stores the creds in browser storage and then re-attempts. It works and the panel just carries on, but not the most secure, what do you think?
> 
> Not the most secure at all as localStorage is viewable. But I would be very surprised if someone goes through the effort of connecting the Android panel to a computer to read the browser localStorage.

Yeah I know it is not secure, but as the panels do not have a usb port on them and they are network gapped, it is not too bad for the immediate use case. I have added a note in the README that the local account authentication should only be used on trusted devices

PR #808, opened 2026-04-21
feat/widget update

deanredfern26:
allow hiding of generated metric on MeterDensityCard
Add floor name to FloorTraitCells
Adjust PeopleCountCard to fit nicely into 2 height


ui/ops/src/dynamic/widgets/meter/MeterDensityCard.vue
```diff
@@ -163,6 +164,10 @@ const props = defineProps({
     type: Boolean,
     default: true
   },
+  showGenerated: {
+    type: Boolean,
+    default: false
```
dloveen-v: Nit: can this be default to true, at all? It'll save having to change the ui-config.json in more repositories.

dloveen-v (approved):

PR #809, opened 2026-04-21
Allow enrollment with the cloud using a code

hexaglow:
In situations where copy-paste of long client details is not easy, we can use a short, time limited code to allow credential exchange. 

- Add a new `/register` endpoint to cloudsim, which returns the API root and credentials.
- Add a private gRPC API `smartcore.bos.ops.cloud.v1alpha` which is only for use in the Ops UI and not subject to any compatibility guarantees version-to-version.
- Refactor the `cloud` package to expose `Conn` as the main type - unlike `DeploymentUpdater`, this can support changing registration while in operation. I have also taken the opportunity to better separate concerns in the cloud package, with a new separate `DeploymentStore` type to handle the filesystem.
- Randomise the first poll time on boot to prevent request spikes, but always poll immediately when the registration changes.
- Remove client config from sysconf - details are to be stored in the data directory instead.

The UI can be accessed on the System > Components page, and consists of a modal accessed by the context menu on a node card. This allows connecting an unregistered node to the cloud, or unlinking an existing node. This all applies without a node reboot.

Deployment installation behaviour remains unchanged.

Please note that hooking up `sccexporter` to the new cloud registration is not part of this PR - it will need to be in a follow-up.


internal/cloud/autopoll.go
```diff
@@ -0,0 +1,76 @@
...
+			ticker.Reset(interval)
+			// update immediately
+		case <-ticker.C:
+		case <-initialTick:
+			// reset ticker phase
+			ticker.Reset(interval)
+		}
+
+		logger.Debug("checking for deployment updates",
+			zap.String("bosapi", cur.BosapiRoot),
```
mattnathan, line 61: 🔴 **Critical — nil-pointer panic**

`cur` is dereferenced here but can be nil. On boot with no registration on disk, `initialTick` stays nil (per the `if cur != nil` guard above), we fall through to the `<-ticker.C` branch after `interval`, and panic reading `cur.BosapiRoot`.

Reproduced against `example/config/vanti-ugs` (pollInterval 10s):

```
panic: runtime error: invalid memory address or nil pointer dereference
  at internal/cloud/autopoll.go:57
```

With the default 5 min interval, the crash is silent for longer but still deterministic — any node with `cloud:` configured but no registration eventually crashes.

Simplest fix is `if cur == nil { continue }` before the debug log. The `ErrNotRegistered` branch from `conn.Update` below covers the unlink-races-with-tick case, so it's worth keeping too.

mattnathan (commented):
End-of-review summary — full review in inline comments. Grouped as critical / improvements / nits in the comment bodies themselves.

**Non-inline notes:**

- Commit message for 1b1536948 says "make the first poll happen between 30 seconds and 1 minute after boot (randomised) if the ticker hasn't already fired by that point", but `autopoll.go` actually uses `rand.Int64N(interval.Nanoseconds())` — uniform over `[0, interval)`, which is 0–5 min at the default. The docstring on `AutoPoll` matches the code; only the commit message is wrong. Worth a fixup if you amend/rebase.

**UI / visual feedback (not inline):**

- `LinkCloudDialog` spacing: the `v-expansion-panels` "Advanced" header and the nested Register URL field don't line up with the text fields above. Nested padding on `v-card-text` + panel + panel-text. Either pull flush with `mx-n4`, move the accordion outside `v-card-text`, or simplify to a plain "Advanced settings ▾" toggle — only one input inside.
- "Manage Cloud Link" label overstates what's currently a config readout. Given the proto carries `state`, `last_check_in_time`, and `last_error`, the dialog could include a current-state pill and a "Last checked in Xm ago / Never" line to earn the label. Today "View cloud link" would be more honest.
- "Manage Cloud Link" button ordering: UNLINK bottom-left, CLOSE bottom-right, inverted vs. the Link to Cloud dialog's CANCEL/LINK. Bottom-right is where the eye expects the primary action. Options: swap (CLOSE left, UNLINK right in red), or drop CLOSE entirely and rely on ESC / click-outside / the top-right ✕.
- Unlink has no confirmation and no toast. A single misplaced click removes the link silently; re-enrolling needs an operator to generate a fresh code. Options: confirm step, snackbar, or hide destructive actions behind an inner kebab (three-dots → Unlink) so the primary dialog surface has no destructive affordance. The last is arguably overkill for one action but worth noting.
- `lastError` in the Manage dialog renders as a single-line `v-list-item-title` and truncates with an ellipsis partway through a URL. The text itself is a Go wrapped error (`check-in: send request: Post "…": …`) which isn't user-friendly. Two fixes: render multi-line, and classify errors server-side so the UI can show a clean label with the raw text on expand/tooltip.
- No "pending" state: after Register succeeds, and on bos reboot with credentials on disk, `ConnState` starts as `Disconnected` (fresh creds, no check-in yet) before AutoPoll's first successful poll completes. On a real-world network that'd flash a spurious "cloud offline" chip. A distinct "connecting…" state, or simply not rendering the orange chip until there's been one round-trip, would avoid it.
- Register URL discoverability: if the system config supplies a default, the user has no indication of that inside the dialog — they'd need to expand "Advanced" just to check. If no default is set, the user gets an HTTP error (see the inline comment on `opsapi/server.go`). A hint line under the Enrollment code input ("Register URL: <default> — override in Advanced") would cover both.

internal/opsapi/server.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		reg, err = cloud.Register(ctx, ec.GetCode(), regUrl, s.nodeName)
+		if err != nil {
+			if cloud.IsInvalidEnrollmentCode(err) {
+				// UI treats this as a special value with a better error message, as this is the most common error case.
+				return nil, status.Error(codes.PermissionDenied, "invalid_enrollment_code")
+			}
+			return nil, status.Errorf(codes.Internal, "register: %v", err)
+		}
+	case *cloudpb.RegisterCloudConnectionRequest_Manual:
```
mattnathan, line 106: 🔴 **Critical — `Manual` creds never validated**

We save whatever the user typed and return success. If the credentials are wrong, the user sees the dialog close (success UX) and then the card chip flips to `cloud offline` a moment later once AutoPoll's first check-in fails — with a developer-grade error buried in the Manage dialog.

The EnrollmentCode branch above already does a synchronous HTTP exchange and returns a friendly `invalid_enrollment_code` error.

Worth framing this as a "test connection" concept rather than bolting validation onto register specifically:

- On manual register, run a sync check-in before returning — "Linked" on success, `codes.Unauthenticated` / `codes.Unavailable` on failure with a friendly mapping.
- Expose the same primitive from the Manage dialog as a "Test connection" button, so operators can re-verify after a credential rotation or when debugging.

The enrollment-code branch gets test-connection semantics implicitly via the sync HTTP exchange it already does, so the two flows end up with matching behaviour.

ui/ops/src/routes/system/components/CohortNodeCard.vue
```diff
@@ -34,13 +34,22 @@
...
+              cloud offline
+            </v-chip>
           </div>
           <div class="text-caption text-medium-emphasis d-flex align-center mt-1">
             <v-icon size="11" class="mr-1">mdi-network</v-icon>
             {{ node.grpcAddress }}
           </div>
         </div>
-        <v-menu v-if="node.role !== NodeRole.INDEPENDENT" min-width="175px">
+        <v-menu min-width="175px">
```
mattnathan, line 56: 🔴 **Critical — menu now shown on independent nodes**

Removing the `v-if="node.role !== NodeRole.INDEPENDENT"` guard was necessary to surface "Link to Cloud" on independent nodes, but it also exposes the existing "View Certificate", "Download Logs" and "View Live Logs" items there for the first time. **View Certificate fails** when invoked from an independent node.

Either re-gate that specific item (`v-if="node.role !== NodeRole.INDEPENDENT"` on the `<v-list-item>`), or update the certificate viewer to handle the independent-node case. Download / View Logs appear to work fine and are probably safe to leave exposed.

internal/cloud/conn.go
```diff
@@ -0,0 +1,218 @@
...
+// The check-in outcome is recorded internally so that State/PullState reflect it.
+func (c *Conn) Update(ctx context.Context) (needReboot bool, err error) {
+	c.mu.Lock()
+	u := c.updater
+	reg := c.state.Registration
+	c.mu.Unlock()
+	if u == nil {
+		return false, ErrNotRegistered
+	}
+	needReboot, err = u.Update(ctx)
```
mattnathan, line 219: 🔴 **Critical — `Update` can race with `Unlink`**

Between `c.mu.Unlock()` above and `u.Update(ctx)` returning here, `Unlink` can land, clearing the registration and swapping out the updater. The in-flight `u.Update` keeps going with the now-disowned credentials and can write an installing deployment into the shared `DeploymentStore`. `recordCheckIn` correctly discards the state broadcast via the `reg.ClientID` check, but the filesystem side-effects are already done.

Low-probability in practice — needs an unlink mid-download — but worth either:

- passing a per-registration `context.Context` into the updater and cancelling it on `Unlink`, or
- serialising the two via a longer-held lock.

The first keeps `Update` responsive and is probably the least intrusive.

internal/opsapi/server.go
```diff
@@ -0,0 +1,203 @@
...
+		// validate required fields
+		if ec.GetCode() == "" {
+			return nil, requiredFieldError("code")
+		}
+
+		regUrl := s.defaultRegisterURL
+		if ec.GetRegisterUrl() != "" {
+			regUrl = ec.GetRegisterUrl()
+		}
+		reg, err = cloud.Register(ctx, ec.GetCode(), regUrl, s.nodeName)
```
mattnathan, line 102: 🟡 **Improvement — empty register URL falls through to a generic HTTP error**

If neither `Cloud.RegisterURL` is configured nor `ec.GetRegisterUrl()` is supplied, `regUrl` ends up as `""`, which `http.NewRequestWithContext` rejects with `http: no Host in request URL`. The user sees that through three layers of `fmt.Errorf` wrapping.

Worth validating up front — if `regUrl == ""`, return `codes.FailedPrecondition` with something like `"no register URL: supply one in the request or set cloud.registerUrl in system config"`. Matches the style of the existing `code`-missing check.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,199 @@
...
+  cloudConnection: {type: Object, default: () => null}
+});
+const emit = defineEmits(['linked', 'unlinked']);
+
+const dialog = defineModel({type: Boolean, default: false});
+
+const tab = ref('code');
+
+// Enrollment code tab
+const enrollmentCode = ref('');
```
mattnathan, line 181: 🟡 **Improvement — inputs & error alert not reset on close**

`v-dialog` hides rather than destroys the child, so these `ref('')` initialisers only run once. The reset block inside `doRegister` only fires on successful submit — Cancel, ESC, and click-outside all leave the entered code (and any `registerTracker.error` alert) sitting there for the next open.

A single `watch(dialog, open => { if (open) { /* reset refs + clear tracker errors */ } })` handles both the stale-input and stale-error cases.

proto/smartcore/bos/ops/cloud/v1alpha/cloud_connection.proto
```diff
@@ -0,0 +1,92 @@
...
+  string bosapi_root = 4;
+  google.protobuf.Timestamp last_check_in_time = 5;
+  string last_error = 6;
+
+  enum State {
+    STATE_UNSPECIFIED = 0;
+    UNCONFIGURED = 1;
+    CONNECTED = 2;
+    DISCONNECTED = 3;
+    ERROR = 4;
```
mattnathan: 🟡 **Improvement — `ERROR` state is dead**

Nothing on the Go side emits `ERROR`. `Conn.Connectivity` only has `Unconfigured / Connected / Disconnected`, and `cloudStateToProto` can only map to those three. The UI's `cloudLinked` check in `CohortNodeCard.vue` includes `ERROR`, so the code path is wired but never hit.

Two options, both fine:

- Drop the enum value (we're in `v1alpha`, so a breaking change is cheap here).
- Start distinguishing persistent failures (bad credentials, 4xx from SCC) from transient blips and emit `ERROR` for the former. That'd also give the UI something to render differently — e.g. red chip + "Credentials invalid" for `ERROR` vs orange chip + "Cloud offline" for `DISCONNECTED`.

internal/cloud/types.go
```diff
@@ -77,3 +78,11 @@ type APIError struct {
 func (e *APIError) Error() string {
 	return fmt.Sprintf("HTTP %d: %s: %s", e.StatusCode, e.Code, e.Message)
 }
+
+func IsInvalidEnrollmentCode(err error) bool {
```
mattnathan, line 118: 🟡 **Improvement — `IsInvalidEnrollmentCode` contract isn't tested**

The `"invalid_enrollment_code"` string is the contract between whatever sends the error (sim today, real SCC in prod) and the UI's friendly-error mapping in `LinkCloudDialog.vue`. If the server ever renames this code, the friendly message silently degrades and nothing fails loudly.

A one-liner test against the sim that calls `cloud.Register` with a bad code and asserts `IsInvalidEnrollmentCode(err) == true` would pin it. Belongs next to the existing `TestRegister/invalid enrollment code returns 401` in `register_test.go`.

internal/cloud/autopoll.go
```diff
@@ -0,0 +1,76 @@
...
+// to spread check-in load across many nodes starting simultaneously.
+func AutoPoll(ctx context.Context, conn *Conn, interval time.Duration, logger *zap.Logger) bool {
+	initial, changes := conn.PullState(ctx)
+	changes = concurrent.BreakBackpressure(changes) // only care about the latest value, drop others
+	cur := initial.Registration
+
+	var initialTick <-chan time.Time
+	if cur != nil {
+		ns := rand.Int64N(interval.Nanoseconds())
+		initialTick = time.After(time.Duration(ns))
```
mattnathan, line 30: 🟡 **Improvement — `time.After` can't be stopped**

`time.After` returns a Timer that runs to completion regardless of whether the select ever reads from it — stays in the runtime heap until the channel is drained. Not a leak of any consequence here (one timer per AutoPoll call), but stylistically:

```go
timer := time.NewTimer(time.Duration(ns))
defer timer.Stop()
initialTick = timer.C
```

lets the ctx-cancel and registration-change paths clean up properly.

internal/cloud/register.go
```diff
@@ -0,0 +1,72 @@
+package cloud
```
mattnathan, line 1: 🟢 **Nit — terminology overload vs. `pkg/manage/enrollment`**

We already have `pkg/manage/enrollment` for cohort enrollment (Area Controller ↔ Building Controller cert binding). This PR's "enrollment code" is unrelated — SCC registration.

A one-liner at the top of `register.go` noting the distinction would save future devs a few minutes of grep.

internal/opsapi/server.go
```diff
@@ -0,0 +1,203 @@
...
+	if err := s.validateName(req.Name); err != nil {
+		return err
+	}
+
+	send := func(v cloud.ConnState) error {
+		return stream.Send(&cloudpb.PullCloudConnectionResponse{
+			Changes: []*cloudpb.PullCloudConnectionResponse_Change{{
+				Name:            s.nodeName,
+				Type:            typespb.ChangeType_UPDATE,
+				ChangeTime:      timestamppb.New(time.Now()),
```
mattnathan: 🟢 **Nit — `ChangeTime` stamped with `time.Now()` per send**

Each time we fan out the same `ConnState` to a new subscriber, `ChangeTime` is `time.Now()` rather than the actual transition time. Two subscribers receiving the same underlying transition see different timestamps. Not a bug today, but if anything downstream records these it'll be mildly confusing.

Fix is to thread a transition timestamp through `ConnState` (set in `recordCheckIn` / `setRegistration`) and use that here.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,199 @@
...
+              <v-tabs-window-item value="manual">
+                <v-text-field
+                    v-model="clientId"
+                    label="Client ID"
+                    variant="filled"
+                    class="mb-3 mt-1"/>
+                <v-text-field
+                    v-model="clientSecret"
+                    label="Client secret"
+                    type="password"
```
mattnathan, line 124: 🟢 **Nit — Chrome "save password?" prompt on manual**

`type="password"` without an explicit `autocomplete` triggers Chrome's password-save heuristic when the form submits. Setting `autocomplete="off"` on the ClientID / BOS API root inputs and `autocomplete="new-password"` on the secret suppresses it and avoids polluting the user's saved-password list.

internal/cloud/autopoll.go
hexaglow: I disagree. The channel fires before the first tick, at which point it can be collected automatically.

https://pkg.go.dev/time#After

> As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.

internal/cloud/register.go
hexaglow: I've realised we've got an unnecessary mix of terminology in this feature:
- Enrollment 
- Registration
- Link
Perhaps we could cut it down to just one. What is it called inside Smart Core Connect @mattnathan ?

mattnathan: SCC uses all these terms in different contexts:
- Enrolment is the process of making SCC and BOS aware of each other
- Registration is the mechanism of discovery (/v1/device/register)
- Link is how we surface the feature in SCC UI - I don't think the back end actually uses this term

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,316 @@
...
+const defaultRegisterUrl = ref(null);
+const advancedExpanded = ref(false);
+
+// reset dialog on open
+watch(dialog, async (open) => {
+  if (!open) {
+    testSuccess.value = false;
+    return;
+  }
+  if (defaultRegisterUrl.value !== null) return;
```
mattnathan: 🔴 **Bug — fields only reset on first open of the dialog**

The early-return that caches `defaultRegisterUrl` also skips `resetFields()` on every subsequent open:

```js
watch(dialog, async (open) => {
  if (!open) { testSuccess.value = false; return; }
  if (defaultRegisterUrl.value !== null) return;   // ← also skips resetFields
  resetFields();
  // fetch defaultRegisterUrl…
});
```

Repro: open dialog → type `ABC123` → Cancel → re-open → `ABC123` is still there.

The commit message for cd580c2df says "Reset fields on dialog open, to clear out stale values" — intent looks right, the early-return is load-bearing for the wrong thing. Lifting `resetFields()` above the early-return fixes it.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,316 @@
...
+                    label="BOS API root"
+                    hint="e.g. https://bosapi.example.com"
+                    persistent-hint
+                    variant="filled"
+                    autocomplete="off"/>
+              </v-tabs-window-item>
+            </v-tabs-window>
+          </v-card-text>
+
+          <v-alert v-if="registerTracker.error" type="error" class="mx-4 mb-2" density="compact" closable>
```
mattnathan: 🔴 **Bug — dismissed error alert doesn't reappear on next submit**

With `closable` on a `v-alert` whose visibility is driven by `v-if` (no `v-model`), Vuetify tracks an internal "dismissed" flag separately from the `v-if` condition. Once the × is clicked, the alert stays hidden even when the underlying `tracker.error` is replaced with a different error.

Repro: submit garbage → error shows → dismiss × → submit more garbage → new error is set on the tracker but the alert stays hidden.

Same pattern exists on all three `closable` alerts in this file (register ~L135, test ~L47, unlink ~L63).

Two options:
- Drop `closable` — the alert disappears naturally on the next submit cycle, and errors are information the user probably wants to see.
- Keep `closable` but bind `v-model` and reset `tracker.error` at submit start.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,316 @@
...
+              </v-list-item-title>
+            </v-list-item>
+            <v-list-item>
+              <v-list-item-subtitle>Server</v-list-item-subtitle>
+              <v-list-item-title class="text-body-2 font-weight-medium">
+                {{ cloudConnection.bosapiRoot || '—' }}
+              </v-list-item-title>
+            </v-list-item>
+            <v-list-item>
+              <v-list-item-subtitle>Last checked in</v-list-item-subtitle>
```
mattnathan, line 32: 🔴 **Bug — "Last checked in" row displays nonsense relative times, and the subtitle wording reads awkwardly**

Three related issues in this row:

- **Tense.** Values always render in the future (`"in 28 seconds"`) instead of the past (`"28 seconds ago"`), which doesn't make sense for a value that's always historical.
- **Cadence.** The value jumps forward by ~10 s at a time (tracking the server-side poll cadence) rather than ticking up smoothly second-by-second.
- **Wording.** `"Last checked in"` pairs as a past-tense verb (`"I last checked in…"`) with the value, producing `"Last checked in in 28 seconds"` / `"Last checked in 28 seconds ago"` — neither grammatical.

Ticking `now` at 1-second resolution should fix both display issues — the relative time will count up smoothly and stay ahead of incoming check-ins:

```js
const {now} = useNow(SECOND);
```

For the wording, a noun-phrase subtitle composes cleanly with the relative-time value:

```html
<v-list-item-subtitle>Last checked</v-list-item-subtitle>
<v-list-item-title>{{ lastCheckInText }}</v-list-item-title>
```

→ `"Last checked — 28 seconds ago"` / `"Last checked — just now"` / `"Last checked — Never"`. `"Last check-in"` also fine if it matches the API vocabulary you prefer.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,316 @@
+<template>
+  <v-dialog v-model="dialog" max-width="480">
+    <v-card>
+      <v-card-title>{{ isLinked ? 'Manage Cloud Link' : 'Link to Cloud' }}</v-card-title>
+
+      <!-- Connected state: show current link info + Unlink button -->
+      <template v-if="isLinked">
+        <v-card-text>
+          <v-list density="compact">
+            <v-list-item>
```
mattnathan: 🟡 **Improvement — group the Manage list by semantic role**

"Status", "Last checked in" and "Last error" are all about the connection's current health; "Client ID" and "Server" are static metadata. Currently they're interleaved so the eye doesn't get a clear "is this working?" / "what is this connected to?" distinction.

Suggested ordering:

```
Status         [chip]
Last checked   28 seconds ago
Last error     …           ← only when present
Client ID      1
Server         http://localhost:8080
```

(Keeps the eye tracking: headline state first, then supporting data.)

mattnathan (commented):
Second-pass review, mostly UI polish — inline comments cover the bugs and concrete improvements.

**Non-inline notes:**

- **Password-manager prompt on Client secret save.** Can't be suppressed reliably via `autocomplete` — Chrome's save heuristic is form-submit-triggered and ignores the attribute for `type=password`. Acceptable to land as-is, but see the inline comment about dropping the `new-password` value, which currently makes things worse by triggering the password-generator prompt on focus.

- **Error text for credential rejection.** `"Invalid client ID or secret"` is the UI's mapping of `invalid_client_credentials`. Reads a bit terse — something like `"Credentials rejected — check the Client ID and Client secret"` would be friendlier. Purely stylistic.

- **Optional: cross-populate URL between tabs.** If an engineer enters a host into either form's URL field, it'd be handy to offer to pre-fill the other tab's equivalent (they're related by convention: `Register URL = {BosapiRoot}/v1/device/register`, though not forced to match). Floating as a future thought.

- **Optional: "Test" button on the Link form.** Link submit already does a sync check-in via `Conn.Register`, so every Link press is effectively a test. But if someone wants to validate creds before committing them, there's no way to dry-run. Not for this PR — just noting the asymmetry between Link and Manage (Manage has Test Connection; Link doesn't).

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,316 @@
...
+              <v-tabs-window-item value="code">
+                <v-text-field
+                    v-model="enrollmentCode"
+                    label="Enrollment code"
+                    hint="6-character code from Smart Core Connect"
+                    persistent-hint
+                    variant="filled"
+                    class="mb-3 mt-1"
+                    autofocus/>
+                <v-checkbox
```
mattnathan, line 97: 🟡 **Improvement — tighten spacing around the "Show advanced options" checkbox**

`v-checkbox` reserves vertical space for hint/error text by default (the `details` slot), which adds a visible gap between the checkbox and the Register URL input when expanded. Adding `hide-details` removes the reserved space:

```html
<v-checkbox
    v-model="advancedExpanded"
    hide-details
    label="Show advanced options"/>
```

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,316 @@
...
+                    v-model="advancedExpanded"
+                    label="Show advanced options"/>
+                <div v-if="advancedExpanded">
```
mattnathan: 🟡 **Improvement — animate the Advanced block expand/collapse**

Right now the block pops in/out. Wrapping in `<v-expand-transition>` slides it smoothly. Note this requires `v-if` → `v-show` (the transition needs an element to animate):

```html
<v-expand-transition>
  <div v-show="advancedExpanded">
    <v-text-field v-model="registerUrlOverride" … />
  </div>
</v-expand-transition>
```

ui/ops/src/routes/system/components/LinkCloudDialog.vue
```diff
@@ -0,0 +1,316 @@
...
+                    v-model="clientId"
+                    label="Client ID"
+                    variant="filled"
+                    class="mb-3 mt-1"
+                    autocomplete="off"/>
+                <v-text-field
+                    v-model="clientSecret"
+                    label="Client secret"
+                    type="password"
+                    autocomplete="new-password"
```
mattnathan: 🟡 **Improvement — drop `autocomplete="new-password"` on Client secret**

`autocomplete="new-password"` signals to the browser "this is a signup field", which triggers Chrome's *suggest a generated password* popup when the field is focused. For a field where the user pastes a credential issued elsewhere, that's actively unhelpful — it offers irrelevant junk the user has to dismiss.

Chrome will still offer to save the pasted secret on submit regardless of `autocomplete` (that's a browser-design decision, not attribute-controlled), so removing `autocomplete="new-password"` doesn't add a prompt — but it does remove the generator prompt. Strictly better UX.

Suggest either dropping the attribute entirely, or `autocomplete="off"` for parity with the other two manual fields.

hexaglow: Have followed the new suggestion to remove them, but I'm a little confused as you suggested adding them? Did that used to work but not any more? I'd like to know for future reference.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
hexaglow: I've fixed the bug where future timestamps appear. I don't agree with the comment about the phrasing. 'Check in' is the verb here, so 'last checked in' scans fine. What's the objection exactly?

hexaglow:
@mattnathan 

> "Test" button on the Link form

This isn't a good idea, because we don't currently have a way to do a real dry run request. If we test the connection, then a real check in is sent to the server and will presumably be recorded. In the existing situations we perform a test, that's fine:
- When testing an existing registration, we're periodically checking in anyway, so there's no harm in sending an extra one.
- When testing a new registration that we intend to save, then if it fails the server won't record it, and if it succeeds it simply becomes the first check-in of the new saved registration - no harm donme

However, if we allow testing a check-in without saving, that'll appear to the server as a node checking in once and then disappearing - contrary to user expectations I'd say. By binding the testing to the registration, the flow makes it clear that the node and cloud server become linked here.

> cross-populate URL between tabs

Other aspects of the system design have avoided requiring any particular relationship between registration URL and API server URL (no need to even be on the same domain etc.). I know they currently are on the same server but I don't want to build that knowledge into the client because then it becomes a sort of promise we shouldn't be making.

> Error text for credential rejection

I've assumed that the user using this flow is reasonably technical, so terse is fine IMO.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
mattnathan: TBH I didn't catch this feedback (from Claude) when it made it. This is my mistake. The goal of not prompting to save this kind of thing was accurate, but the specific advice was inaccurate.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
mattnathan: When I first read it I read it as _last checked [the connection] in_, which makes no sense. To avoid the confusion either lean into the _the connection_ bit or the _check-in_ bit as "Last checked" or "Last check-in". Does that help?

mattnathan:
> @mattnathan
> 
> > "Test" button on the Link form
> 
> This isn't a good idea, because we don't currently have a way to do a real dry run request. If we test the connection, then a real check in is sent to the server and will presumably be recorded. In the existing situations we perform a test, that's fine:
> 
> * When testing an existing registration, we're periodically checking in anyway, so there's no harm in sending an extra one.
> * When testing a new registration that we intend to save, then if it fails the server won't record it, and if it succeeds it simply becomes the first check-in of the new saved registration - no harm donme
> 
> However, if we allow testing a check-in without saving, that'll appear to the server as a node checking in once and then disappearing - contrary to user expectations I'd say. By binding the testing to the registration, the flow makes it clear that the node and cloud server become linked here.

Sorry for the confusion, the test button was intended only for the manual case. While we could test the url is right (using a health endpoint or something), I think that is out of scope.

mattnathan:
> > cross-populate URL between tabs
> 
> Other aspects of the system design have avoided requiring any particular relationship between registration URL and API server URL (no need to even be on the same domain etc.). I know they currently are on the same server but I don't want to build that knowledge into the client because then it becomes a sort of promise we shouldn't be making.

Fair enough.

ui/ops/src/routes/system/components/LinkCloudDialog.vue
hexaglow: It looks like this now which I think works 
image

hexaglow:
> > [...]
> 
> Sorry for the confusion, the test button was intended only for the manual case. While we could test the url is right (using a health endpoint or something), I think that is out of scope.

The problem is still there for the manual case - if we send a request to test that the creds are OK, that's a real check-in, but calling it a 'test' implies otherwise. We don't have a way to ask the server "check my creds are OK for a check in, but don't process it".

mattnathan (approved):
Approving as the code looks good now. The comment about the phrasing of "last checked in" I'll leave up to you.

mattnathan:
> > > [...]
> > 
> > 
> > Sorry for the confusion, the test button was intended only for the manual case. While we could test the url is right (using a health endpoint or something), I think that is out of scope.
> 
> The problem is still there for the manual case - if we send a request to test that the creds are OK, that's a real check-in, but calling it a 'test' implies otherwise. We don't have a way to ask the server "check my creds are OK for a check in, but don't process it".

Ah, good point. For some reason I'd skipped this, I was thinking of the exchange/register flow only where testing would literally adjust the registration state, for manual it just updates the 'last seen' bits - but this is still important and I'd overlooked it.

PR #810, opened 2026-04-23
ui: refactor date constant usage

hexaglow:
Found something strange while working on #809.

`formatTimeAgo` in util/date.js was taking what should be constants as parameters. All the call sites just used the constants you'd expect, and I can see no reason why you'd pass in any other values to them.


mattnathan (approved):
During my travels in SCC I came across vue-use which also includes a useNow composable very similar to utils/date. Maybe one to look into, maybe as part of a DRY review? Not here, but FYI as you brought this up.

PR #812, opened 2026-04-24
Add debug UI to cloudsim

hexaglow:
To make cloudsim easier to set up when doing manual testing, add a very basic (and ugly) web app to allow viewing and managing all the resources exposed through the management API, such as sites, nodes, deployments etc. This is quicker than poking the API endpoints manually. It's especially useful for generating enrollment codes quickly.

Like the management API itself, this has no authentication and is only intended for local development.

It's a templated HTML app because that seemed like the smallest way of doing it and adds no dependencies or build steps.

I had Claude generate this to make my own work on #809 easier, but it's probably handy for others too.


mattnathan:
@hexaglow were all the go changes intentional in this PR? From the PR desc I was expecting this to be mostly HTML templates with some plumbing, but I see pagination-related and other code changes in the sim server code?

hexaglow:
@mattnathan those changes reorder the deployments and check-ins to be most-recent-first

mattnathan (approved):

PR #813, opened 2026-04-27
feat/driver health

deanredfern26:
- Adds a 'system' health check as a first class concept to the driver services framework. This gives drivers a common pattern to use  to say whether they are functioning on a 'system' level. It is still up the the implementation of the driver to determine if they use this or define their own health checks for a 'system level' check.
- Adds a page to the ops UI to surface these checks (or any named checks) in a flexible way, which can show the health of each subsystem from a high level. The idea is that this page is flexible enough to group related checks and desribe any health checks so the health data is more accessible to operators

Also add health info to mock drivers and restructure the vanti ugs config so the mock drivers are named per floor (`floor/02/drivers`) rather than all on a controller level, which reflects reality better and make the health UI cleaner

For drivers which talk to one or just a few servers, this change makes it easy to just implement the check and add the driver's name to the UI config, as the system health check uses the driver's name. But as we have a diverse range of drivers which all work differently, a single health check might not always work. 
I.e. the BACnet driver can talk to multiple controllers from a single driver instance. 

Some ramblings: we could still add this 'system check' pattern to things like the BACnet driver, and then just structure the project config in a way that a single bacnet driver instance talks to a single bacnet controller. That would make detecting and surfacing issues a bit easier without adding more complexity to the driver itself.

image

### DIFF
--- a/example/config/vanti-ugs/app.conf.json
+++ b/example/config/vanti-ugs/app.conf.json
@@ -7,906 +7,2676 @@
   },
   "drivers": [
     {
-      "name": "van/uk/brum/ugs/eg-ac-01/driver/basement",
+      "name": "van/uk/brum/ugs/floor-b1/driver/lighting-ctrl-a",
       "type": "mock",
+      "healthCheck": {
+        "faultProbability": 0.1
+      },
       "devices": [
         {
           "name": "van/uk/brum/ugs/devices/LTF-LN1-01",
-          "traits": [{"name": "smartcore.traits.Light"}, {"name": "smartcore.bos.Status"}],
-          "appearance": {"title": "Bar Lights"},
-          "membership": {"subsystem": "lighting"},
-          "location": {"floor": "Basement", "zone": "Bar"}
+          "traits": [
+            {
+              "name": "smartcore.traits.Light"
+            },
+            {
+              "name": "smartcore.bos.Status"
+            }
+          ],
+          "appearance": {
+            "title": "Bar Lights"
+          },
+          "membership": {
+            "subsystem": "lighting"
+          },
+          "location": {
+            "floor": "Basement",
+            "zone": "Bar"
+          }
         },
         {
           "name": "van/uk/brum/ugs/devices/LTF-LN1-02",
-          "traits": [{"name": "smartcore.traits.Light"}, {"name": "smartcore.bos.Status"}],
-          "appearance": {"title": "Ceiling Lights"},
-          "membership": {"subsystem": "lighting"},
-          "location": {"floor": "Basement"}
+          "traits": [
+            {
+              "name": "smartcore.traits.Light"
+            },
+            {
+              "name": "smartcore.bos.Status"
+            }
+          ],
+          "appearance": {
+            "title": "Ceiling Lights"
+          },
+          "membership": {
+            "subsystem": "lighting"
+          },
+          "location": {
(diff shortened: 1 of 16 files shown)


example/config/vanti-ugs/app.conf.json
```diff
@@ -7,906 +7,2676 @@
   },
```
dloveen-v, line 7: Can you affect the vanti-ugs cohort to be consistent too?

ui/ops/src/routes/system/components/SubsystemHealthCard.vue
```diff
@@ -0,0 +1,394 @@
...
+                      <v-icon size="12" class="mr-1" color="warning">mdi-clock-alert-outline</v-icon>
+                      Abnormal since {{ formatTimestamp(item.abnormalTime) }}
+                    </div>
+                    <!-- Faults -->
+                    <div
+                        v-for="(fault, fi) in item.faults?.currentFaultsList"
+                        :key="fi"
+                        class="detail-line text-caption text-warning">
+                      <v-icon size="12" class="mr-1" color="warning">mdi-alert-outline</v-icon>
+                      {{ fault.summaryText }}
```
dloveen-v: This should be formatted to sentence case in the UI

dloveen-v: Or placed inside a `<code></code>` block so the fact that it comes directly from the backend is intuitive.

example/config/vanti-ugs/ui-config.json
```diff
@@ -1258,6 +1279,69 @@
...
       }
+    },
+    "system": {
+      "subsystemHealth": {
+        "subsystems": [
+          {
+            "title": "Lighting",
+            "icon": "mdi-lightbulb-group-outline",
+            "description": "Lighting control",
+            "checks": [
```
dloveen-v, line 1290: How will checks render in a building with many floors (e.g. 36+)?

deanredfern26: I think we will just have to have a look and see how it is per site. If it is too hard to read, we can always have a bigger range to group the floors on the top level like `floors 1-5` instead of one floor per item as in the example. Then when expanded it will list all floor 1-5 checks.

ui/ops/src/routes/system/components/SubsystemHealthCard.vue
```diff
@@ -0,0 +1,400 @@
...
+                        <span v-if="item.reliability.lastError.detailsText" class="text-medium-emphasis d-block ml-4">
+                          {{ item.reliability.lastError.detailsText }}
+                        </span>
+                      </div>
+                      <!-- Root cause -->
+                      <div v-if="item.reliability.cause" class="detail-line text-caption text-medium-emphasis">
+                        <v-icon size="12" class="mr-1">mdi-transit-connection-variant</v-icon>
+                        Caused by: {{ item.reliability.cause.displayName || item.reliability.cause.name }}
+                        <span v-if="item.reliability.cause.error && item.reliability.cause.error.summaryText" class="d-block ml-4">
+                          {{ toSentenceCase(item.reliability.cause.error.summaryText) }}
```
dloveen-v, line 106: Should this be a computed variable? As `toSentenceCase` will re-run on each render

dloveen-v (approved):

PR #817, opened 2026-04-30
feat/driver system health

deanredfern26:
[3CS-75](https://vanti.youtrack.cloud/issue/3CS-75) Update the opcua, hikcentral and gallagher drivers to use the driver system health check. Below is it running at 3CS (the lift server is actually offline). 

I have started with these 3 and done them all at once as they are all single server drivers so easy to add the check.

image

### DIFF
--- a/pkg/driver/gallagher/client.go
+++ b/pkg/driver/gallagher/client.go
@@ -1,21 +1,26 @@
 package gallagher
 
 import (
+	"context"
 	"crypto/tls"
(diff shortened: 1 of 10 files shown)


dloveen-v (approved):
LGTM. `serverError` is a little bit misleading for `http authorization / forbidden` errors but I get why it's more closely related to a commissioning problem than a client related issue.

PR #818, opened 2026-05-01
auth/policy: fix mutating opa compilers in cached policies

dloveen-v:


pkg/auth/policy/rego.go
```diff
@@ -49,9 +55,21 @@ func (p *cachedStatic) EvalPolicy(ctx context.Context, query string, input Attri
...
 	}
 
+	// Lock is held per cache entry during evaluation; see regoCacheEntry.mu
+	// for synchronization details.
+	entry := p.getCacheEntry(query)
+	entry.mu.Lock()
+	defer entry.mu.Unlock()
+
 	return partial.Rego(rego.Input(input)).Eval(ctx)
 }
```
deanredfern26: getCacheEntry can return nil — latent panic risk (rego.go:55-65)
                                                                                                                                                                                                                                                                                                    
  EvalPolicy calls loadPartialCached then immediately calls getCacheEntry and dereferences the result:                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                    
```
  entry := p.getCacheEntry(query)                                                                                                                                                                                                                                                                   
  entry.mu.Lock()  // panics if entry == nil
```

  getCacheEntry returns p.cache[query], which is nil for a missing key. This is safe only because loadPartialCached always populates the cache before returning nil error — but that invariant isn't enforced by the type system. If loadPartialCached is ever refactored, this becomes a           
  nil-pointer panic.

Also, partial is returned from loadPartialCached, while entry is fetched separately via getCacheEntry. Both refer to the same cache slot, but this isn't visible from the code — a reader has to hold both code paths in their head to understand that the mutex on entry is what makes using partial   
  safe. The suggestion above eliminates this confusion.
                                                                                                                                                                                                                                                                                                    
  Suggestion: Have loadPartialCached return *regoCacheEntry directly (it already has the entry), and remove getCacheEntry entirely. This makes the relationship between the lock and the partial result explicit:                                                                                   
   
```
  entry, err := p.loadPartialCached(ctx, query)                                                                                                                                                                                                                                                     
  if err != nil { return nil, err }
  entry.mu.Lock()
  defer entry.mu.Unlock()                                                                                                                                                                                                                                                                           
  return entry.partialResult.Rego(rego.Input(input)).Eval(ctx)
```

pkg/auth/policy/rego.go
```diff
@@ -31,15 +31,21 @@ func (p *static) EvalPolicy(ctx context.Context, query string, input Attributes)
...
-	cache    map[string]*regoCacheEntry
-	cacheM   sync.Mutex
+	modules map[string]*ast.Module
+	cache   map[string]*regoCacheEntry
+	cacheM  sync.Mutex
 }
 
 func newCachedStatic(compiler *ast.Compiler) *cachedStatic {
+	// Create a defensive shallow copy of the module map so cachedStatic owns
+	// an immutable snapshot, protecting against external mutations.
```
deanredfern26: The module map values (*ast.Module) are still shared pointers. If OPA ever mutates module objects in-place (rather than just the map), this copy wouldn't help. The current copy correctly protects against map mutations (key additions/removals on the external compiler's module map), but     
  calling it "immutable" is slightly misleading. Worth wording it as "protects against external map mutations" rather than "immutable snapshot."

deanredfern26 (changes requested):

deanredfern26 (approved):

PR #819, opened 2026-05-01
feat/bacnet controller health

deanredfern26:
[3CS-75](https://vanti.youtrack.cloud/issue/3CS-75) This change adds a 'controller level' health check concept to the BACnet driver. Each controller (distinguished by its IP address) automatically gets a named health check like `<driver-name>/<ip-address>`. The controller is unhealthy if a configurable percentage of its devices are erroring. This is what is looks like at 3CS (which is accurate as floors 6-9 are currently all off):

image

Also adds a small change to the subsystem health page to make the floors collapsable

### DIFF
--- a/pkg/driver/bacnet/config/root.go
+++ b/pkg/driver/bacnet/config/root.go
@@ -46,6 +46,10 @@ type Root struct {
 	// SystemHealth represents the system-level health check configuration. If not configured,
 	// occupant and equipment impact will default to UNSPECIFIED.
 	SystemHealth Health `json:"systemHealth"`
+
+	// ControllerHealthThreshold is the % proportion of devices on a BACnet controller (IP address)
+	// that must be failing before the controller itself is marked unhealthy. Range [0, 100], default 50%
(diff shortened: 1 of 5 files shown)


pkg/driver/bacnet/config/root.go
```diff
@@ -46,6 +46,10 @@ type Root struct {
 	// SystemHealth represents the system-level health check configuration. If not configured,
 	// occupant and equipment impact will default to UNSPECIFIED.
 	SystemHealth Health `json:"systemHealth"`
+
+	// ControllerHealthThreshold is the proportion of devices on a BACnet controller (IP address)
+	// that must be failing before the controller itself is marked unhealthy. Range [0, 1], default 0.5.
+	ControllerHealthThreshold float64 `json:"controllerHealthThreshold,omitempty"`
```
dloveen-v: This would be better as a percentage IMO

dloveen-v (approved):
LGTM

PR #822, opened 2026-05-01
driver/helvarnet: add controller level health checks

deanredfern26:
adds controller level health checks which give named checks to IP controllers and report if they are healthy. similar to the recently added ones for the BACnet driver.

image

### DIFF
--- a/pkg/driver/helvarnet/health.go
+++ b/pkg/driver/helvarnet/health.go
@@ -2,7 +2,9 @@ package helvarnet
 
 import (
 	"context"
+	"fmt"
 	"strconv"
+	"sync"
 
 	"github.com/smart-core-os/sc-bos/pkg/proto/healthpb"
 )
@@ -12,6 +14,7 @@ const (
 	BadResponseCode       = -2
 	UnrecognisedErrorCode = -100
 	SystemName            = "HelvarNet Lighting"
+	ControllerUnhealthy   = "ControllerUnhealthy"
 )
 
 type State struct {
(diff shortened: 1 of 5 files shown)


pkg/driver/helvarnet/health.go
```diff
@@ -120,6 +123,92 @@ func updateDeviceFaults(ctx context.Context, status int64, fc *healthpb.FaultChe
...
+	c.states[name] = true
+	c.recalculate(ctx)
+}
+
+// setOK marks the named device as healthy and recalculates controller health.
+func (c *controllerHealth) setOK(ctx context.Context, name string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.states[name] = false
+	c.recalculate(ctx)
```
dloveen-v, line 182: `recalculate` calls can get expensive. I should've spotted this in the BacNET PR too but is there a way to make `setOK` check the existing state is already false before calling `recalculate()`?  Same for `setFailing`.

dloveen-v (commented):
There seems to be a lot of duplication in the health.go files for each driver. Is there a way each driver can just import a shared health.go?

deanredfern26:
> There seems to be a lot of duplication in the health.go files for each driver. Is there a way each driver can just import a shared health.go?

[SCB-1358](https://vanti.youtrack.cloud/issue/SCB-1358) can we share health code accross drivers?
I have made a ticket to look into it, possibly outside the scope of this PR

dloveen-v (approved):

PR #820, opened 2026-05-01
auth/policy: migrate OPA imports and rego files to v1 API

hexaglow:
We're using the deprecated OPA v0 API.

1. Switch to using the OPA v1 API when using Rego
2. Rewrite all Rego policy files to account for language changes in v1

PR #823, opened 2026-05-01
auth/policy: fix race when multiple shared query names access the cache

dloveen-v:


hexaglow:
@dloveen-v can I get some background on this please? What was the issue? I am making changes to the same area in #820 so would like to understand

dloveen-v:
@hexaglow 

This is part of an attempt to fix a runtime hang when initializing the default sc-bos Rego policies alongside Rego policies needed for custom project traits. The previous change https://github.com/smart-core-os/sc-bos/commit/33943fd5e3c0c6725c25ed87f10b5fc93f9258ed was implemented as a shared compiler was being mutated for each new query being passed. https://github.com/smart-core-os/sc-bos/commit/33943fd5e3c0c6725c25ed87f10b5fc93f9258ed instead caches the Rego modules and initializes a new Rego compiler for each query parameter.

Then with this commit, I realized `get or create` should be a single atomic operation as opposed to 2 separate operations.

However, there is still an underlying problem that hasn't been discovered as most if not all of the deployments that use sc-bos do not use `cached: true` when initializing the default sc-bos policies. The hang is a major problem when using `cached: true` in sc-bos/pkg/auth/policy.Default. I have found a test case that reproduces the hang. I am working on a fix now.

hexaglow:
@dloveen-v request a review from me when the follow-up changes are ready

PR #826, opened 2026-05-05
auth/policy: add ability to avoid using cached rego policies

dloveen-v:
Projects that use custom smartcore traits (and therefore, custom .rego policy files) need a way to specify whether to cache compiled .rego files.

### DIFF
--- a/pkg/auth/policy/rego.go
+++ b/pkg/auth/policy/rego.go
@@ -37,11 +37,11 @@ type cachedStatic struct {
 }
 
 func newCachedStatic(compiler *ast.Compiler) *cachedStatic {
-	// Create a defensive shallow copy of the module map so cachedStatic owns
+	// Create a defensive deep copy of the module map so cachedStatic owns
 	// a snapshot, protecting against external mutations.
 	modules := make(map[string]*ast.Module, len(compiler.Modules))
 	for k, v := range compiler.Modules {
-		modules[k] = v
+		modules[k] = v.Copy()
 	}
 	return &cachedStatic{
 		cache:   make(map[string]*regoCacheEntry),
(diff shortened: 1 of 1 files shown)


dloveen-v:
@hexaglow I have been trying to reproduce the issue experienced in a site deployment last week. I think the problem was when using cached policies, there was a race condition that would ultimately lead to duplicated map entries in `rego.go` if `cached = true`. It's looking like the recent commits have resolved this.

The reason for this PR is to expose the choice of caching to test files that sit in project repositories.

pkg/auth/policy/rego.go
```diff
@@ -195,13 +195,16 @@ func Default(cached bool) Policy {
 	}
 }
 
-func FromFS(f fs.FS) (Policy, error) {
+func FromFS(useCaching bool, f fs.FS) (Policy, error) {
```
hexaglow: This change doesn't seem to achieve anything because:
1. It's only called in the tests
2. The tests enable caching, which is the same as the previous behaviour

hexaglow (changes requested):
I don't understand why using custom traits implies the need to turn off caching. Please expand the PR description + commit message to specify what breaks when caching is enabled and why the change fixes it.

pkg/auth/policy/rego.go
dloveen-v: @hexaglow This is a change to the function signature so that it can be called from other repositories.

hexaglow: 👍

hexaglow (commented):
Still uncertain about details here:

1. Is there a problem with the cached policy? i.e. is there still a bug or has it been fixed by the previous commit and the deep copy here?
2. Is there any difference between using the default policy and an external project-defined one that means external ones shouldn't use caching? What's the difference?

You've said that allowing disabling caching is for test files in another repo. If you have a test that can detect a problem with the caching, better commit it here. If you have a link to the consuming code that would be helpful.

hexaglow:
@dloveen-v side note: I've realised while thinking about this package that a 'compiler pool' based implementation might remove a lot of potential concurrency issues. The concurrency invariants in Rego don't seem to be well documented, so it's difficult to analyse what types we can share between goroutines and which need to be single-threaded etc. Waht me to put that together?

dloveen-v:
> Is there a problem with the cached policy? i.e. is there still a bug or has it been fixed by the previous commit and the deep copy here?
> Is there any difference between using the default policy and an external project-defined one that means external ones shouldn't use caching? What's the difference?
> You've said that allowing disabling caching is for test files in another repo. If you have a test that can detect a problem with the caching, better commit it here. If you have a link to the consuming code that would be helpful.

The problem with cached policy has been fixed by previous commits and the deep copy here is to ensure robustness.

The use of project-defined policies over default ones are subject to decision at a project level, not by sc-bos. This is the reason the tests aren't in sc-bos. For instance, sc-bos has an allow all policy. If these sc-bos rego files are processed first, there is no need for a project-level rego policy.

The link is [here](https://github.com/vanti-dev/nvn-40-lh/pull/131/changes/9f345c7936ac7fddfbbb518b0d3ba4f6b8bc604b#diff-686ccc1a8e970632fdcfb31f11da2c464061abbe35d47ed9cf7aee90bd04333eR54)

hexaglow (changes requested):
I understand now.

I believe that all consumers should be using the cached implementation really. But adding a caching option to `FromFS` makes it more symmetric with `Default` so that's OK. The only thing is, adding a parameter to `FromFS` breaks API compatibility, which is to be avoided. Suggestion:

```go
// FromFSCached returns a Policy from the .rego files in the provided FS with
// optional caching.
func FromFSCached(f fs.FS, cached bool) (Policy, error) {
   // ... existing implementation
}

// FromFS is the same as FromFSCached with cached=true.
// It is present for backward-compatibility.
// 
// Deprecated: Use FromFSCached directly.
func FromFS(f fs.FS) (Policy, error) {
    return FromFSCached(f, true)
}
```

(tweak the name of the new function as you see fit)

hexaglow (approved):

PR #827, opened 2026-05-08
driver/bacnet,driver/helvarnet: fix initial state being false always stops recalculate

dloveen-v:


deanredfern26 (approved):
fix looks good. even better would be to add a test that would have caught the original bug and test future regressions :D

dloveen-v:
> even better would be to add a test that would have caught the original bug and test future regressions 

Yeah the per driver system health check stuff needs refactoring. There's a lot of duplicated code that tests would just add to. Let's revisit.

PR #829, opened 2026-05-08
policy: restrict mutating CloudConnectionApi actions to admin

hexaglow:
CloudConnectionApi access is managed by the default smartcore policy file, which permits anyone with trait:write to perform any operation.

Only an admin should be able to add or remove a connection to
Smart Core Connect. Add a specific policy file to enforce this.

### DIFF
--- a/pkg/auth/policy/default/test_general.rego
+++ b/pkg/auth/policy/default/test_general.rego
@@ -146,3 +146,118 @@ test_bos_zone_mismatch if {
     with input as tenant_request("smartcore.bos.light.v1.LightApi", "GetBrightness", {"name": "zone/2"}, ["zone/1"])
     with data.system.known_traits as mock_known_traits
 }
+
+# --- CloudConnectionApi: mutations restricted to admin / cert; reads gated on trait:read ---
+
+cert_request(service, method, request) := result if {
+  result := {
+    "service": service,
+    "method": method,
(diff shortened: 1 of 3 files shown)


pkg/auth/policy/default/test_general.rego
```diff
@@ -146,3 +146,114 @@ test_bos_zone_mismatch if {
...
+test_cloud_trait_write_Get if {
+  data.smartcore.bos.ops.cloud.v1alpha.CloudConnectionApi.allow
+    with input as permission_request(cloud_service, "GetCloudConnection", {}, ["trait:write"])
+}
+
+# Test op is gated on trait:read.
+test_cloud_trait_read_Test if {
+  data.smartcore.bos.ops.cloud.v1alpha.CloudConnectionApi.allow
+    with input as permission_request(cloud_service, "TestCloudConnection", {}, ["trait:read"])
+}
```
deanredfern26: this test is checking that read permission is allowed, but the actual rule is specifiying write permission

```
# TestCloudConnection is not configuration - regular write access is sufficient
allow if {
  input.method == "TestCloudConnection"
  token_has_permission("trait:write")
}
```

deanredfern26 (changes requested):

hexaglow:
@deanredfern26 I've added a new CI job that could have detected that test discrepancy.

deanredfern26 (approved):

PR #830, opened 2026-05-11
feat/space UI preset, health card & mock, bacnet health

deanredfern26:
- update the health error description in the bacnet driver
- add `health` support for the mock driver
- remove experimental flag preventing health card not showing
- space ui pull presets on stream reconnect and resize preset name

### DIFF
--- a/ui/space/src/routes/components/LightCard.vue
+++ b/ui/space/src/routes/components/LightCard.vue
@@ -7,17 +7,17 @@
           <v-btn
               v-bind="menuProps"
               :disabled="blockActions"
-              size="small"
+              size="large"
               variant="tonal"
               prepend-icon="mdi-playlist-check">
-            {{ currentPresetTitle || 'Presets' }}
+            {{ selectedPreset?.title || 'Presets' }}
           </v-btn>
         </template>
         <v-list density="compact" class="presets-list">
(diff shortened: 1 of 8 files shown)


ui/space/src/routes/components/LightCard.vue
```diff
@@ -78,12 +78,28 @@ const props = defineProps({
...
+  }
+});
+
 const lightValue = reactive(
     /** @type {ResourceValue<Brightness.AsObject, PullBrightnessResponse.AsObject>} */
     newResourceValue());
 
-// if device name changes
-watch(() => props.name, async (name) => {
+// stream brightness from effectiveName (zone or specific light group when a preset is selected)
```
dloveen-v: Can you improve this comment? It's not clear why we need `effectiveName`

deanredfern26: I have updated the comment. Why this exists is because the user wanted to be able to select a lighting preset for a whole zone, but then wanted to adjust the brightness for only the lights that remained on. This provides a way to do that by using a zone's lightGroups, so when a preset is selected the dimmable name becomes `<zone-name>/<preset-name>`.

dloveen-v (approved):

PR #834, opened 2026-05-18
refactor/driver health

deanredfern26:
[SCB-1358](https://vanti.youtrack.cloud/issue/SCB-1358) Share the common health check code accross drivers

### DIFF
--- a/pkg/driver/health/controller.go
+++ b/pkg/driver/health/controller.go
@@ -25,8 +25,9 @@ type ControllerHealth struct {
 	threshold  int    // [0,100]: % of failing devices that triggers unhealthy
 	systemName string // used in error code System field
 
-	mu     sync.Mutex
-	states map[string]healthState // device name → current state
+	mu           sync.Mutex
+	states       map[string]healthState // device name → current state
+	failingCount int
 }
 
 func NewControllerHealth(fc *healthpb.FaultCheck, threshold int, systemName string) *ControllerHealth {
@@ -60,6 +61,7 @@ func (c *ControllerHealth) SetFailing(ctx context.Context, name string) {
 		return
 	}
 	c.states[name] = stateFailing
+	c.failingCount++
 	c.recalculate(ctx)
 }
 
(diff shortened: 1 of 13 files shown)


pkg/driver/health/controller.go
```diff
@@ -0,0 +1,104 @@
...
+
+type healthState int
+
+const (
+	stateUnknown healthState = iota
+	stateFailing
+	stateOK
+)
+
+// ControllerHealth aggregates per-device health states for a single controller (identified by IP address).
```
dloveen-v: The comment says controllers are identified by IP address, but is this accurate? What if we use a hostname instead, or port numbers to the same IP.

pkg/driver/health/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/smart-core-os/sc-bos/pkg/proto/healthpb"
+)
+
+const controllerUnhealthy = "ControllerUnhealthy"
+
+type healthState int
+
+const (
+	stateUnknown healthState = iota
+	stateFailing
```
dloveen-v, line 18: It may be better to set `stateFailing` as the higher iota value than `stateOK`

pkg/driver/health/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	stateOK
```
dloveen-v: `stateOK` should be `stateOk` (lower case `k`)

pkg/driver/health/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	defer c.mu.Unlock()
+	if c.states[name] == stateFailing {
+		return
+	}
+	c.states[name] = stateFailing
+	c.recalculate(ctx)
+}
+
+// SetOK marks the named device as healthy and recalculates controller health.
+func (c *ControllerHealth) SetOK(ctx context.Context, name string) {
```
dloveen-v: lower case k `Ok`

pkg/driver/health/controller.go
```diff
@@ -0,0 +1,104 @@
...
+	defer c.mu.Unlock()
+	if c.states[name] == stateOK {
+		return
+	}
+	c.states[name] = stateOK
+	c.recalculate(ctx)
+}
+
+// recalculate must be called with c.mu held.
+func (c *ControllerHealth) recalculate(ctx context.Context) {
```
dloveen-v, line 84: `nit` now that `recalculate` is shared - there are some optimisations we can do. E.g. if previous state is `ok` and new state is `failing`, then increase percentage of failures by 1/total * 100, rather than blanket iterations over the entire map each time.

dloveen-v (approved):

PR #835, opened 2026-05-18
refactor/bootstrap

deanredfern26:
[SCB-1355](https://vanti.youtrack.cloud/issue/SCB-1355) refactor app Bootstrap function

### DIFF
--- a/pkg/app/controller.go
+++ b/pkg/app/controller.go
@@ -26,6 +26,7 @@ import (
 	"google.golang.org/grpc/credentials"
 
 	"github.com/smart-core-os/sc-bos/internal/account"
+	"github.com/smart-core-os/sc-bos/internal/audit"
 	"github.com/smart-core-os/sc-bos/internal/cloud"
 	"github.com/smart-core-os/sc-bos/internal/manage/devices"
 	"github.com/smart-core-os/sc-bos/internal/node/nodeopts"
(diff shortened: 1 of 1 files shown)


dloveen-v:
Although functional, I think the comments and documentation are too important to remove.

pkg/app/controller.go
```diff
@@ -184,41 +137,200 @@ func Bootstrap(ctx context.Context, config sysconf.Config) (*Controller, error)
...
-	// Create a private key if it doesn't exist.
-	// This key is used by the controller for incoming and outgoing connections, and as part of the enrolment process.
+
 	key, keyPEM, err := pki.LoadOrGeneratePrivateKey(files.Path(config.DataDir, certConfig.KeyFile), logger)
 	if err != nil {
-		return nil, err
+		return pkiInfo{}, err
 	}
 
-	// enrollServer manages this controllers participation in a cohort.
```
dloveen-v, line 195: We shouldn't be losing any of these comments either.

dloveen-v (approved):
LGTM. Another way to refactor the init/buildxxx steps is to have a series of variadic functions for each init/buildxxx step, gated behind feature flags. It will make enabling and disabling functionality during the bootstrap process a lot simpler than the current `if` conditions. Just a suggestion. It doesn't have to be part of this PR.

PR #837, opened 2026-05-20
driver/mock: add MockDeviceApi for live trait forcing and config editing UI

deanredfern26:
## Summary

- Adds a `MockDeviceApi` gRPC service (`ForceTraitValue`, `SetDeviceAutomation`) that lets callers override live trait values and pause/resume simulated automations on mock driver devices — useful for development testing today, and intended as the foundation for E2E cloud workflow tests
- Improves the mock driver config page in the ops UI with inline editing, add/delete with confirmation, fault probability control, and a per-device state dialog for triggering the new API

## Changes

**Backend (`driver/mock`):**
- New `proto/smartcore/bos/mock/v1/mock.proto` + generated Go/JS
- `control.Controller`: thread-safe registry of `ForceFunc` handlers and `service.Lifecycle` refs per (device, trait)
- `forceJSON[T]`: generic helper bridging protojson → model update
- `MockDeviceApi` registered on each device via `node.HasServer`; key-routed by the node router using the `name` request field

**UI (`ui/ops`):**
- `MockDevices.vue`: inline name/trait editing, add device, delete-with-confirmation, fault probability slider, state button per row
- `MockDeviceStateDialog.vue`: per-trait controls (toggle/slider/number), optimistic updates that revert on error, automation pause/resume toggle
- `service.js`: exports `doSave` directly so callers don't couple to `saveAttrs` internals

## Test plan

- [ ] Add a mock device in the config page, assign traits, save — verify device appears in the node
- [ ] Open the state dialog, force an OccupancySensor to OCCUPIED — verify the trait value changes and automation is paused
- [ ] Re-enable automation via the Auto/Paused toggle — verify simulation resumes
- [ ] Force an AirTemperature `ambientTemperature` value — verify it reflects in any subscribing widget
- [ ] Mark a device for deletion, confirm — verify it is removed from config
- [ ] Attempt to save with an unnamed device — verify Save is disabled with tooltip
- [ ] Verify `go test -short ./pkg/driver/mock/...` passes


proto/smartcore/bos/mock/v1/mock.proto
```diff
@@ -0,0 +1,57 @@
...
+  repeated TraitValue values = 2;
+}
+
+// TraitValue pairs a trait with a protojson-encoded resource value.
+message TraitValue {
+  // trait is the fully-qualified trait name, e.g. "smartcore.traits.OccupancySensor".
+  string trait = 1;
+  // value_json is the trait resource encoded as protojson,
+  // e.g. for OccupancySensor: {"state":"OCCUPIED"}
+  string value_json = 2;
```
dloveen-v: Would this be better named `proto_json_value` - allowing for more encoding types later.

proto/smartcore/bos/mock/v1/mock.proto
```diff
@@ -0,0 +1,57 @@
...
+  repeated TraitAutomation automations = 2;
+}
+
+// TraitAutomation selects one or more automations and sets their active state.
+message TraitAutomation {
+  // trait is the fully-qualified trait name.
+  // Empty means all traits on the device.
+  string trait = 1;
+  // id identifies the specific automation within this trait.
+  // Empty means all automations for the matched trait(s).
```
dloveen-v, line 51: Does this need an `optional` qualifier?

proto/smartcore/bos/mock/v1/mock.proto
deanredfern26: I don't think so, this is pretty clear I feel and would still make sense if we needed more encoding types which would be in a `oneof`

proto/smartcore/bos/mock/v1/mock.proto
deanredfern26: I don't think that would add anything, both the nil case and the empty string would both have the same functionality. Making it optional would just add nil checks

proto/smartcore/bos/mock/v1/mock.proto
dloveen-v: I mean `proto_json_value` is explicit about the format of the field. JSON accepts 64 bit numbers from what I remember whereas proto json doesn't

deanredfern26: I have renamed it to `value_protojson`, it did mention in the comment but no harm in changing the name at this point

proto/smartcore/bos/mock/v1/mock.proto
```diff
@@ -0,0 +1,57 @@
...
+// TraitValue pairs a trait with a protojson-encoded resource value.
+message TraitValue {
+  // trait is the fully-qualified trait name, e.g. "smartcore.traits.OccupancySensor".
+  string trait = 1;
+  // value_protojson is the trait resource encoded as protojson,
+  // e.g. for OccupancySensor: {"state":"OCCUPIED"}
+  string value_protojson = 2;
+}
+
+message ForceTraitValuesResponse {}
```
dloveen-v, line 34: Would benefit from an `applied: bool` field? I am thinking of situations where the device and trait are found in the mock but the value being applied doesn't comply with checks. What's your opinion?

deanredfern26: Happy to add that if you can think of a use case, I was just going to rely on the rpc error response:
 - Device/trait not found → codes.NotFound
 - Value fails validation in ForceFunc → codes.InvalidArgument
 
 You comment did get me thinking though about the trait updates and I realised it did not respect the update_masks, so an update would reset all non included fields to defaults. I have fixed that now and added some tests which also show the expected behvaiour for the above error cases.

dloveen-v: Okey, that's fine. I was more thinking about partial updates being handled with partial errors than anything. The fact the `message` body is empty means we can adapt this later. So rather than restricting it now let's keep it open depending on usage (it could be only trusted sources use this RPC).

dloveen-v (commented):
Everything looks good. I am just thinking about the `occupancy` example that preserves 5 people whilst `unoccupied`. Shouldn't the model be handling this?

deanredfern26:
> Everything looks good. I am just thinking about the `occupancy` example that preserves 5 people whilst `unoccupied`. Shouldn't the model be handling this?

One of the intended goals for implementing this API is that in the future we can have tests where we can inject unrealistic / unexpected data and test that whatever flow we have works for these situations. Either a health fault check detects this or connect detects it in another way or something else. If a faulty sensor/bad driver logic is telling us there are 5 people but also saying unoccupied, we want to have a good way of surfacing this, instead of silently rejecting it and it just sitting in an error log.

Maybe the models could detect this and be responsible for data correctness. Then they would have to be linked with health in some way though and also then wired into the drivers. I think that is out of scope at the moment.

dloveen-v (approved):

PR #838, opened 2026-05-20
status: remove internal consumers and deprecate trait

deanredfern26:
## Summary

The status trait has been superseded by the health trait. This PR removes all internal consumers and marks the proto/package as deprecated so downstream projects can migrate before a final removal PR.

- Delete `statusalerts` and `statusemail` autos (no known usages; health trait covers the same need)
- Delete `cmd/tools/test-statusemail` (manual test harness for statusemail)
- Delete `pkg/zone/feature/status` zone feature (not registered anywhere)
- Remove dead `tcpClient.status` field from helvarnet driver (written but never exposed as a trait)
- Remove all UI components, composables, and API clients for the status trait
- Migrate airthings driver from `statusmap` to `healthpb.NewFaultCheck`
- Mark both proto services with `option deprecated = true` and add package-level deprecation notice to `pkg/proto/statuspb`
- Regenerate protos (also fixes a stale `data_retention` → `dataretention` UI gen path)

The `pkg/proto/statuspb` package, `pkg/util/statusmap`, the mock driver case, and the history auto entry are intentionally left in place until downstream projects have migrated.

## Test plan

- [ ] `go test -short ./...` passes (pre-existing `TestServeHTTPS` flake in `pkg/app` is unrelated)
- [ ] `cd ui/ops && yarn lint:nofix` passes
- [ ] `cd ui/space && yarn lint:nofix` passes


dloveen-v:
I think we need a discussion on how to deprecate this from sites where people rely on it. I don't think health provides 100% feature parity yet and if it does, project-based drivers need adapting before being able to deprecate from the sc-bos build. IMO, we are months away from this being merged.

deanredfern26:
No problem, we can discuss, I don't expect this to be immediate. I have created tickets to review the individual projects and from memory we aren't using the autos & zone feature (I didn't even know this one existed) in 90% of projects but happy to be corrected. 

I do think it is necessary to do this though as most of the backend code here is dead code which is being maintained uncessarily and Claude gets confused and thinks Status is a thing we want to use

deanredfern26:
@dloveen-v one last thing before you leave, yay or nay?

dloveen-v (approved):
The PR looks good. I am just conscious when committing this to main branch every dependant project repository will be forced to remove their usage of the `status` trait when simply attempting to update the sc-bos dependency. It's up to you if you want to carry this forward. I know emergency light autos depend on `status` trait and these are found in various projects.

deanredfern26:
the trait is still there, just deprecated. I think that must be enterprise wharf you are referring too as the new emergency light stuff has its own trait and does not rely on status

PR #839, opened 2026-05-21
auto/udmi: add retained field to config

dloveen-v:


deanredfern26 (approved):
I think it's worth making the qos configurable too while you're here, but not blocking

dloveen-v:
I will do this in a separate PR.

PR #840, opened 2026-05-21
feat: add ui for storage infomation

deanredfern26:
## Summary

- Adds a PostgreSQL storage monitoring card to the system dashboard, showing per-node database size (used / total with percentage when capacity is available)
- Removes the row count stat — postgres does not report \`bytes.capacity\` so a percentage cannot be shown; the card shows used size only as a single clean line per node
- Adds \`dataRetention\` trait client and \`WithDataRetention\` composable wired to \`{node}/stores/postgres\`
- Adds \`formatBytes\` utility to \`util/number.js\`
- Replaces stale generated JS/TS stubs at \`ui-gen/.../data_retention/v1\` (old proto package name) with the correct \`ui-gen/.../dataretention/v1\` matching the current proto package \`smartcore.bos.dataretention.v1\`

## Test plan

- [ ] System dashboard shows PostgreSQL card when at least one cohort node has a postgres store
- [ ] Card is hidden when no postgres store is reachable
- [ ] Each node row shows used size (e.g. \`1.06 GB\`) with no row count or broken concatenation
- [ ] Disconnected nodes show a red indicator and "disconnected" label
- [ ] Card does not appear for nodes with no postgres store configured
- [ ] No regressions from removal of \`data_retention/v1\` generated files (nothing imported them)


ui/ops/src/routes/system/components/DatabaseCard.vue
```diff
@@ -0,0 +1,132 @@
...
+        const s = resource.value;
+        const utilization = (s.bytes?.used != null && s.bytes?.capacity != null && s.bytes.capacity > 0)
+            ? (s.bytes.used / s.bytes.capacity) * 100
+            : null;
+        const usedStr = formatBytes(s.bytes?.used);
+        const totalStr = s.bytes?.capacity != null ? ` / ${formatBytes(s.bytes.capacity)}` : '';
+        const pctStr = utilization != null ? ` (${utilization.toFixed(1)}%)` : '';
+        const bytesLabel = `${usedStr}${totalStr}${pctStr}`;
+
+        return h('div', {class: 'pg-node-row'}, [
```
dloveen-v: Why isn't this simply part of the template as opposed to dynamic using render functions?

ui/ops/src/routes/system/components/CohortNodeCard.vue
```diff
@@ -251,9 +260,61 @@ import {useHasHubSystem, usePullService, usePullServiceMetadata} from '@/composa
...
+    retention: {type: Object, default: () => null},
+    itemLabel: {type: String, default: 'item'}
+  },
+  setup(props) {
+    return () => {
+      const s = props.retention;
+      if (!s) return null;
+      const rows = [];
+
+      if (s.bytes != null) {
```
dloveen-v: IMO, this needs some refactoring. It is LLM readable which is the standard nowadays but I'd still like to be able to understand what is happening at a glance.

Maybe something like this:

```
// Helper function to handle progress bar threshold colors
const getBarColor = (pct) => {
  if (pct >= 80) return 'error';
  if (pct >= 60) return 'warning';
  return 'success';
};

// 1. Storage (Bytes) Row
const { used: bytesUsed, capacity: bytesCapacity } = s.bytes || {};
const hasCapacity = bytesCapacity != null && bytesCapacity > 0;

const utilization = (bytesUsed != null && hasCapacity)
  ? (bytesUsed / bytesCapacity) * 100
  : null;

const bar = utilization != null
  ? h('v-progress-linear', {
      modelValue: utilization,
      color: getBarColor(utilization),
      height: 3,
      rounded: true,
      class: 'resource-bar'
    })
  : null;

const usedStr = formatBytes(bytesUsed);
const totalStr = bytesCapacity != null ? ` / ${formatBytes(bytesCapacity)}` : '';
const pctStr = utilization != null ? ` (${utilization.toFixed(1)}%)` : '';

rows.push(
  h('div', { class: 'resource-row' }, [
    h('span', { class: 'resource-label' }, 'Used: '),
    bar,
    h('span', { class: 'resource-value', style: { width: 'auto' } }, `${usedStr}${totalStr}${pctStr}`)
  ])
);

// 2. Items Row
if (s.items?.used != null) {
  const { used: itemsUsed, capacity: itemsCapacity } = s.items;
  const totalItems = itemsCapacity != null ? ` / ${itemsCapacity.toLocaleString()}` : '';

  rows.push(
    h('div', { class: 'resource-row' }, [
      h('span', { class: 'resource-label' }, 'Items: '),
      h('span', { class: 'resource-value', style: { width: 'auto' } }, `${itemsUsed.toLocaleString()}${totalItems} ${props.itemLabel}s`)
    ])
  );
}
```

I am still unsure about the dynamic render `h` calls.

dloveen-v (approved):

PR #845, opened 2026-05-27
fix(bacnet/health): improve BACnet device health checks and UI

deanredfern26:
## Summary

- **Refactor**: replaces the `activeHigh` bool mode in BACnet health checks with a unified `okValue` integer comparison path; existing configs are backwards-compatible (`activeHigh: true` → `okValue=0`, `activeHigh: false` → `okValue=1`) at parse time
- **Fix**: BACnet devices now transition from `UNKNOWN` to `RELIABLE` on successful reads (via `MarkRunning`); traits are eagerly polled during health checks to keep fault state fresh; `RemoveFault(DeviceUnreachable)` is used in place of `ClearFaults()` to avoid clearing unrelated faults on a successful read
- **Config**: per-trait `pollKeepAlive` interval is now configurable (default `5m`), replacing a hardcoded value
- **UI**: health check table rows now show fault details and reliability error text inline, and a UDMI points dialog is accessible from the health table; `UNSPECIFIED` normality/reliability values are excluded from abnormal checks

## Test plan

- [ ] Deploy a BACnet device with `activeHigh: true` config — confirm it still raises faults correctly after migration to `okValue`
- [ ] Deploy a BACnet device with `activeHigh: false` config — confirm fault fires when point reads 0 (not 1)
- [ ] Confirm a device that was stuck in `UNKNOWN` transitions to `RELIABLE` after a successful poll
- [ ] Open the health check table — verify fault descriptions show the read value and expected `okValue`
- [ ] Verify the UDMI points dialog opens from the health table
- [ ] Confirm `UNSPECIFIED` normality/reliability does not appear as an abnormal row

### DIFF
--- a/pkg/driver/bacnet/comm/value.go
+++ b/pkg/driver/bacnet/comm/value.go
@@ -250,6 +250,11 @@ func IntValue(data any) (int64, error) {
 	switch v := data.(type) {
 	case error:
 		return 0, v
+	case bool:
+		if v {
+			return 1, nil
+		}
+		return 0, nil
 	case uint8:
 		return int64(v), nil
 	case uint16:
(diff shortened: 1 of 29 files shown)


dloveen-v (approved):
Everything LGTM. Only suggestion is `PollKeepAlive` should be simply named `KeepAlivePollPeriod` to be consistent with the other fields.

PR #846, opened 2026-05-28
pkg/driver: set device type on all driver announcements

deanredfern26:
[SCB-1363](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1363)
Add device type metadata to announcements in drivers. Most are DEVICES but there are some exceptions.
Also removes redundant nil guards on HasMetadata calls.

### DIFF
--- a/pkg/driver/gallagher/driver.go
+++ b/pkg/driver/gallagher/driver.go
@@ -12,6 +12,7 @@ import (
 	"github.com/smart-core-os/sc-bos/pkg/driver"
 	"github.com/smart-core-os/sc-bos/pkg/driver/gallagher/config"
 	"github.com/smart-core-os/sc-bos/pkg/node"
+	"github.com/smart-core-os/sc-bos/pkg/proto/metadatapb"
 	"github.com/smart-core-os/sc-bos/pkg/proto/occupancysensorpb"
 	"github.com/smart-core-os/sc-bos/pkg/proto/securityeventpb"
 	"github.com/smart-core-os/sc-bos/pkg/task/service"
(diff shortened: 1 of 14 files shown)


pkg/driver/gallagher/driver.go
```diff
@@ -93,6 +94,7 @@ func (d *Driver) applyConfig(ctx context.Context, cfg config.Root) error {
 	announcer.Announce(cfg.ScNamePrefix,
 		node.HasServer(securityeventpb.RegisterSecurityEventApiServer, securityeventpb.SecurityEventApiServer(sc)),
 		node.HasTrait(securityeventpb.TraitName),
+		node.HasDeviceType(metadatapb.Metadata_SERVICE),
```
hexaglow: SERVICE device should be a service collection according to enum documentation - i.e. implements `smartcore.bos.services`. Intended for the service endpoints registered by the controller. This doesn't fit here.

pkg/driver/opcua/driver.go
```diff
@@ -86,11 +87,11 @@ func (d *Driver) applyConfig(ctx context.Context, cfg config.Root) error {
 
 	client := NewClient(opcClient, d.logger, cfg.Conn.SubscriptionInterval.Duration, cfg.Conn.ClientId)
 
-	a.Announce(cfg.Name, node.HasMetadata(cfg.Meta))
+	a.Announce(cfg.Name, node.HasMetadata(cfg.Meta), node.HasDeviceType(metadatapb.Metadata_SERVICE))
```
hexaglow: Same thing, usage does not match documentation of `Metadata_SERVICE`

hexaglow (changes requested):

hexaglow (approved):

PR #849, opened 2026-05-28
pkg/app: announce automations with AUTOMATION device type

deanredfern26:
[SCB-1364](https://vanti.youtrack.cloud/issue/SCB-1364)

Each automation instance is now announced with HasDeviceType(Metadata_AUTOMATION) so it appears with the correct type in the devices list. The announcement is managed dynamically as automations are created and removed from the service map.

Also adds doc comments to service.Map.GetAndListen and GetAndListenState.

### DIFF
--- a/pkg/app/controller.go
+++ b/pkg/app/controller.go
@@ -747,6 +747,7 @@ func (c *Controller) Run(ctx context.Context) (err error) {
 	}
 	announceAutoServices(c, autoServices, c.SystemConfig.AutoFactories)
 	go logServiceMapChanges(ctx, c.Logger.Named("auto"), autoServices)
+	go announceAutoDeviceTypes(ctx, c.Node, autoServices)
 	// load and start the zones
 	zoneServices, err := c.startZones(initialConfig.Zones)
 	if err != nil {
--- a/pkg/app/services.go
+++ b/pkg/app/services.go
@@ -290,6 +290,24 @@ func announceServices[M ~map[string]T, T any](c *Controller, name string, servic
 	return announceNodeServer(c.Node, name, srv)
 }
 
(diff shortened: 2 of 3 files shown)


hexaglow (changes requested):
Additional justification required.

This change results in all automations being announced as devices (under their automation name), which is not currently the case. 

- Which automations currently announce a device with their own name? I need more background on what automations would want to appear on the device list.
- Why introduce this change only for automations, and not for drivers & systems?
- We have a services API and corresponding section in the Ops UI. It is potentially confusing for an automation to appear as both a service and a device.

Providing user stories that demonstrate the need for this would be helpful.

deanredfern26:
> Additional justification required.
> 
> This change results in all automations being announced as devices (under their automation name), which is not currently the case.
> 
>     * Which automations currently announce a device with their own name? I need more background on what automations would want to appear on the device list.
> 
>     * Why introduce this change only for automations, and not for drivers & systems?
> 
>     * We have a services API and corresponding section in the Ops UI. It is potentially confusing for an automation to appear as both a service and a device.
> 
> 
> Providing user stories that demonstrate the need for this would be helpful.

Honestly I was just trying to close old tickets ([SCB-830](https://vanti.youtrack.cloud/issue/SCB-830)) that are well within the BOS sphere and not in the connect grey area. I don't have use cases.

I have added your comment to the automation specific sub-ticket [SCB-1364](https://vanti.youtrack.cloud/issue/SCB-1364), it might be that we don't want this.

hexaglow:
@deanredfern26 I remember a discussion (probably quite a long time ago) about moving services into the device namespace instead of the separate namespace we have now. Might be related to that. But that would have been for all services so a bigger thing than this. Happy to leave this closed.

PR #850, opened 2026-05-28
ui/ops: standardise OpenClose composable names

deanredfern26:
## Summary

- Drops the `Positions` suffix from all OpenClose composable exports so they follow the same `use{TraitName}` convention as OnOff and AirTemperature
- Fixes `useDescribePositions` which had silently lost its `OpenClose` prefix
- Aligns the emitted component event to `updateOpenClose` (was `updatePositions`)

| Before | After |
|--------|-------|
| `usePullOpenClosePositions` | `usePullOpenClose` |
| `useUpdateOpenClosePositions` | `useUpdateOpenClose` |
| `useDescribePositions` | `useDescribeOpenClose` |
| `useOpenClosePositions` | `useOpenClose` |
| emit `updatePositions` | emit `updateOpenClose` |

The API layer (`open-close.js`) is unchanged — those names map directly to gRPC method names.

## Test plan

- [ ] `cd ui/ops && yarn lint:nofix` passes (verified locally — 0 errors)
- [ ] Open a device with the OpenClose trait in the device sidebar and confirm the slider still works

### DIFF
--- a/ui/ops/src/routes/devices/components/DeviceSideBarContent.vue
+++ b/ui/ops/src/routes/devices/components/DeviceSideBarContent.vue
@@ -26,7 +26,7 @@
           :info="info"
           :update-tracker="updateTracker"
           :name="deviceId"
-          @update-positions="update"/>
+          @update-open-close="update"/>
     </with-open-close>
     <with-air-quality v-if="traits['smartcore.traits.AirQualitySensor']" :name="deviceId" v-slot="{resource}">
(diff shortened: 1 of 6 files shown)


deanredfern26:
Reverting — the original names were correctly aligned with the proto method names (PullPositions, UpdatePositions, DescribePositions on OpenCloseApi/OpenCloseInfo). The naming issue this was meant to fix had already been resolved.

PR #851, opened 2026-05-28
Single implementation of file downloads

hexaglow:
To allow file downloads from the Ops UI, BOS implements a pattern where a gRPC endpoint is used to generate a download URL, and the client code then triggers a download of that URL. The download URL is accessed without an Authorization header (due to browser limitations). The URL itself is signed to authorise the request.

This pattern is implemented twice:
- In `manage/devices`
- In `logdownload`, which is used by the log system and audit log download feature

We should avoid having 2 implementations of the same thing, especially when that feature is security critical.

Add a canonical implementation of the download URL pattern in `internal/download`. The Router allows multiple download types to be registered (e.g. by systems) and handles all the URL generation and signature verification. It has more extensive tests than either of the previous implementations.

This changes the download URLs that are issued, but this is not a problem because the client always retrieves the URL from the server rather than constructing it directly. All URLs are now `/download/<type>/<token>`. The configurable base URL and TTL have been removed from the log system config - these should not be required in practice and therefore safe to remove.

I have also added regression tests to the packages serving downloads to check that the refactoring hasn't introduced any new bugs.

The `manage/devices` downloads supported a configurable key which allowed multi-gateway scenarios - a gateway could handle a download URL issued by another gateway as long as their HMAC keys matched. This functionality is now available for all download types.


deanredfern26 (approved):
LGTM, nice test coverage improvements too

PR #852, opened 2026-06-01
feat/bacnet-comms-rw

deanredfern26:
improve tools/bacnet-comm-test to be able to fully demonstrate r/w comms with a remote BACnet controller & compare values on either side.

- add functionality to list objects and their values and add readme.
- fix small bug when trying to use without explicit local port.
- add ability to write a value to a point

### DIFF
--- a/cmd/tools/bacnet-comm-test/main.go
+++ b/cmd/tools/bacnet-comm-test/main.go
@@ -21,13 +21,35 @@ import (
 
 func main() {
 	args := os.Args
-	if l := len(args); l < 3 || l > 4 {
-		log.Fatalf("Usage: <cmd> nic[:port] server[:port] [device]")
+	if l := len(args); l < 3 || l > 7 {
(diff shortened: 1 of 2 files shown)


cmd/tools/bacnet-comm-test/main.go
```diff
@@ -3,30 +3,57 @@ package main
...
+	if l := len(args); l < 3 || l > 7 {
+		log.Fatalf("Usage: <cmd> nic[:port] server[:port] [device] [-write Type:instance=value] [-priority 1-16]")
 	}
 	nicPort, serverPort := args[1], args[2]
 	deviceStr := "4194303"
-	if len(args) == 4 {
-		deviceStr = args[3]
+	writeSpec := ""
+	writePriority := uint(8)
+	for i := 3; i < len(args); i++ {
```
dloveen-v: This parsing of flags is getting overly complicated. Appreciate it's not the intent of the PR but can we just use `flags.Parse()` instead?

dloveen-v (approved):
LGTM - cleaner too now. Maybe a small comment on capitalization of `objecttype.FromString` is also needed. But this works as-is for me.

PR #854, opened 2026-06-01
router: return Unimplemented instead of NotFound for known device, unknown service

deanredfern26:
[SCB-37](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-37) 

When a device key is known to the router (has at least one route for another service) but the requested service has no route for that key, return codes.Unimplemented rather than codes.NotFound.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -306,10 +306,12 @@ func (r *Router) ResolveMethod(fullName string) (Method, error) {
 
 	connResolver := ConnResolverFunc(func(mr MsgRecver) (grpc.ClientConnInterface, error) {
 		var candidates []routeID // what routes should we try to match?
+		var key string
 
 		if keyFunc, exists := service.keys[methodName]; exists {
 			// we can route by key
-			key, err := keyFunc(mr)
+			var err error
+			key, err = keyFunc(mr)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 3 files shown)


hexaglow:
@deanredfern26 this looks like a sensible change. Couple of points:

1. Tests are failing - I'm guessing one of them depends on the old behaviour
2. Have you investigated if anything depends on the old behaviour - maybe gateways? I can't think of anything off the top of my head but there might be some code that attempts to detect if an API is implemented for a device somewhere.

deanredfern26:
> @deanredfern26 this looks like a sensible change. Couple of points:
> 
>     1. Tests are failing - I'm guessing one of them depends on the old behaviour
> 
>     2. Have you investigated if anything depends on the old behaviour - maybe gateways? I can't think of anything off the top of my head but there might be some code that attempts to detect if an API is implemented for a device somewhere.

Tests are failing due to race detector tests, looks unrelated.
image

I did not check myself, I asked Claude to but I will not be lazy and check myself. I'll comment when I have.

hexaglow:
@deanredfern26 there's something weird going on with the tests. The test failing in your screenshot is not actually present in the `fix/unimplmented-not-found` tree. Rebase onto main?

deanredfern26:
Rebased and checked for issues. I can't see any issues that this change would introduce

hexaglow:
The intermittent test failure in `system/log` is addressed in #858

hexaglow (approved):

PR #856, opened 2026-06-01
auto/lights: fix flaky TestPirsTurnLightsOn (convert to synctest) + retry-abort deadlock

deanredfern26:
[SCB-1362](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1362)

## Summary

`TestPirsTurnLightsOn` intermittently failed (e.g. `auto_test.go: want failed to update brightness this time, got nil`). This PR fixes the flakiness by converting the test to `testing/synctest`, and along the way fixes a genuine latent deadlock in the automation that the old fake clock had been masking.

Two commits, kept separate since the `auto.go` fix is independently meaningful:

### 1. `auto/lights: fix retry-abort deadlock in processStateChanges`
When brightness updates fail more than `MaxRetries` times, the abort branch ran `if !cancelRetry() { <-retry }`. That branch is only ever reached from the `case <-retry:` handler, where the retry timer has **already fired and had its value drained**. With a real `time.Timer`, `Stop()` returns `false` after firing, so `<-retry` blocks forever — the processing loop hangs permanently after 3 consecutive failed updates.

The hand-rolled `testClock` previously masked this because its `Stop` always returned `true`. The fix removes the erroneous cancel/drain; a fresh timer is scheduled immediately below regardless.

### 2. `auto/lights: convert TestPirsTurnLightsOn to synctest`
The test used a hand-rolled clock and an unbuffered `processComplete` channel, assuming the test and automation goroutines alternated in lock-step. Real scheduling doesn't guarantee this, so the test occasionally consumed a stale refresh event where it expected an error-retry event.

The test now runs inside a `testing/synctest` bubble against a deterministic fake clock: it settles the automation with `synctest.Wait` and inspects the last completed process, and advances timers explicitly via `time.Sleep` of the observed TTL so each fires exactly once. Occupancy timestamps are kept relative to a fixed epoch so "N minutes ago" retains its meaning as the fake clock advances.

## Testing
- `go test -run TestPirsTurnLightsOn -count=500 -race ./pkg/auto/lights/` — pass
- `go test -short -race ./pkg/auto/lights/...` — pass
- `go vet ./pkg/auto/lights/` — clean

### DIFF
--- a/pkg/auto/lights/auto.go
+++ b/pkg/auto/lights/auto.go
@@ -209,11 +209,7 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 					zap.Error(err),
 					zap.Int("retryCounter", retryCounter),
 				)
-				// reset retries to prevent too many repeated attempts.
-				// We only get here from the <-retry branch, so the retry timer has
-				// already fired and been drained, and a fresh timer is scheduled
-				// below; there is nothing to cancel or drain here. (Calling
-				// cancelRetry()+<-retry would block, as the channel is already empty.)
+				// reset retries to prevent too many repeated attempts
 				retryCounter = 0
 			} else {
 				b.logger.Error("processState failed; scheduling retry",
--- a/pkg/auto/lights/auto.go
+++ b/pkg/auto/lights/auto.go
@@ -209,11 +209,12 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 					zap.Error(err),
 					zap.Int("retryCounter", retryCounter),
 				)
-				// reset retries to prevent too many repeated attempts
+				// reset retries to prevent too many repeated attempts.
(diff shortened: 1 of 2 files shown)


pkg/auto/lights/auto.go
```diff
@@ -209,11 +209,12 @@ func (b *BrightnessAutomation) processStateChanges(ctx context.Context, readStat
 					zap.Error(err),
 					zap.Int("retryCounter", retryCounter),
 				)
-				// reset retries to prevent too many repeated attempts
+				// reset retries to prevent too many repeated attempts.
+				// We only get here from the <-retry branch, so the retry timer has
+				// already fired and been drained, and a fresh timer is scheduled
+				// below; there is nothing to cancel or drain here. (Calling
```
dloveen-v: `cancelRetry()` always returns `true` so this comment is false, and the removal is redundant but the removal can stay.

dloveen-v (approved):

PR #863, opened 2026-06-04
drivers/helvarnet: normalise test times to UTC via timezone config

deanredfern26:
[SCB-1232](https://vanti.youtrack.cloud/issue/SCB-1232)
Helvar routers report emergency test completion times as epoch seconds
derived from their local wall-clock, so when router clocks are set to
local time (e.g. BST) the reported instant is offset from true UTC.

Add a timezone config option naming the IANA zone the router clocks
follow; reported times are re-interpreted in that zone to recover the
correct UTC instant. Defaults to UTC (no correction), and the shift is
DST-aware so it stays correct year-round.

### DIFF
--- a/pkg/driver/helvarnet/light.go
+++ b/pkg/driver/helvarnet/light.go
@@ -53,15 +53,18 @@ type Light struct {
 	database      *bolthold.Store
 	testResultSet *resource.Value // *emergencylightpb.TestResultSet
 	isEm          bool
+	// location is the timezone the device's clock is set to, used to correct device-reported times to UTC
+	location *time.Location
 }
 
-func newLight(client *tcpClient, l *zap.Logger, conf *config.Device, db *bolthold.Store, em bool) *Light {
+func newLight(client *tcpClient, l *zap.Logger, conf *config.Device, db *bolthold.Store, em bool, loc *time.Location) *Light {
 	return &Light{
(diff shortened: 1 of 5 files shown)


pkg/driver/helvarnet/light.go
```diff
@@ -473,6 +479,12 @@ func parseGetCompletionTimeResponse(r string) (*time.Time, error) {
 	if t.IsZero() {
 		return nil, fmt.Errorf("function test completion time is zero")
 	}
+	if loc != nil {
+		// the epoch seconds encode the device's local wall-clock, take that wall-clock and
+		// place it in the device's timezone to get the true instant
+		wall := t.UTC()
```
dloveen-v, line 485: This is changing the timestamp, rather than converting the timezone. Are you sure this is what is required?

e.g.
Before:
```
t = 2026-06-04 15:30:45.123456789 UTC
loc = America/New_York
```

After:
```
t = 2026-06-04 15:30:45 -0400 EDT
```

Notice the numerics are identical.

I think you want:
`t = t.In(loc).Truncate(time.Second)`

deanredfern26: This is intentional, we want to change the timestamp to UTC before it gets sent as `google.protobuf.Timestamp`. The issue is the router gives us a number that claims to be epoch seconds but was actually computed from a BST wall-clock, so we must subtract the zone offset to get the true instant before it ever goes into a Timestamp

dloveen-v (approved):

PR #864, opened 2026-06-05
logpb,ui/ops: add per-service log view widget to service sidebars

deanredfern26:
[SCB-976](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-976) Log view widget
## Summary 

Adds a live log view to the service sidebar in the ops UI, so logs for the selected driver/automation/zone/system can be read in place without leaving the page.

### LogApi: field_filter

- New `field_filter` on `PullLogMessagesRequest` — exact-match AND semantics over `LogMessage.fields`, applied to both the replay and live paths.
- Replay returns up to `initial_count` *matching* messages via the new `Model.TailMatching`, scanning older buffer entries as needed so quiet services aren't diluted by chatty neighbours.

### Ops UI: LogsCard

- New `LogsCard` in the shared service sidebar, filtering the node's log stream by the `service.id`/`service.kind` fields every service lifecycle logger carries (key names defined as constants on both the Go and JS sides).
- Ctrl+F style search: matches are highlighted in place (no filtering), Enter/Shift+Enter and chevrons cycle through them browser-style. Keystrokes are debounced and per-message formatting is cached so live appends don't re-split the whole buffer.
- Pop-out dialog showing the same live stream in a larger view, plus client-side text export of the buffered messages.
- Quiet "not available" fallback for nodes without the log system (UNIMPLEMENTED/NOT_FOUND); other stream errors are surfaced as errors.
- Log line rendering is extracted from `LogsPage` into shared `LogViewport`/`LogSearchView`/`format.js` components, with the capped message buffer shared via a `useLogBuffer` composable.

### Ops UI: service config dialog

- The always-visible inline config textarea in the service sidebar is replaced with a **View Config** button opening a read-only, copyable dialog — freeing sidebar space for the status and logs cards. Viewing stays available to users with system edit blocked; only the copy action is gated.

Server-side filtered download of full log history is deferred.

## Testing

image

- `go test -short ./pkg/proto/logpb ./pkg/app` passes (excluding `TestServeHTTPS`/`TestServeGRPC`, which hang in the WSL dev environment on a clean tree too)
- `yarn lint:nofix` passes in `ui/ops`
- Exercised manually: card stream + filtering, search navigation, pop-out, export, config dialog

### DIFF
--- a/pkg/proto/logpb/model.go
+++ b/pkg/proto/logpb/model.go
@@ -2,6 +2,7 @@ package logpb
 
 import (
 	"context"
+	"slices"
 	"sync"
 
 	"github.com/smart-core-os/sc-bos/pkg/resource"
@@ -91,17 +92,14 @@ func (m *Model) TailMatching(n int, match func(*LogMessage) bool) []*LogMessage 
 	cap := len(m.buf)
 	// Walk newest -> oldest collecting matches, then reverse into
 	// chronological order.
-	var rev []*LogMessage
-	for i := 0; i < m.bufSize && len(rev) < n; i++ {
+	var result []*LogMessage
+	for i := 0; i < m.bufSize && len(result) < n; i++ {
 		msg := m.buf[(m.bufHead-1-i+cap)%cap]
 		if match == nil || match(msg) {
-			rev = append(rev, msg)
+			result = append(result, msg)
 		}
 	}
-	result := make([]*LogMessage, len(rev))
-	for i, msg := range rev {
-		result[len(rev)-1-i] = msg
-	}
+	slices.Reverse(result)
 	return result
 }
 
--- a/pkg/proto/logpb/model.go
+++ b/pkg/proto/logpb/model.go
@@ -76,19 +76,31 @@ func (m *Model) AppendMessage(msg *LogMessage) {
 // TailMessages returns the most recent n messages in chronological order.
 // If n exceeds the number of buffered messages, all buffered messages are returned.
 func (m *Model) TailMessages(n int) []*LogMessage {
+	return m.TailMatching(n, nil)
+}
+
+// TailMatching returns up to the n most recent messages for which match
+// returns true, in chronological order, scanning back through the whole
+// buffer as needed. A nil match behaves like TailMessages(n).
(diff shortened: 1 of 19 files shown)


pkg/proto/logpb/model.go
```diff
@@ -76,19 +76,31 @@ func (m *Model) AppendMessage(msg *LogMessage) {
...
+	for i := 0; i < m.bufSize && len(rev) < n; i++ {
+		msg := m.buf[(m.bufHead-1-i+cap)%cap]
+		if match == nil || match(msg) {
+			rev = append(rev, msg)
+		}
+	}
+	result := make([]*LogMessage, len(rev))
+	for i, msg := range rev {
+		result[len(rev)-1-i] = msg
 	}
```
hexaglow: Unnecessary copy - reverse can be performed in-place. Use standard library functions (`slices.Reverse` in this case) to perform common slice operations.

ui/ops/src/routes/system/components/service-cards/LogsCard.vue
```diff
@@ -0,0 +1,151 @@
...
+  }, messagesResource);
+}
+
+// Nodes without the log system respond with UNIMPLEMENTED/NOT_FOUND; show a
+// quiet message rather than an error. Other errors are shown as errors - the
+// stream retries automatically and a successful batch clears them.
+watch(() => messagesResource.streamError, (err) => {
+  if (!err) return;
+  const code = err.error?.code;
+  if (code === 12 /* UNIMPLEMENTED */ || code === 5 /* NOT_FOUND */) {
```
hexaglow: Use constants from

    import {StatusCode} from 'grpc-web';

hexaglow (changes requested):
At the default sidebar size, the log widget in the service sidebar is so narrow it's not useful.

Image

Even with the sidebar expanded to its maximum width, even a short log line wraps at 56 columns. I think I'd want at least 100 columns for a good log viewing experience. Based on that, I don't think the inline log viewer is a good idea. The pop-up works fine.

Image

hexaglow (approved):
Looks right now.

I notice that the new logs button is only present in the System section, not the Workflows & Automations section (`/automations/all`). Consider adding this before merging, or as a follow-up PR.

PR #865, opened 2026-06-05
driver/steinel: support multiple devices per driver instance

deanredfern26:
[SCB-1007](https://vanti.youtrack.cloud/issue/SCB-1007) 

## Summary

Converts the Steinel HPD driver from one device per driver instance to a `devices` list, so sites with many sensors per controller don't need a driver service per sensor.

- Each device gets its own client, poller and fault check, following the shelly/trv pattern. Root level `passwordFile` and `pollInterval` act as shared defaults with per-device overrides.
- The existing single-device config (root level `ipAddress`) is still supported and is normalised into a one-element devices list named after the driver.
- A device that fails to set up is logged and skipped rather than failing the whole driver, unless every device fails — then the service reports the error.
- Reconfiguring waits for the previous devices to stop and release their health checks before re-registering them, so devices don't vanish on a config reload. Sensor HTTP requests honour the context and have a timeout so that wait can't hang on an unresponsive sensor.
- Each device registers its own health check proto — `healthpb` mutates the check it's given, so a shared instance would compound check ids.

## Breaking config changes

- **Plaintext passwords are rejected** with `plaintext passwords in config are not supported, use passwordFile`. Passwords must come from a non-empty password file; the hardcoded vendor default fallback (`Steinel123`) is removed.
- Root `metadata`/`udmiTopicPrefix` alongside `devices` is an error rather than silently ignored (they don't cascade; set them per device).
- The **UDMI trait is only announced when a device has a `udmiTopicPrefix`**, and prefixes must be unique between devices — otherwise every device would export to the same MQTT topic (`events/pointset`). A legacy single-device config that relied on an empty prefix loses the UDMI trait and needs a prefix set.

## Testing

- `go test -race -count=1 ./pkg/driver/steinel/...` — includes a new re-apply regression test (`TestDriver_applyConfig_reapply`) covering health-check ids and config reload, plus expanded `ParseConfig` coverage (defaults, cascades, all error cases).
- `go build ./...`, `go vet`, `gofmt` clean.

### DIFF
--- a/pkg/driver/steinel/hpd/driver.go
+++ b/pkg/driver/steinel/hpd/driver.go
@@ -2,9 +2,11 @@ package hpd
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"sync"
 
 	"go.uber.org/zap"
-	"golang.org/x/sync/errgroup"
 
 	"github.com/smart-core-os/sc-bos/pkg/driver"
 	"github.com/smart-core-os/sc-bos/pkg/driver/steinel/hpd/config"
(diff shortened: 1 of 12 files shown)


pkg/driver/steinel/hpd/driver.go
```diff
@@ -46,64 +48,86 @@ type Driver struct {
...
 
-	poller := newPoller(d.client, cfg.PollInterval.Duration, d.logger.Named("SteinelPoller").With(zap.String("ipAddress", cfg.IpAddress)), faultCheck, d.airQualitySensor, d.occupancy, d.temperature)
-
-	grp.Go(func() error {
-		poller.startPoll(ctx)
-		return nil
-	})
+	poller := newPoller(client, cfg.PollInterval.Or(config.DefaultPollInterval), logger.Named("SteinelPoller"), faultCheck, airQualitySensor, occupancy, temperature)
 
+	wg.Add(1)
```
dloveen-v: This can be simplified to `wg.Go(...)`

dloveen-v (approved):

PR #866, opened 2026-06-08
driver/bacnet: clear trait fault on clean poll

dloveen-v:
updateTraitFaultCheck wrote the failure into reliability.LastError via UpdateReliability, but LastError was never cleared and the success path's ClearFaults() didn't touch it, so trait faults latched after recovery.

Switch the failure path to SetFault so it pairs with the ClearFaults() already on the success path: ABNORMAL while failing, NORMAL on the next clean poll. Fixes the sticky fault for every trait routed through this helper (air_temperature, electric, meter, ...).


pkg/driver/bacnet/merge/trait.go
deanredfern26: `ctx` parameter is now unused, consider using _ or add a comment as to why it is there

deanredfern26 (approved):

PR #867, opened 2026-06-09
Feat/udmi bacnet compliance

dloveen-v:


pkg/auto/udmi/messages.go
```diff
@@ -36,6 +50,59 @@ type PointValue struct {
...
+
+// MetadataSystem is the system block of a metadata message (model_system.json);
+// all fields are optional, but a present location must carry a site.
+type MetadataSystem struct {
+	Name     string            `json:"name,omitempty"`
+	Location *MetadataLocation `json:"location,omitempty"`
+}
+
+// MetadataLocation describes where the device is installed; site is required
+// when location is present and must match ^[A-Z]{2}-[A-Z]{3,4}-[A-Z0-9]{2,9}$.
```
deanredfern26: this comment says `when present must match ...` but I can't see anywhere that this is validated.

pkg/driver/bacnet/merge/udmi.go
```diff
@@ -212,7 +307,19 @@ func (f *udmiMerge) pollPeer(ctx context.Context) error {
...
+
+	// Re-emit state and metadata every poll (not just on change/connect), so a
+	// point-in-time subscriber reliably sees them rather than only catching the
+	// pair published at stream connect. state reports current poll health, including
+	// operational=false when the device is unreachable, so it runs before the
+	// all-failed return below.
+	allFailed := len(errs) == len(f.config.Points)
+	f.operational.Store(!allFailed)
+	if f.config.EmitStateMetadata {
+		f.sendStateMetadata(ctx)
```
deanredfern26: in practice, sending metadata on every poll might prove to be wasteful. we can see how it goes, but other systems I have worked with that emit metadata do so at a less frequent rate as it is unlikely to change much.
also, the pointset data uses an `hasUpdate` flag to check if things have changed before publishing

deanredfern26 (approved):

pkg/auto/udmi/messages.go
dloveen-v: "Must match" due to the open UDMI standards. I will amend to say it's the callers responsibility.

PR #870, opened 2026-06-11
auto/udmi: make retention different per topic

dloveen-v:


pkg/auto/udmi/mqtt.go
```diff
@@ -30,6 +40,14 @@ func mqttPublisher(client mqtt.Client, qos byte, retained bool) Publisher {
 	})
 }
 
+// isEventTopic reports whether an MQTT topic is a UDMI event topic. Event topics
+// use an "/event" or "/events" segment (e.g. ".../event/pointset/points",
+// ".../events/pointset"); state ("/state") and metadata ("/metadata.json")
+// topics do not.
+func isEventTopic(topic string) bool {
+	return strings.Contains(topic, "/event")
```
deanredfern26, line 48: this is probably fine, but it is a little loose and would incorrectly return true if a site or device was called `event'. not sure if that is actually likely though

deanredfern26 (approved):

pkg/auto/udmi/mqtt.go
dloveen-v: > a site or device was called `event'.

That's not BDNS compliant 😆

PR #872, opened 2026-06-12
history/dataretention: raise health check on high storage utilisation

deanredfern26:
[SCB-1113](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1113) 

Link the DataRetention trait to the Health system so that store devices report a storage utilisation HealthCheck (normality HIGH) when disk usage approaches capacity.

The check is produced at the source, inside the existing poll loop. For SQLite the utilisation is the actual fullness of the filesystem containing the store (matching df's Use%, counting all files on the disk), since running out of space breaks the store regardless of who consumed it; the trait still reports the DB's own bytes.used against the disk's bytes.capacity. Postgres has no inherent capacity, so its check is only raised when the new PostgresConfig.MaxSizeBytes is configured, with utilisation measured against that quota. The high threshold is configurable via stores.Config.StorageHealthHighPercent (default 90%).

The health check makes the SQLite poll always-on, but counting rows is a full table scan, too expensive to repeat forever when nobody is watching. Polls therefore only refresh the row count at startup, when a PullDataRetention subscriber connects, and while one stays connected; unsubscribed polls just refresh the cheap size and disk fullness figures, keeping the last known count.

### DIFF
--- a/pkg/history/dataretention/storagehealth.go
+++ b/pkg/history/dataretention/storagehealth.go
@@ -1,0 +1,131 @@
+package dataretention
+
+import (
+	"context"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/smart-core-os/sc-bos/pkg/node"
+	"github.com/smart-core-os/sc-bos/pkg/proto/dataretentionpb"
+	"github.com/smart-core-os/sc-bos/pkg/proto/healthpb"
+)
+
+// storageHealth lazily maintains a BoundsCheck reporting disk utilisation for a store.
+// The check is created on the first update, so stores without capacity information
+// (e.g. SQLite on Windows, or Postgres with no configured max size) never call update
+// and so never announce a storage check.
+type storageHealth struct {
+	checks  *healthpb.Checks
+	name    string
+	highPct float32
+	logger  *zap.Logger
+
+	check *healthpb.BoundsCheck
+}
+
+// update reports the current disk utilisation (a percentage, 0-100) to the health check,
+// creating the check on first use.
+func (h *storageHealth) update(ctx context.Context, utilization float32) {
+	if h == nil || h.checks == nil {
+		return
+	}
+	if h.check == nil {
+		high := h.highPct
+		check, err := h.checks.NewBoundsCheck(h.name, &healthpb.HealthCheck{
+			Id:              "storage",
+			DisplayName:     "Storage Utilisation",
+			Description:     "Reports HIGH when disk utilisation approaches capacity.",
+			EquipmentImpact: healthpb.HealthCheck_FUNCTION,
+			Check: &healthpb.HealthCheck_Bounds_{
+				Bounds: &healthpb.HealthCheck_Bounds{
+					Expected: &healthpb.HealthCheck_Bounds_NormalRange{
+						NormalRange: &healthpb.HealthCheck_ValueRange{High: healthpb.FloatValue(float64(high))},
+					},
+					DisplayUnit: "%",
+				},
+			},
+		})
+		if err != nil {
+			h.logger.Warn("failed to create storage health check", zap.Error(err))
+			return
+		}
+		h.check = check
+	}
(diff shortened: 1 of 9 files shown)


pkg/history/dataretention/storagehealth.go
```diff
@@ -0,0 +1,131 @@
...
+// whenever a PullDataRetention subscriber connects, and on every poll while at least one
+// subscriber is connected; unsubscribed polls pass false so updates that nobody is
+// watching stay cheap.
+//
+// When alwaysPoll is false the loop idles (no polling) until a subscriber connects,
+// avoiding work when nobody is watching. When alwaysPoll is true the loop polls
+// unconditionally — used when a storage HealthCheck is an always-on consumer of the
+// polled data.
+func pollLoop(ctx context.Context, model *dataretentionpb.Model, alwaysPoll bool, update func(ctx context.Context, full bool)) {
+	tick := time.NewTicker(30 * time.Second)
```
hexaglow: `30 * time.Second` constant is used 3 times and presumably they are supposed to match. Extract to e.g. `const tickInterval = 30 * time.Second`

hexaglow (commented):
The SQLite health check calculates its utilisation percentage purely based on disk usage of the data volume, which is reasonable - writes will start failing when the disk is full. This is different to how the DataRetention trait works, which reports usage (of the DB only) and whole disk capacity.

These interact in quite a surprising way - if the disk is full of other data, then DataRetention will report a low usage but the health check may have a high disk usage percentage. I think that would be confusing to the user.

In my opinion, this indicates a deficiency in the DataRetention trait design. We can divide disk usage into 3 parts:
1. Free space
2. Space used by the thing we are measuring (the database)
3. Space used by other things sharing the same volume

The DataRetentionBytes message doesn't contain all this info, or enough to reconstruct it. 

Idea: add `other_used` to DataRetentionBytes to represent other data used on the disk. This allows the client to calculate the other relevant quantities:
- Disk fullness (what the health check measures) = `(used + other_used) / capacity`
- Free space = `capacity - used - other_used`
- Alternative fullness metric = `used / (capacity - other_used)` has the advantage of starting at 0 when the DB is empty but still reaching 100% when the disk is full

So what you've added here is good but I think these need to be consistent. Do you agree?

deanredfern26:
> The SQLite health check calculates its utilisation percentage purely based on disk usage of the data volume, which is reasonable - writes will start failing when the disk is full. This is different to how the DataRetention trait works, which reports usage (of the DB only) and whole disk capacity.
> 
> These interact in quite a surprising way - if the disk is full of other data, then DataRetention will report a low usage but the health check may have a high disk usage percentage. I think that would be confusing to the user.
> 
> In my opinion, this indicates a deficiency in the DataRetention trait design. We can divide disk usage into 3 parts:
> 
>     1. Free space
> 
>     2. Space used by the thing we are measuring (the database)
> 
>     3. Space used by other things sharing the same volume
> 
> 
> The DataRetentionBytes message doesn't contain all this info, or enough to reconstruct it.
> 
> Idea: add `other_used` to DataRetentionBytes to represent other data used on the disk. This allows the client to calculate the other relevant quantities:
> 
>     * Disk fullness (what the health check measures) = `(used + other_used) / capacity`
> 
>     * Free space = `capacity - used - other_used`
> 
>     * Alternative fullness metric = `used / (capacity - other_used)` has the advantage of starting at 0 when the DB is empty but still reaching 100% when the disk is full
> 
> 
> So what you've added here is good but I think these need to be consistent. Do you agree?

Yes I agree, we should expand the trait to give us as much clarity on disk usage metrics as possible. Having health info and getting alerted on high utilisation for our hardware before it becomes a problem is very useful, so the more info the better. If you have any more thoughts please chip in. I also haven't thought too much yet about getting useful info for this kind of stuff when it is being run on containers/volumes.

hexaglow (approved):
Sorry for the review delay, this one slipped through. Looks good.

PR #873, opened 2026-06-12
driver/sim: add coupled simulated building driver

deanredfern26:
[SCB-1049](https://vanti.youtrack.cloud/issue/SCB-1049) 

Add a building simulation driver (pkg/driver/sim): a physically-coupled building simulation configured as floors, rooms and device archetypes rather than an explicit device list. A single engine drives occupancy through the working day, which drives lighting and FCU load, from which electrical demand and metered energy are derived; daylight feeds brightness sensors and occupancy feeds CO2, motion and enter/leave.

Time can be accelerated (timeMultiplier) with per-tick state changes clamped so large steps stay stable. Occupancy uses smoothed targets and hysteresis so steady rooms don't manufacture enter/leave churn. Device names are numbered per location and type, and config validation rejects duplicate or unnamed floors and rooms.

Extract the mock driver's day-curve mechanism into pkg/util/timescale, shared by mock's NineToFive and sim's configurable WorkingHours curve.

Include a self-contained two-floor example (example/config/sim-building) with zones, ui-config and system config, plus run configurations and an ops UI env file for running it locally.

### DIFF
--- a/example/config/sim-building/app.conf.json
+++ b/example/config/sim-building/app.conf.json
@@ -1,0 +1,127 @@
+{
+  "name": "sim-building",
+  "metadata": {
+    "appearance": {
+      "title": "Simulated Building"
+    }
+  },
+  "drivers": [
+    {
+      "name": "sim/demo/driver",
+      "type": "sim",
+      "namePrefix": "sim/demo",
+      "timeMultiplier": 288,
+      "tickInterval": "5s",
+      "baseLoadW": 5000,
+      "workingHours": {"start": 8, "end": 18, "days": ["Mon", "Tue", "Wed", "Thu", "Fri"]},
+      "healthCheck": {"faultProbability": 0.05},
+      "floors": [
+        {
+          "name": "ground",
+          "title": "Ground Floor",
+          "maxOccupancy": 60,
+          "rooms": [
+            {
+              "name": "open-plan",
+              "title": "Open Plan",
+              "archetypes": [
+                {"type": "lighting", "count": 6},
+                {"type": "fcu", "count": 2, "setPointC": 21},
+                {"type": "pir"},
+                {"type": "brightness"},
+                {"type": "airquality"}
+              ]
+            },
+            {
+              "name": "meeting-1",
+              "title": "Meeting Room 1",
+              "maxOccupancy": 8,
+              "archetypes": [
+                {"type": "lighting", "count": 2},
+                {"type": "fcu", "setPointC": 21},
+                {"type": "pir"}
+              ]
+            }
+          ]
+        },
+        {
+          "name": "first",
+          "title": "First Floor",
+          "maxOccupancy": 40,
+          "rooms": [
+            {
+              "name": "open-plan",
+              "title": "Open Plan",
+              "archetypes": [
+                {"type": "lighting", "count": 4},
+                {"type": "fcu", "count": 2, "setPointC": 22},
+                {"type": "pir"},
+                {"type": "brightness"}
+              ]
+            },
+            {
+              "name": "meeting-1",
+              "title": "Meeting Room 1",
+              "maxOccupancy": 6,
+              "archetypes": [
+                {"type": "lighting", "count": 2},
+                {"type": "pir"}
+              ]
+            }
+          ]
+        }
+      ],
+      "buildingDevices": [
+        {"type": "meter", "title": "Main Incomer"},
+        {"type": "electric", "title": "Main Incomer"}
+      ]
+    }
+  ],
+  "zones": [
+    {
+      "name": "sim/demo/zones/ground",
+      "type": "area",
+      "metadata": {"appearance": {"title": "Ground Floor"}, "membership": {"subsystem": "zones"}},
+      "lights": [
+        "sim/demo/ground/open-plan/lighting-01", "sim/demo/ground/open-plan/lighting-02",
+        "sim/demo/ground/open-plan/lighting-03", "sim/demo/ground/open-plan/lighting-04",
+        "sim/demo/ground/open-plan/lighting-05", "sim/demo/ground/open-plan/lighting-06",
+        "sim/demo/ground/meeting-1/lighting-01", "sim/demo/ground/meeting-1/lighting-02"
+      ],
+      "thermostats": [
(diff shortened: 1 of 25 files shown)


pkg/driver/sim/archetype.go
```diff
@@ -0,0 +1,558 @@
...
+	"github.com/smart-core-os/sc-bos/pkg/proto/onoffpb"
+	"github.com/smart-core-os/sc-bos/pkg/proto/statuspb"
+	"github.com/smart-core-os/sc-bos/pkg/proto/typespb"
+	"github.com/smart-core-os/sc-bos/pkg/resource"
+	"github.com/smart-core-os/sc-bos/pkg/trait"
+)
+
+// Supported archetype types.
+const (
+	ArchetypeLighting   = "lighting"
```
dloveen-v, line 33: These Archetype strings should be SmartCore trait names? Unless I am missing something this is just duplicating an existing standard.

example/config/sim-building/app.conf.json
```diff
@@ -0,0 +1,127 @@
+{
```
dloveen-v, line 1: There are no history automations in this. One of the most fundamental use-cases for running a sim is seeing dashboard effects. Please amend the ui-config.json and app.conf.json to support rich visualization representations.

pkg/driver/sim/archetype.go
```diff
@@ -0,0 +1,558 @@
...
+				})
+			}
+			switch {
+			case !primed:
+				// Baseline event so the sensor reports totals before any movement.
+				emit(enterleavesensorpb.EnterLeaveEvent_DIRECTION_UNSPECIFIED)
+			default:
+				// One event per direction that changed this tick, so simultaneous
+				// enters and leaves (e.g. across rooms at building level) are both seen.
+				if enter != lastEnter {
```
dloveen-v: This is the only trait supporting gated update emissions. All other traits / archetypes should also gate the current emission with a verified change - if no change in current, do not emit. This should put less stress on history automations if they get added.

pkg/driver/sim/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Example
+
+A self-contained two-floor example lives at
+[`example/config/sim-building/app.conf.json`](../../../example/config/sim-building/app.conf.json),
+including zones so the data aggregates in the ops UI.
+
+## Notes
+
+- The simulation is the single source of truth; there is no per-device force/control API as in
```
dloveen-v: Is there a way we can delay the next tick on an API write call? I am just thinking of the efficacy of overwriting API writes.

pkg/driver/sim/archetype.go
```diff
@@ -0,0 +1,558 @@
...
+// buildArchetype constructs the trait servers and updaters for a single device of
+// the given archetype. room is nil for building-level aggregates.
+func buildArchetype(a config.Archetype, room *Room) (features []node.Feature, updaters []Updater, subsystem string, err error) {
+	// Room-scoped archetypes read per-room state; rejecting them here (rather than
+	// letting their updater dereference a nil room on the engine goroutine) keeps a
+	// misconfigured buildingDevices entry a config error instead of a panic.
+	if room == nil && roomScoped(a.Type) {
+		return nil, nil, "", fmt.Errorf("archetype %q is room-scoped and cannot be used as a building-level device", a.Type)
+	}
+	switch a.Type {
```
dloveen-v: It may be better to use a single Archectype descriptor registry than separate switch statements each time an iteration needs to occur.

```
  type archetypeDesc struct {
        Type            string
        DisplayName     string
        Subsystem       string
        RoomScoped      bool
        OccupancyDriven bool
        Build           func(a config.Archetype, room *Room) ([]node.Feature, []Updater)
  }

  registered in one var archetypeList = []archetypeDesc{...} (panic on duplicate Type at init). Then buildArchetype becomes a map lookup + the existing nil-room guard, and roomScoped/occupancyDriven/displayName become one-line field reads.
```

pkg/driver/sim/README.md
deanredfern26: I was thinking about this comment and I have just wired this sim up to the MockApi now in `control`. We should be able to now force values to show devices misbehaving etc. and see how that effects energy use or other things.

pkg/driver/sim/archetype.go
deanredfern26: archetypes are not 1:1 with traits, e.g. fcu is a composite of AirTemperature + FanSpeed + OnOff
archetypes are more like typical devices

dloveen-v: Wouldn't it make more sense to support multiple combinations of SmartCore traits? Just in terms of parsing / readability does it make more sense?

deanredfern26: I think for things like AirQualitySensor then you're right, having trait names would be better. For simpler devices it would be easy enough.

I just don't know how that would work with more complex devices like an FCU. In multiple traits this would be something like `["AirTemperature","FanSpeed","OnOff"]` but this could also represent a fridge or other things, which have different behaviour profiles and are not linked as much to occupancy / time etc.

dloveen-v: I get what you mean but a fridge isn't a trait - it's a type of device. I'd be more content if an "Archetype" represented a set of SmartCore traits (with potential for duplication under different archetype names) than a set of strings that is constrained to only this driver.

I am picturing a scenario where "archetypes" are sent to a UI to configure the simulation outside of JSON config. Currently, there is not a SmartCore-oriented concept of Archetype. If we define an archetype as a "named device type that supports multiple unique SmartCore traits" then it makes more sense to define it this way so we can extend the sim driver functionality later.

deanredfern26: OK thanks fir the example, I think I understand now, have made the change.

example/config/sim-building/ui-config.json
```diff
@@ -0,0 +1,240 @@
+{
```
dloveen-v, line 1: It looks like one thing that isn't currently visualized or shown are the health checks the sim has set up.

dloveen-v (approved):

PR #874, opened 2026-06-15
fix(proxy): proxy full device metadata, not just traits

deanredfern26:
Previously, proxied devices only had the auto-generated trait list announced, leaving appearance, location, membership and device_type empty after proxying.
Announce the remote device's full metadata via node.HasMetadata (as the gateway does),


pkg/driver/proxy/driver.go
dloveen-v, line 145: LGTM. Can you update this comment too?

dloveen-v (approved):

PR #877, opened 2026-06-24
gateway: aggregate Log trait across the cohort

deanredfern26:
[SCB-974](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-974) Implement "log" trait for the gateway

Expose a cohort-wide aggregated Log trait on the gateway under the device name "logs", merging the PullLogMessages streams of every cohort node into one and stamping each message with its source node. The aggregate also includes the gateway's own node, which is not itself a cohort member, so the feed covers the node an operator is connected to.

The Ops UI logs page gains an "All nodes" option (shown only when connected to a gateway) that streams the aggregate with a per-line source column. Per-node log level and download controls are disabled for the aggregate, since those remain per-node operations reachable via the gateway proxy.


pkg/system/gateway/logaggregator.go
```diff
@@ -0,0 +1,349 @@
+package gateway
```
dloveen-v, line 1: File should be called `logaggregator` or something more specific.

proto/smartcore/bos/log/v1/log.proto
```diff
@@ -65,6 +65,9 @@ message LogMessage {
   SourceLocation source_location = 6;
   // Stack trace at the point of logging. Omitted when unavailable.
   string stack_trace = 7;
+  // Name of the node this entry originated from. Set by aggregators (e.g. the
+  // gateway) that merge logs from multiple nodes; empty for single-node streams.
+  string source = 8;
```
dloveen-v: May be worth renaming this to `source_node`? It isn't clear what's an acceptable `source` given there is `source_location` too.

dloveen-v (approved):

PR #885, opened 2026-06-26
feat(auto): add setpointhealth automation

deanredfern26:
[SCB-1208](https://vanti.youtrack.cloud/issue/SCB-1208)
Add a setpointhealth auto that raises a health check when a measured value fails to track its set point for a sustained period (e.g. a fan coil unit whose return air stays at 24C an hour after the set point dropped to 18C indicates a cooling fault).

It compares two values (measured vs set point) read from the same trait resource and uses a fault check with an on-delay timer: the check only trips once abs(measured - setPoint) exceeds the configured tolerance continuously, against the same set point, for the configured duration. The countdown restarts on a set point change (fresh window to reach the new target) and clears when the deviation returns within tolerance; data gaps report unreliable and freeze the countdown rather than tripping.

To share trait-reading plumbing with healthbounds:
- move anytrait to pkg/auto/internal/anytrait so both autos can import it
- lift pathFieldsAreSet into protopath2.FieldsAreSet
- share resourceFetcher via anytrait.Resource as a Fetcher method, so healthbounds and setpointhealth no longer carry byte-identical copies
image


pkg/auto/setpointhealth/README.md
```diff
@@ -0,0 +1,84 @@
...
+3. Starts an on-delay timer when the deviation exceeds `tolerance`.
+4. Marks the health check abnormal (a fault) once the deviation has continuously exceeded the
+   tolerance, against the same set point, for `duration`.
+5. Clears the fault when the deviation returns within `tolerance`.
+
+### Timer behaviour
+
+- The countdown is **reset when the deviation returns within tolerance** — only a continuous breach
+  trips the check.
+- The countdown is **restarted when the set point changes** — the equipment is given a fresh window
```
dloveen-v, line 30: I am thinking about this... if someone waits until the target set point is reached and re-sets the set point in attempts to force the temperature to change quickly, this will mask the issue in SmartCore as the timer will be reset.

Is it better to ignore new set points when running down the timer? I can see why it's this way but I am thinking about the actual practicality of resetting the timer.

deanredfern26: I was unsure how to handle this, so just went with the simplest way for the initial plan. You're right though resetting the timer could easily mask faults.

I have added the option of a 'backstop' timer, which will raise an error when the deviation is outside the tolerance after a given duration. Have a look, I think this is enough to catch the cases with changing goalposts

pkg/auto/setpointhealth/config/root.go
```diff
@@ -0,0 +1,206 @@
...
+	// Duration is how long the deviation must continuously exceed Tolerance before the check goes
+	// abnormal. The countdown restarts when the set point changes (the equipment gets a fresh window
+	// to reach the new target) and is cleared when the deviation returns within Tolerance.
+	// Must be greater than zero.
+	Duration jsontypes.Duration `json:"duration"`
+	// MaxDuration, when set, is an absolute backstop: the check goes abnormal once the deviation has
+	// exceeded Tolerance continuously for this long, regardless of how often the set point changes.
+	// It catches equipment that never tracks while its set point is repeatedly adjusted, which would
+	// otherwise keep restarting the Duration countdown and never trip. Unset (zero) disables the
+	// backstop. Must be greater than Duration when set.
```
dloveen-v, line 40: Is it worth mentioning neither or both of Duration and MaxDuration need to be set?

pkg/auto/setpointhealth/worker.go
```diff
@@ -0,0 +1,233 @@
...
+	tolerance   float64
+	duration    time.Duration
+	maxDuration time.Duration // zero disables the backstop
+	device      string
+	logger      *zap.Logger
+}
+
+func (w *worker) run(ctx context.Context, changes <-chan anytrait.Value) error {
+	var timer *time.Timer
+	var timerC <-chan time.Time
```
dloveen-v, line 52: nit picking here but I am not sure why the chan time.Time references need to be separated variables. It makes just as much if not more sense to reference using timer?

dloveen-v (approved):

PR #889, opened 2026-06-29
ui/ops: add log view to automations sidebar

deanredfern26:
[SCB-1367](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1367) 
Render the existing generic LogsCard in the Workflows & Automations sidebar so per-automation logs can be viewed there, matching the System section sidebar.

image


dloveen-v (commented):
Does the `LogsCard` automatically filter for logs belonging to the automation kind or name?

deanredfern26:
> Does the `LogsCard` automatically filter for logs belonging to the automation kind or name?

Yes it grabs data from `useSidebarStore`

dloveen-v (approved):

PR #890, opened 2026-06-29
Authenticate to SCC with X.509 mutual-TLS device certificates

mattnathan:
## What

Replaces the OAuth2 client-credentials flow for BOS ↔ Smart Core Connect cloud authentication with **mutual-TLS X.509 client certificates**.

## How it works

- **Enrolment** — BOS generates an EC P-256 keypair and PKCS#10 CSR locally and exchanges it at `/v1/device/register` (authorised by a one-time enrolment code as a Bearer token) for a signed certificate chain. The private key never leaves the controller.
- **Authentication** — the certificate is presented over mutual TLS on every check-in (`/v1/device/check-in`) and renewal.
- **Renewal** — BOS auto-renews at roughly two-thirds of the certificate lifetime, hot-swapping the new certificate in place with no restart. A manual `RenewCloudConnection` RPC is also exposed (surfaced as a "Renew now" action in the Ops UI).
- **Chains** — the full issued chain is stored and presented, so SCC issuing through an intermediate CA works (verified against real SCC).

## Surface

- **proto** — `CloudConnection` reworked: `node_id`/`api_endpoint`, enrolment-code-only registration, `RenewCloudConnection`, and certificate issued/expiry/next-renewal timestamps.
- **`internal/util/pki`** — EC P-256 key + PKCS#10 CSR helpers.
- **`internal/cloud`** — credential type and file-backed store, mTLS HTTP client, register/renew, and the auto-renew loop.
- **`internal/cloud/sim` + `cmd/cloudsim`** — the reference SCC server issues X.509 credentials (dev CA, primary/secondary credential slots); token issuance and secret-based node auth removed.
- **`internal/opsapi`, `pkg/app`** — wiring; **Ops UI** — enrolment-code link dialog with certificate details and renewal.

## Testing

- Credential round-trip, register/renew against the in-memory sim, auto-renew timer arithmetic and re-arm, and an intermediate-chain mTLS handshake (presenting the intermediate is required; the negative control confirms it).
- Manual end-to-end against the real SCC bos-api through its dev mTLS proxy: enrol → check-in → renew, including SCC's new intermediate-CA chain.

## Notes

- Commits are split so each builds on its own: the standalone `pki` helper, then the atomic cutover (proto, cloud client, sim, opsapi, app and UI change in lockstep).
- `example/config/vanti-ugs/system.conf.json` is intentionally **not** included — its cloud block points at a local dev proxy with `insecureSkipVerify`.


mattnathan:
Companion SCC-side PR: smart-core-os/smart-core-connect#895 — the Smart Core Connect bos-api that issues and validates the X.509 device certificates this change authenticates with. Verified end-to-end against it: enrol → check-in → renew, including SCC's intermediate-CA chain.

internal/cloud/sim/api_test.go
```diff
@@ -265,6 +270,17 @@ func newTestServer(t *testing.T) *httptest.Server {
 }
 
 func newTestServerWithStore(t *testing.T) (*httptest.Server, *store.Store) {
```
hexaglow: The only purpose of this helper is now to discard a single result parameter. Given that it's a private function, I would prefer just one function returning all 3, with usages refactored accordingly, and I suggest that it be called `newTestServerWithStore`

internal/cloud/sim/api_test.go
```diff
@@ -276,10 +292,81 @@ func newTestServerWithStore(t *testing.T) (*httptest.Server, *store.Store) {
...
 	t.Cleanup(testServer.Close)
 
-	return testServer, s
+	return testServer, s, apiServer
+}
+
+// enrollClient enrolls a node via the CSR exchange using the given enrollment
+// code, then returns an HTTP client that presents the issued certificate over
+// mTLS (for calling the device check-in / renew endpoints).
+func enrollClient(t *testing.T, ts *httptest.Server, apiServer *Server, code string) *http.Client {
```
hexaglow: Used once, doesn't even make the code shorter really. Inline it.

internal/cloud/sim/auth.go
```diff
@@ -0,0 +1,83 @@
...
+	dev, err := s.authenticateDevice(r)
+	if err != nil {
+		return 0, err
+	}
+	return dev.nodeID, nil
+}
+
+// authenticateDevice validates the request's client certificate against live
+// credential state. The TLS layer (ClientCAs + VerifyClientCertIfGiven) has
+// already verified the chain to the Connect CA and the validity window; here we
```
hexaglow: Reference to Connect here is misleading - cloudsim is a different implementation. Might want to rephrase this doc comment too, it was easier to determine what the function does by reading the code than the comment.

internal/cloud/sim/check_in_test.go
```diff
@@ -267,53 +246,57 @@ func TestCheckIn_CompletedAndFailedExcluded(t *testing.T) {
...
+	// not authenticate: either the TLS handshake is rejected, or the guard returns
+	// 401. Either way it must never succeed.
+	resp, err := bogus.Do(req)
 	if err != nil {
-		t.Fatalf("request failed: %v", err)
+		return // handshake rejected
 	}
 	defer func() { _ = resp.Body.Close() }()
-	assertStatus(t, resp, http.StatusUnauthorized)
+	if resp.StatusCode/100 == 2 {
```
hexaglow: This guard is looser than it could be. I would expect specifically `http.StatusUnauthorized` here. Is there a reason it could return another code?

internal/cloud/sim/register_test.go
```diff
@@ -85,97 +84,115 @@ func TestDeviceRegister(t *testing.T) {
...
+func doRegister(t *testing.T, client *http.Client, base, code, cn string) (*http.Response, []*x509.Certificate) {
 	t.Helper()
+	return doRegisterReq(t, client, deviceRegisterURL(base), code, cn)
+}
 
-	body, err := json.Marshal(map[string]string{"client_name": clientName})
+// doRegisterReq builds a CSR for cn and POSTs it (base64 DER, application/pkcs10)
+// to url, optionally with a bearer token. It returns the response and the parsed
+// PEM certificate chain (nil on non-2xx).
+func doRegisterReq(t *testing.T, client *http.Client, url, bearer, cn string) (*http.Response, []*x509.Certificate) {
```
hexaglow, line 146: Style nit: don't need a new helper, can just make doRegister take the URL directly. There's only 4 uses to adjust.

internal/cloud/autorenew.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	drain := func() {
+		if !timer.Stop() {
+			select {
+			case <-timer.C:
+			default:
+			}
+		}
+	}
+	arm := func(cred *Credential) {
```
hexaglow: Draining a timer channel after calling `Stop` has been unnecessary since Go 1.23

internal/cloud/client.go
```diff
@@ -3,71 +3,160 @@ package cloud
...
+	// without reconnecting or restarting (hot reload).
+	SetCredential(cred *Credential)
+}
+
+// credHolder holds the current credential behind a mutex so the cert presented
+// on the mTLS connection can be hot-swapped on renewal.
+type credHolder struct {
+	mu   sync.Mutex
+	cred *Credential
 }
```
hexaglow: Can be replaced by `atomic.Pointer[Credential]`

internal/cloud/client.go
```diff
@@ -145,16 +247,21 @@ func (c *HTTPClient) DownloadPayload(ctx context.Context, url string) (io.ReadCl
...
 	u, err := url.Parse(urlStr)
 	if err != nil {
 		return false
 	}
-	baseURL, err := url.Parse(c.checkInEndpoint)
-	if err != nil {
-		return false
+	return u.Scheme+"://"+u.Host == c.apiOrigin
+}
+
```
hexaglow, line 293: This implementation is less robust than the previous one (e.g. to a trailing slash). Code currently provides correct values, but I don't see any reason to change this method anyway.

internal/cloud/conn.go
```diff
@@ -79,90 +94,159 @@ type Conn struct {
 	bus minibus.Bus[ConnState]
 }
 
-// OpenConn creates a new Conn backed by the given stores.
-// depStore is used to construct a DeploymentUpdater whenever a Registration is loaded or set.
-func OpenConn(ctx context.Context, regStore RegistrationStore, depStore *DeploymentStore, opts ...ConnOption) (*Conn, error) {
+// OpenConn creates a new Conn backed by the given stores. baseURL is any URL on
+// the SCC API origin (e.g. the configured register URL); the check-in and renew
+// endpoints are derived from it. depStore is used to construct a DeploymentUpdater
```
hexaglow: Derived how? As implemented, the baseURL is passed to NewHTTPClient which truncates the path. This is a pretty important detail so should be documented here. Although I'd prefer this restriction to be removed (see comment on NewHTTPClient).

internal/cloud/client.go
```diff
@@ -3,71 +3,160 @@ package cloud
...
+// cred is the current credential (may be nil until enrolled).
+func NewHTTPClient(cred *Credential, baseURL string, opts ...HTTPClientOption) *HTTPClient {
+	cfg := &httpClientConfig{}
 	for _, opt := range opts {
-		opt(c)
+		opt(cfg)
+	}
+
+	origin := OriginOf(baseURL)
+	c := &HTTPClient{
```
hexaglow: Taking a `baseURL` but ignoring a path component, although documented here, seems like a footgun. There doesn't seem to be anything that would prevent HTTPClient from operating at a non-root baseURL, so why enfore it? Sure, in production we only intend to use a root URL as the base, but that's not a necessary constraint in this type. And if it is actually important, better to check that baseURL is a root URL rather than silently truncating.

internal/cloud/register_test.go
```diff
@@ -103,19 +119,20 @@ func TestRegister(t *testing.T) {
...
-	// Pin the API contract: the sim (and real SCC) must return the "invalid_enrollment_code"
-	// error code so that IsInvalidEnrollmentCode remains the correct predicate and the UI
-	// can map it to a friendly message.
-	t.Run("invalid enrollment code satisfies IsInvalidEnrollmentCode", func(t *testing.T) {
-		tsURL, client, _ := setupRegisterEnv(t)
+	// Pin the API contract: a rejected code is a generic 401 (SCC returns
+	// {error:"unauthorized"}), which the register endpoint's only-auth-is-the-code
+	// rule lets opsapi classify as an invalid enrollment code. IsInvalidCredentialsError
+	// is the predicate that holds against both the sim and real SCC.
+	t.Run("rejected enrollment code is a 401", func(t *testing.T) {
```
hexaglow: This just seems like a less-specific version of the test above now.

internal/cloud/sim/ca.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// credentialIDFromCert extracts the credentialId from a device leaf's URI SANs,
+// returning "" if absent.
+func credentialIDFromCert(cert *x509.Certificate) string {
+	for _, u := range cert.URIs {
+		if s := u.String(); strings.HasPrefix(s, credentialURIPrefix) {
+			return strings.TrimPrefix(s, credentialURIPrefix)
+		}
+	}
```
hexaglow: Can use `strings.CutPrefix`

internal/cloud/sim/ca.go
```diff
@@ -0,0 +1,170 @@
...
+	return ""
+}
+
+// leafFingerprint returns the hex-encoded SHA-256 of a certificate's DER, used
```
hexaglow: Return value is not hex-encoded

internal/cloud/credential_store.go
```diff
@@ -0,0 +1,141 @@
...
+	keyPEM, err := pki.EncodePrivateKey(cred.Key)
+	if err != nil {
+		return fmt.Errorf("encode cloud key: %w", err)
+	}
+	certPEM := pki.EncodeCertificates(cred.Chain)
+
+	if err := os.MkdirAll(filepath.Dir(s.keyPath), 0700); err != nil {
+		return fmt.Errorf("create cloud credential dir: %w", err)
+	}
+	if err := atomicWrite(s.keyPath, keyPEM, 0600); err != nil {
```
hexaglow: There's nothing in NewFileCredentialStore to enforce that keyPath and certPath are in the same directory, so we need to Mkdirall the certPath directory too.

internal/cloud/credential_store.go
```diff
@@ -0,0 +1,141 @@
...
+			errs = errors.Join(errs, err)
+		}
+	}
+	return errs
+}
+
+// atomicWrite writes data to path via a temp file in the same directory followed
+// by a rename, so a reader never observes a partially-written file. The file is
+// created with perm.
+func atomicWrite(path string, data []byte, perm os.FileMode) error {
```
hexaglow: The module already requires `github.com/google/renameio/v2` which can do this sort of thing.

internal/cloud/credential_store.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	cred, err := newCredential(key, chain)
+	if err != nil {
+		return nil, false, fmt.Errorf("load cloud credential: %w", err)
+	}
+	return cred, true, nil
+}
+
+func (s *fileCredentialStore) Save(_ context.Context, cred *Credential) error {
```
hexaglow: There's an edge case here: if you Save twice in a row without an intermediate Clear, and the process is interrupted after the keyPath save, then the key and cert on disk won't match which is going to cause problems (worth analysing what would happen).

Could be circumvented by just calling Clear at the start of Save.

pkg/app/sysconf/config.go
```diff
@@ -348,9 +348,23 @@ func (c *Certs) FillDefaults() *Certs {
...
+	// origin (scheme+host) is also where the check-in and renewal endpoints live.
 	RegisterURL string `json:"registerUrl,omitempty"`
 
+	// KeyFile and CertFile are where the controller's cloud client certificate
+	// material (EC P-256 private key and the Connect-CA-issued chain, both PEM)
+	// are stored. Relative paths are resolved under the cloud data directory.
+	// Default to "cloud.key.pem" / "cloud.cert.pem".
+	KeyFile  string `json:"keyFile,omitempty"`
+	CertFile string `json:"certFile,omitempty"`
+
```
hexaglow: Why do these need to be configurable?

hexaglow (changes requested):
Creating a new `Credential` type instead of reusing the existing `Registration` created a lot of code churn which made
the diff large, but the code would be just as clear either way. In my opinion, keep the Registration.

## Overridden `register_url` not used consistently
When calling the RegisterCloudConnection RPC, `register_url` can be used to specify what URL to contact the
registration server at. This URL is used to register, and is saved in `Conn.baseURL`. The URL is used derive URLs
for check-in and renewal. However, it is not persisted with the credential. When the system restarts, the `register_url` 
is lost, so when the system tries to renew the credential it will use the default URL (from the config file) instead
of the URL that was actually used to issue the credential. Therefore, it contacts the wrong server for renewal of the
saved credential. The `register_url` should be persisted along with the credential.

Previously, the `registration.json` file saved the necessary info.

Similarly, in the `opsapi` server, `CloudConnection.ApiEndpoint` is computed based on the default API endpoint. 
If the current connection was established with a custom `register_url`, then the `ApiEndpoint` will be wrong. 

Some parts of the system assume that the URLs can be only be changed in the config file, while others assume it can
vary on a per-enrollment basis. The previous implementation allowed a per-enrollment override and carried that through,
including persisting it on disk so it survives restarts. Either support this feature properly, or remove it and document
that only the URL from the cloud config can be used.

## Design of the CSR protocol

The design of the endpoints at /v1/device/register is similar to that described in RFC 7030 (Enrollment over Secure 
Transport). This is implied by the acronym 'EST' which appears in a few places. The implementation is incompatible with
the requirements of the RFC in several ways, including:
- Deviation from the URLs required by the standard
- Lack of a `cacerts` endpoint
- Use of an `application/pem-certificate-chain` encoded response instead of `application/pkcs7-mime` required by the
  standard.

EST uses a nonstandard HTTP header `Content-Transfer-Encoding` (normally only used in MIME) which is surprising for 
a reader not familiar with that RFC. Given that the implementation is already incompatible with actual EST
implementations in other ways, I would prefer if this header is not used, as it is not technically necessary, and also
that any references to EST in comments make it clear that the protocol is merely inspired by EST, not an implementation.

The changes remove the ability for the server to specify a different origin for the BOS API. Do you propose to remove 
this ability permanently?

## Cloudsim
Now that cloudsim uses HTTPS instead of HTTP, consider changing the default port from 8080 to something more suggestive
of HTTPS.

mattnathan:
Thanks for the thorough review — all addressed. I've force-pushed (so the inline threads show as outdated); rundown below, and I've resolved the threads that are fixed.

**Substantive**
- **register_url persistence (#1):** `Registration` now carries the API endpoint, set at enrolment (`OriginOf(registerURL)`) and persisted in `registration.json`, so check-in/renew reach the issuing origin across restarts. The reported `ApiEndpoint` now reflects the registration's actual endpoint (falling back to the config default only pre-enrolment). This also answers your "remove the server-specified origin permanently?" question — no, it's now supported properly (per-enrolment, persisted).
- **Reverted to `Registration` (#2):** dropped the `Credential` type; the type lives in `client.go` as before (not a new file), evolving in place from the OAuth struct to the X.509 one. Kept the endpoint field as `APIEndpoint` rather than `BosapiRoot` since it's now a locally-derived origin, not the server-returned `bosapi_root` field (which no longer exists).
- **EST (#3):** dropped the `Content-Transfer-Encoding` header and reworded all references to "inspired by EST (RFC 7030), not an implementation." NB this needs bos-api (#895) to accept `application/pkcs10` without the CTE header.
- **cloudsim port (#4):** default → `:8443`.

**Store**
- Now a single atomic `registration.json` (via `renameio`), which removes the two-file key/cert mismatch window, the separate-dir `MkdirAll` gap, and the hand-rolled temp+rename. `KeyFile`/`CertFile` config removed — there's one file now.

**Cleanups**
- `atomic.Pointer` for the cert holder; dropped the Go-1.23 timer drain; `strings.CutPrefix`; fixed the "hex-encoded" comment (it returns raw bytes); `auth.go` reworded (sim's dev CA, not Connect).
- `NewHTTPClient` is documented to take an API origin (no silent path truncation); `OpenConn` derives that origin from the configured register URL — that's the "derived how?" answer.
- Tests: consolidated the server helper (`newTestServerWithStore` returns all three), inlined the single-use `enrollClient` and `doRegister`, merged the redundant register test, and the check-in test now asserts **401** specifically — with a comment on why it's 401 and not a handshake rejection (the self-signed cert doesn't match the server's advertised client-CAs, so the TLS stack omits it and the guard 401s).

Re-requesting review — thanks again.

internal/util/pki/csr.go
```diff
@@ -0,0 +1,53 @@
...
+	return ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+}
+
+// CreateCSRDER builds a DER-encoded PKCS#10 certificate signing request for key,
+// carrying commonName in the subject. Any dnsSANs are added as DNS subject
+// alternative names.
+//
+// The result is raw DER (the wire format for application/pkcs10); callers
+// transferring it over HTTP base64-encode it themselves
+// (Content-Transfer-Encoding: base64). The signature algorithm is derived from
```
hexaglow: Stray reference to `Content-Transfer-Encoding`

cmd/cloudsim/main.go
```diff
@@ -23,7 +24,7 @@ var (
 )
 
 func init() {
-	flag.StringVar(&flagListen, "listen", "127.0.0.1:8080", "interface:port to listen on")
+	flag.StringVar(&flagListen, "listen", "127.0.0.1:8443", "interface:port to listen on (serves HTTPS)")
```
hexaglow: Port 8443 clashes with e.g. the vanti-ugs example for BOS, which means a developer is likely to need to change it. Please find a port that doesn't clash with any of our other default ports.

internal/cloud/autorenew.go
```diff
@@ -0,0 +1,117 @@
...
+func AutoRenew(ctx context.Context, conn *Conn, logger *zap.Logger) {
+	initial, changes := conn.PullState(ctx)
+	changes = concurrent.BreakBackpressure(changes)
+	cur := initial.Registration
+
+	timer := time.NewTimer(time.Hour)
+	timer.Stop()
+	// Since Go 1.23 a Stop/Reset reliably discards any pending fire, so there is
+	// no need to drain timer.C before re-arming.
+	arm := func(reg *Registration) {
```
hexaglow: I don't think we need such comments, it's implied by the Go version in `go.mod`

internal/cloud/sim/store/store/migrations/0004_credentials.sql
```diff
@@ -0,0 +1,35 @@
...
+-- opaque marker minted per enrollment-code exchange (stable across renewals) and
+-- is what revocation acts on.
+
+DROP INDEX nodes_secret_hash;
+ALTER TABLE nodes DROP COLUMN secret_hash;
+
+CREATE TABLE credentials (
+    id              INTEGER PRIMARY KEY,
+    node_id         INTEGER NOT NULL,
+    credential_id   TEXT NOT NULL,    -- opaque marker, minted per code exchange, carried in the cert SAN
```
hexaglow, line 13: Nit: On second viewing, having both an `id` and `credential_id` column feels a bit confusing, although the comment does explain the difference sufficiently. 

Could normalise into `credentials` + `credential_versions` or similar.

It's OK as is to be honest, given that it's just a development tool, but thought I'd raise.

hexaglow (changes requested):
Looking good now, couple of things need tidying up.

mattnathan:
Round two, thanks:
- **csr.go** — removed the stray `Content-Transfer-Encoding` reference (missed it first time).
- **cloudsim port** — good catch, `:8443` collides with BOS's own HTTPS default; moved to `:9443` (checked it doesn't clash with any of our other defaults).
- **autorenew** — dropped the Go-1.23 timer comment.
- **credentials migration `id`/`credential_id`** — leaving as-is per your note (dev-only sim store); agree it's a reasonable future normalisation if the sim ever grows.

Re-requesting.

hexaglow (approved):
Lint failures are unrelated to the changes in this PR, will investigate later

PR #892, opened 2026-06-30
driver/steinel: detect missing sensor 

deanredfern26:
A connected base whose sensor module is missing, responds to GET /sensor with code 200 and empty body {}, which previously unmarshalled to a zero-value response and was treated as healthy. Flag an absent SensorName as a bad response instead.

Also fix the malformed-JSON check, which used errors.Is against a nil *json.UnmarshalTypeError and never matched; use errors.As and also cover *json.SyntaxError so unparseable bodies report BAD_RESPONSE.

### DIFF
--- a/pkg/driver/steinel/hpd/poller.go
+++ b/pkg/driver/steinel/hpd/poller.go
@@ -55,12 +55,25 @@ func (p *poller) process(ctx context.Context) {
 			return // we're stopping, not a device fault
 		}
 		h := noResponse
-		var unsupportedTypeErr *json.UnmarshalTypeError
-		if errors.Is(err, unsupportedTypeErr) {
+		var (
+			typeErr   *json.UnmarshalTypeError
+			syntaxErr *json.SyntaxError
+		)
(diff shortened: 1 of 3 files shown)


pkg/driver/steinel/hpd/poller.go
```diff
@@ -55,15 +55,28 @@ func (p *poller) process(ctx context.Context) {
...
 		p.logger.Error("failed to GET sensor", zap.Error(err))
 		return
 	}
 
+	if response.SensorName == "" {
+		// The base can answer 200 with an empty body ({}) when it is connected but the
+		// sensor module itself is missing. Comms are fine, but the response carries no
+		// sensor data, so treat it as a bad response rather than healthy.
+		p.faultCheck.UpdateReliability(ctx, badResponse)
+		p.logger.Error("sensor returned no data, the sensor module may be missing or disconnected")
```
dloveen-v: The error log is carrying very useful information that isn't in the `badResponse` error. Can we include it in the `Reliability` message?

dloveen-v (approved):

PR #895, opened 2026-06-30
Add supervisor to install updates from the cloud

hexaglow:
Add a new supervisor process (in `supervisor/`) that exposes a gRPC service allowing BOS to update itself. This PR includes support for BOS installations running in Podman. The Supervisor itself is designed to be installed directly on the host.

Not included in this PR, but planned:
- FreeBSD support
- Supervisor self-update support

The supervisor is located in `supervisor/` because, during development, I wasn't sure if it belonged in its own repo, or tightly integrated with this one. I'd like your thoughts on this, so I can relocate it accordingly, to 1) `cmd/supervisor`; or 2) its own repo. It will, at the very least, have a different release cycle to BOS itself.

BOS is still responsible for communicating with the Smart Core Connect server, and discovers available updates similar to config versions. When a new version needs installing, BOS will pass along the URL and other details to the Supervisor, which downloads the new version and makes Podman run it.

The Supervisor supports automatic rollback. BOS is expected to call a `Commit` RPC within a certain deadline of restarting after an upgrade. If this doesn't happen, the Supervisor will roll it back to the previous version. BOS will not call Commit until its cloud connection has been established, ensuring that updates which break the cloud connection are automatically rolled back. BOS also interrogates the supervisor to see if updates succeeded or failed, and reports this to SCC via check-in.

Cloudsim has been extended to support the new functionality on the check-in endpoint. It can now store and serve update artefacts, much like config versions. Nodes now have a platform string associated with them.

For testing out the code, there's some demo configs and scripts in `supervisor/demo/`. This shows our new planned deployment scheme - with Podman quadlets (generated systemd units) replacing our typical podman-compose setup.


pkg/app/version.go
```diff
@@ -21,6 +22,20 @@ type VersionInfo struct {
 	*debug.BuildInfo
 }
 
+// EffectiveVersion returns the version string BOS reports to the Supervisor (via Commit) and
+// compares against an update target to distinguish a successful update from a rollback.
+// It is the build-info main module version, overridable for development only by the
+// BOS_VERSION_OVERRIDE environment variable.
+func EffectiveVersion() string {
+	if override := os.Getenv("BOS_VERSION_OVERRIDE"); override != "" {
```
mattnathan, line 41: 🔴 **How are production images versioned, and what guarantees `EffectiveVersion()` equals the version string the cloud puts on the artefact?**

Tracing it end-to-end with the in-repo build: `Dockerfile:48` runs a plain `go build` with no `-ldflags`, and the image sets no `BOS_VERSION_OVERRIDE` (the `GIT_VERSION` arg is wired only into the UI stage). So `EffectiveVersion()` falls through to `BuildInfo.Main.Version`, which for a plain `go build` of the main module is `(devel)` — `-ldflags` cannot set `Main.Version`; only `go install module@version` does.

The supervisor confirms an update only when `awaitCommit(t.Version)` sees `lastCommit == t.Version` (`supervisor/internal/server/server.go:208,248`), where `t.Version` is the cloud's artefact version (e.g. `v1.2.3`). But the updated BOS calls `Commit("(devel)")` via `RunStartupCommit`. `"(devel)" != "v1.2.3"` → never confirms → times out after `commitDeadline` → **rolls back a healthy update, every time.**

The demo works only because `01-build.sh` bakes `BOS_VERSION_OVERRIDE=$v` per image (and asserts it). Two things follow: (1) production needs an equivalent stamp that byte-matches the cloud's artefact version; (2) the comment on L27-28 calling the override "development only" is backwards — it's the thing making updates confirm. If there's an out-of-tree release pipeline that stamps this, point me at it and I'll withdraw; otherwise I'd treat this as a merge blocker.

internal/cloud/conn.go
```diff
@@ -216,9 +255,34 @@ func (c *Conn) Update(ctx context.Context) (needReboot bool, err error) {
...
 	c.mu.Unlock()
 
-	needReboot, err = u.Update(ctx)
-	c.recordCheckIn(reg, time.Now(), err)
-	return needReboot, err
+	// Build the request from the config channel's state, perform a single check-in, and dispatch the
+	// response to each handler. The update channel reports via its own follow-up check-ins in HandleUpdate.
+	req, err := u.CheckInRequest(ctx)
+	if err != nil {
+		return false, err
```
mattnathan, line 378: 🔴 **Store-read error leaves connectivity stale, disarming rollback.**

This early `return false, err` fires *before* the `defer c.recordCheckIn(...)` on L268 is registered. The previous `Update` always recorded every error (including store errors). Now a deployment-store read failure in `CheckInRequest` leaves `Connectivity` at its prior `Connected` value → `WaitConnected` returns immediately → `commitToSupervisor` commits the running version and cancels the auto-rollback for a node that can no longer check in.

Move the `defer` above this early return (or record explicitly on this path). Low probability (needs a local store failure) but a clean regression with a one-line fix.

internal/cloud/binary_updater.go
```diff
@@ -0,0 +1,201 @@
...
+	switch {
+	case err == nil:
+		// The Supervisor accepted the install and will restart BOS shortly. Nothing more to do here.
+		return nil
+	case status.Code(err) == codes.FailedPrecondition:
+		// The Supervisor is already working on an update. Benign: a later poll resolves the outcome.
+		u.logger.Info("supervisor reports an update already in progress",
+			zap.String("deploymentId", deploymentID))
+		return nil
+	default:
```
mattnathan, line 251: 🟠 **A bad artefact wedges the update channel permanently.**

This `default` branch reports `InstallingUpdate`-with-error (moving the deployment to `in_progress`) and returns, but never reports `FailedUpdate`. The cloud re-offers the active deployment every poll (~5 min default, `controller.go:759`), so it re-attempts and re-fails indefinitely — no terminal state, no backoff.

Two confirmed triggers: an artefact version that isn't a valid OCI tag (e.g. semver `1.2.3+build` — `createUpdateArtefact` only checks non-empty, no format validation), or an empty `sha256` (nullable, "empty until computed", offerable during the hashing window). Both make the supervisor reject with `InvalidArgument`, which isn't `FailedPrecondition`.

Suggest reporting `FailedUpdate` on non-retryable errors so a bad deployment reaches a terminal state.

internal/cloud/update_updater.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// The Supervisor correlates its version-keyed outcome with this deployment via the opaque id BOS
+	// passed on InstallUpdate. Only act on the Supervisor's terminal/in-flight state when it is about
+	// this same deployment; otherwise treat the deployment as new and install it.
+	if supStatus.GetDeploymentId() == deploymentID {
+		switch supStatus.GetState() {
+		case supervisorpb.UpdateStatus_COMPLETED:
+			// The Supervisor confirmed the update, but BOS is not (yet) running the target version. Report
+			// current to match the Supervisor's view.
+			if _, err := client.CheckIn(ctx, CheckInRequest{
```
mattnathan: 🟡 **COMPLETED branch reports only `CurrentUpdate`.**

The version-match fast path (L99-102) deliberately sends *both* `InstallingUpdate` and `CurrentUpdate` so a still-`pending` deployment completes in one check-in. This `COMPLETED` branch sends only `CurrentUpdate`, which the cloud handler no-ops on a `pending` deployment (`sim/check_in.go:224`, which only transitions from `in_progress`).

Reachable only if the intermediate `InstallingUpdate` (L174) was lost, so it's an edge — but the fix is trivially correct (mirror L99-102, send both), so worth just doing.

cmd/cloudsim/update_artefacts.gohtml
```diff
@@ -0,0 +1,63 @@
...
+<h2>Artefacts{{if .SiteID}} (site {{.SiteID}}){{end}}{{if .Platform}} (platform {{.Platform}}){{end}}</h2>
+<table border="1">
+<tr><th>ID</th><th>Site ID</th><th>Platform</th><th>Version</th><th>SHA-256</th><th>Size</th><th>Description</th><th>Created</th><th>Actions</th></tr>
+{{range .UpdateArtefacts}}
+<tr>
+  <td>{{.ID}}</td>
+  <td>{{if .SiteID.Valid}}{{.SiteID.Int64}}{{end}}</td>
+  <td>{{.Platform}}</td>
+  <td>{{.Version}}</td>
+  <td><small>{{.Sha256.String}}</small></td>
```
mattnathan: 🟡 **`.Sha256.String` on a `[]byte` column breaks the page.**

`queries.UpdateArtefact.Sha256` is `[]byte` (`models.go:71`); `{{.Sha256.String}}` makes `html/template` error mid-render as soon as any artefact exists, and `render()` has already written a 200, so the page truncates mid-row. Sim tooling, so low blast radius. Use `{{printf "%x" .Sha256}}` (the API path already hex-encodes correctly).

supervisor/internal/install/download.go
```diff
@@ -0,0 +1,84 @@
...
+//
+// The URL must use HTTPS, since it is a capability URL carrying its own authorisation, unless
+// allowInsecure is true.
+// If client is nil, http.DefaultClient is used.
+func DownloadAndVerify(ctx context.Context, client *http.Client, srcURL string, wantSHA256 []byte, destDir string, maxBytes int64, allowInsecure bool) (string, error) {
+	u, err := url.Parse(srcURL)
+	if err != nil {
+		return "", fmt.Errorf("parse url: %w", err)
+	}
+	if u.Scheme != "https" && !allowInsecure {
```
mattnathan, line 40: 🟡 **Hardening: HTTPS enforced only on the original URL.**

The scheme check is on the initial URL only, and production uses `http.DefaultClient` (`main.go:71`) which follows redirects with no scheme re-check. An `https` capability URL that 302s to `http://…` traverses cleartext. Integrity is safe (SHA-256 is verified), so this is confidentiality hardening for the capability URL — a `CheckRedirect` that rejects non-`https` hops (unless `allowInsecure`) closes it.

supervisor/internal/install/podman.go
```diff
@@ -0,0 +1,130 @@
...
+		if _, err := p.podman(ctx, "tag", p.tag(version), p.tag("current")); err != nil {
+			return fmt.Errorf("repoint current tag: %w", err)
+		}
+	}
+	return p.restart(ctx)
+}
+
+// compareTags reports whether tags a and b resolve to the same image, by comparing image IDs. A tag that
+// does not exist - e.g. a missing :current on the first install - never matches.
+func (p *PodmanInstaller) compareTags(ctx context.Context, a, b string) bool {
```
mattnathan: 🟡 **Hardening: `compareTags` treats any `inspect` error as "not equal".**

On the resume-after-tag-swap path a transient `podman image inspect` failure makes this return false, so the swap isn't skipped and `:previous` could be overwritten with the current (bad) image — the case the skip in `Apply` exists to prevent. Local-image inspect doesn't really fail transiently, so low risk, but distinguishing "tag absent" from "inspect failed" would be more robust. (Related: the first-install `tag current→previous` error on L58 is swallowed at `Debug`, conflating a genuine failure with the benign no-`:current` case.)

supervisor/internal/config/config.go
```diff
@@ -0,0 +1,70 @@
...
+	// StateDir holds durable state: the update record at <StateDir>/state.json and downloaded artefacts
+	// under <StateDir>/staging.
+	StateDir string `json:"stateDir,omitempty"`
+	// ImageRepo is the image repository whose :current/:previous tags the installer swaps.
+	ImageRepo string `json:"imageRepo,omitempty"`
+	// Unit is the systemd unit (and container name) for BOS that the Supervisor restarts.
+	Unit string `json:"unit,omitempty"`
+	// CommitDeadline bounds how long reconcile waits for BOS to Commit a new version before rolling back.
+	// It is a Go duration string, e.g. "2m" or "90s".
+	CommitDeadline jsontypes.Duration `json:"commitDeadline,omitempty"`
```
mattnathan, line 38: 🟡 **`CommitDeadline` isn't validated.**

An explicit `"0s"` or negative value is accepted; `time.NewTimer(0)` fires immediately so every update rolls back, and `main.go:109`'s `2*CommitDeadline` shutdown wait becomes zero. Needs deliberate misconfiguration (default is 2m), so P4 — but a lower-bound check in `Load` would fail loudly instead of silently rolling back every update.

mattnathan (commented):
Reviewed the update-supervisor design and implementation. The core is genuinely well-built — persisted goal, idempotent reconcile, crash recovery, and the commit-generation guard against confirming on a stale pre-rollback version are all correct and carefully documented. Findings are posted inline and ranked; the first (on `version.go`) is a functional blocker I'd want resolved before merge, the rest are ordered by severity.

**Design notes (non-blocking):**
- `commitDeadline` is a single "everything budget" (download + load + restart + boot + cloud-connect), re-armed on every reconcile/reboot — it conflates "still booting" with "unhealthy". Worth separating a boot/health budget from the download budget eventually.
- `version == OCI tag` has leaked into the wire contract (`tagPattern` at the `InstallUpdate` boundary, `server.go:139`), not just the podman installer. The `Installer` seam is otherwise clean, so FreeBSD is additive work — but this validation and the `version`-string identity will need a contract touch.
- The supervisor has no path to update **itself**, and its restart-and-confirm safety model relies on an external observer (BOS) that doesn't exist for the supervisor. Recommend stating the "supervisor is updated out-of-band" assumption explicitly in `state-model.md` so it's a decision, not an accident.
- No version/capability handshake between BOS and the supervisor, and `state.json` has no `schemaVersion`; `loadState` silently ignores unknown fields, so a supervisor downgrade would drop newer state. Both cheap to add while `v1` is young.
- `commitToSupervisor` gates on cloud connectivity only, sampled once, and skips the wait entirely for `Unconfigured` nodes. That's consistent with the documented "health = a configured cloud connection" model, so it's working as designed — but the rollback safety net is only as deep as that one signal, and a node registering just after the check commits without confirming a check-in. Flagging as a known limitation.

internal/cloud/update_updater.go
hexaglow: Looking at this again, I've decided that it doesn't make sense to report the same version as both 'installing' and 'current', that's contradictory. What should really happen is that once the version is installed, we report it as current, and the server should be able to recognise that as install+success.

pkg/app/version.go
hexaglow: The version is now injected at image build time in a similar way to the UI version string. The docker action has been altered to use add this version argument to the build.

hexaglow:
@mattnathan This PR will need some more changes to account for a somewhat different check-in endpoint design, but that needs to be done after the rebase to include the X.509 auth from #890. Please confirm you're happy with the resolutions to your review points.

mattnathan:
Happy with the resolutions — I went through all eight against `c82ba0e8f` and they hold up. The version stamping (`-ldflags -X buildVersion`), the redirect `CheckRedirect` (nice touch copying the client so `DefaultClient` isn't mutated), the `compareTags` exists-vs-error split, the terminal `FailedUpdate` on `InvalidArgument`, the `recordCheckIn` reorder, the deadline validation, and the hex render all look correct.

On the COMPLETED/current point — agreed, your call is better than my suggestion. Reporting the same version as both installing and current is contradictory; completing from `pending` on a `currentUpdate` server-side is the cleaner fix.

One thing to confirm rather than re-open: the version match still depends on the artefact-publish step and the image-build step deriving the version string from the **same** source, so `buildVersion` byte-matches the cloud's artefact version (`GIT_VERSION` defaults to `(unknown)` if unset). Assuming CI wires both from one source, we're good.

Understood on the check-in endpoint redesign landing after the #890 X.509 rebase — I'll take another pass once that's in.

hexaglow:
Rebased. Significant conflicts due to the parallel changes to the cloud client and cloudsim in #890, but no significant changes of approach or functionality were required.

Proceeding with changes to check-in API structure to reflect recent changes to Smart Core Connect.

hexaglow:
@mattnathan updated to reflect changes made to the schema on the SCC side. Some restructuring was required to account for the fact that SCC has a unified ID space for all deployments, whereas the client (and cloudsim) previously considered them to be different ID spaces. 

Commits marked `[WIP]` will be squashed into previous commits before merge.

mattnathan (commented):
Full re-review after the rebase + check-in-schema rework (unified prefixed deployment-id space, type-prefixed checksums, platform `{os,arch}`). My eight earlier findings are all still fixed, and the redesign is a real improvement in shape — `{os,arch}` is wired end-to-end, deployment ids are correctly opaque to BOS and the supervisor, and the responsibility split held.

This pass focused on **context**: changes that are fine in isolation but wrong against the wider system, and pre-existing code the rework has left out of place. Four systemic seams (tagged 🔴 S1–S4 inline) are where that bites; correctness findings are tagged 🟠 C1–C8 and design/consistency notes 🟡. None are hard blockers for a WIP branch, but **S1 and S2 are the two I'd want fixed before this drives real fleets.**

Summary of the systemic seams:
- **S1** — the `version` string is overloaded as SCC version + OCI image tag + the value BOS must `Commit`, and nothing validates the contract; a mismatch is a permanent, install-time-only failure.
- **S2** — the config channel wasn't brought to parity with the binary channel: fail-open on an empty checksum, and no terminal state on a checksum mismatch (re-downloads forever).
- **S3** — md5 is incoherent end-to-end: accepted on verify, emitted by nobody, and unstorable (the schema hardcodes a 32-byte sha256 column).
- **S4** — the binary-update lifecycle is write-only in the sim's own observability surface (read projection + UI drop it).

internal/cloud/update_updater.go
```diff
@@ -0,0 +1,223 @@
...
+		zap.String("version", version.Version),
+	)
+	installing := ProgressReport{DeploymentID: deploymentID, State: ProgressInstalling}
+	if _, err := client.CheckIn(ctx, CheckInRequest{Progress: []ProgressReport{installing}}); err != nil {
+		return fmt.Errorf("check-in to report installing update: %w", err)
+	}
+
+	installCtx, cancel := context.WithTimeout(ctx, supervisorCallTimeout)
+	defer cancel()
+	_, err = u.installer.InstallUpdate(installCtx, &supervisorpb.InstallUpdateRequest{
```
mattnathan: 🔴 **S1 — `version` is silently overloaded three ways, and nothing validates the contract.**

The cloud accepts any non-empty `version` (`sim/update_artefacts.go` does no format check), but the supervisor requires it to be a legal **OCI image tag** (`server.go:138` `tagPattern`) *and* BOS must later `Commit` a byte-identical `EffectiveVersion()` for the update to confirm (`server.go:247` vs `controller.go:864`). So a perfectly reasonable artefact version — `1.2.3+build`, a `v` prefix, a Go module pseudo-version, or a `BOS_VERSION_OVERRIDE` set inconsistently — makes the update either rejected as InvalidArgument (now terminal-failed) or applied-then-silently-rolled-back, and the constraint is invisible at the point of data entry: it only surfaces at install. The `{os,arch}` rework addressed a different axis and left this one. Worth validating the tag shape at artefact-create time and pinning a single source for the version BOS commits vs the version the cloud ships.

internal/cloud/config_updater.go
```diff
@@ -51,113 +50,179 @@ func NewDeploymentUpdater(store *DeploymentStore, client Client, opts ...Updater
...
 	}
-	defer body.Close()
+	// payload is the reader handed to the store; closing it closes the underlying body.
+	payload := body
+	defer func() { _ = payload.Close() }()
 
-	if err := u.store.WriteInstalling(ctx, latest.Deployment.ID, body); err != nil {
-		req.InstallingDeployment.Error = err.Error()
-		_, reportErr := u.client.CheckIn(ctx, req)
+	// Verify the download against the version's checksum when the server supplies one.
```
mattnathan: 🔴 **S2 — the config channel wasn't brought to parity with the binary channel.**

Two integrity gaps here that the binary path doesn't have:
1. Verification is gated on `latest.Version.Checksum != ""`, so a config version offered with **no checksum is downloaded and written with no integrity check at all** (fail-open). `InstallUpdate` rejects an empty checksum on the binary side (`server.go:133`); this should fail closed too.
2. On a checksum **mismatch**, the deployment is left `installing` with an error and returns — there's no terminal `failed` state, so the server re-offers it and BOS **re-downloads the same corrupt payload every poll forever** (and marks the connection Failed each time). The binary channel got the terminal-failure fix last round (`update_updater.go:203`); config needs the equivalent.

internal/util/checksum/checksum.go
```diff
@@ -0,0 +1,82 @@
...
+	"hash"
+	"strings"
+)
+
+// Algorithm is a supported hash function, the "<algo>" half of a type-prefixed checksum.
+type Algorithm string
+
+// Supported algorithms.
+const (
+	MD5    Algorithm = "md5"
```
mattnathan, line 21: 🔴 **S3 — md5 support is incoherent across the stack (half-in, half-out).**

md5 is accepted on the artefact **verify** path (here, and `download.go`/`server.go` apply no minimum-strength floor), on the one code path that ends in executing a downloaded payload. But **no producer ever emits md5** (every compute site uses sha256), and the store **can't even hold it** — migration 0005 declares `sha256 BLOB` with `CHECK(length = 32)` on both `config_versions` and `update_artefacts`, so a 16-byte md5 digest would be rejected on insert. Net: the wire claims algo-agnostic, the verifier accepts a collision-broken hash, and the DB is sha256-only. Either drop md5 entirely, or fully support it and rename the columns — but pin `>= sha256` at the verify boundary regardless of what the wire allows.

internal/cloud/sim/check_in.go
```diff
@@ -27,34 +33,278 @@ type CheckInAck struct {
...
+	}
+}
+
+func (a *checkInRecord) params(nodeID int64) queries.CreateNodeCheckInParams {
+	return queries.CreateNodeCheckInParams{
+		NodeID:                       nodeID,
+		CurrentDeploymentID:          nullInt64(a.currentConfig),
+		InstallingDeploymentID:       nullInt64(a.installingConfig),
+		InstallingDeploymentError:    nullString(a.configError),
+		InstallingDeploymentAttempts: nullInt64FromInt(a.configAttempts),
```
mattnathan, line 307: 🔴 **S4 — the binary-update lifecycle is write-only in the sim's observability surface.**

The check-in writes `CurrentUpdateDeploymentID` / `InstallingUpdateDeploymentID` / `InstallingUpdateError` into `node_check_ins` here, but the read projection `toNodeCheckIn` (`node_check_ins.go`, unchanged by this PR) exposes **only the config fields** — so `GET …/check-ins` and the cloudsim `check_ins.gohtml` page show nothing while a node is installing or failing a **binary** update. During exactly the rollout this PR adds, an operator watching the sim sees an empty update stream. The read/UI side was never extended to match this write. (Related: the move to `running.config` means steady-state check-ins no longer populate `current_deployment_id` either — `updater.go:83` — so even the config "current deployment" column now renders blank between installs.)

internal/cloud/binary_updater.go
```diff
@@ -0,0 +1,223 @@
...
+		u.logger.Warn("supervisor rejected the update as invalid, reporting failure to server",
+			zap.String("deploymentId", deploymentID),
+			zap.String("version", version.Version),
+			zap.String("reason", reason),
+		)
+		_, reportErr := client.CheckIn(ctx, CheckInRequest{
+			Progress: []ProgressReport{{DeploymentID: deploymentID, State: ProgressFailed, Reason: reason}},
+		})
+		return errors.Join(fmt.Errorf("install update via supervisor: %w", err), reportErr)
+	default:
```
mattnathan, line 251: 🟠 **C1 — an update-channel error fails the whole shared check-in, which gates commit-on-boot.**

`conn.Update` wraps `HandleUpdate`'s error and the deferred `recordCheckIn` sets `Connectivity=Failed` on any non-nil err. `commitToSupervisor` blocks in `WaitConnected` until Connected. So a persistent error on *any* binary deployment — including an unrelated second one, or the InvalidArgument branch that already reported the deployment failed — keeps the connection Failed, BOS never sends its `Commit`, and the supervisor rolls back the healthy running version after its deadline. Update-channel health shouldn't feed the connection's `Connected` signal that rollback depends on.

internal/cloud/update_updater.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	deploymentID := latest.Deployment.ID
+	if deploymentID == "" {
+		return nil
+	}
+	version := latest.Version
+
+	// BOS is running the target version, so the update succeeded. Report it current; the server completes
+	// the deployment whether it was still pending or already in_progress.
+	if u.version != "" && version.Version == u.version {
```
mattnathan: 🟠 **C2 — the version-match branch reports a deployment applied before the supervisor confirms it.**

This branch reports `applied` to the cloud and returns *before* ever calling `GetUpdateStatus`. If the supervisor then rolls back (Commit missed within its deadline — e.g. the socket flaps while the version is actually fine), BOS reverts, but the cloud has already marked the deployment completed and won't re-offer it, so it permanently believes the node runs the new version with nothing left to correct the drift. Because it returns early it can never observe the FAILED for this same id. Consider confirming against the supervisor before reporting applied, or leaving the deployment correctable.

supervisor/internal/server/server.go
```diff
@@ -0,0 +1,464 @@
...
+// artefact, apply it, then await BOS asserting the new version via Commit. On a missing commit within
+// the deadline it rolls back to the previous version. A fresh InstallUpdate and startup recovery both
+// run it.
+func (s *Service) reconcile(t target) {
+	defer s.wg.Done()
+	defer s.endReconcile()
+
+	ctx := context.Background()
+	s.mu.Lock()
+	sinceGen := s.commitGen // only a commit after this point confirms the new version
```
mattnathan: 🟠 **C3 — `sinceGen` is captured too early, so a pre-restart heartbeat can falsely confirm.**

`sinceGen` is read at reconcile *start*, before download + Apply. The comment says "only a commit after this point confirms", but "this point" precedes the restart. A routine `Commit(target)` from the still-running instance during the download window bumps `commitGen` past `sinceGen`, so `awaitCommit(t.Version, sinceGen)` is already satisfied when it runs — settling COMPLETED **without waiting for the freshly-restarted BOS to assert health**. Most reachable on the recovery path (re-driving an already-applied goal) or a reinstall of the running version. Capture `sinceGen` after `Apply`.

supervisor/internal/server/server.go
```diff
@@ -0,0 +1,464 @@
...
+
+// reconcile is the single idempotent operation that drives the host to the goal: download + verify the
+// artefact, apply it, then await BOS asserting the new version via Commit. On a missing commit within
+// the deadline it rolls back to the previous version. A fresh InstallUpdate and startup recovery both
+// run it.
+func (s *Service) reconcile(t target) {
+	defer s.wg.Done()
+	defer s.endReconcile()
+
+	ctx := context.Background()
```
mattnathan: 🟠 **C4 — `reconcile` awaits on `context.Background()`, so a graceful shutdown mid-commit-window forces a spurious rollback.**

When an install is parked in `awaitCommit` after Apply restarted BOS, a supervisor SIGTERM calls `GracefulStop`, which stops serving RPCs — so the restarting BOS can't land its `Commit`. The await isn't tied to any cancellable context, so it burns the full `commitDeadline`, times out, and **rolls back a healthy version**, and shutdown stalls ~2×deadline. The settled FAILED is persisted, so recovery won't re-drive it: the node is left downgraded purely because shutdown raced the commit.

supervisor/internal/server/server.go
```diff
@@ -0,0 +1,464 @@
...
+// running, good version - the rollback baseline for the next update. See commit-protocol.md.
+func (s *Service) Commit(_ context.Context, req *supervisorpb.CommitRequest) (*supervisorpb.CommitResponse, error) {
+	if req.GetVersion() == "" {
+		return nil, status.Error(codes.InvalidArgument, "version is required")
+	}
+
+	s.mu.Lock()
+	s.noteCommitLocked(req.GetVersion())
+	var snapshot *record
+	if !s.reconciling && s.rec.Committed != req.GetVersion() {
```
mattnathan: 🟠 **C5 — a stray heartbeat can flip an already-terminal status, and rollback trusts `Committed` over the tag.**

After an update settles COMPLETED (`Committed==Target.Version`), a later heartbeat `Commit(other)` with no reconcile in flight overwrites `rec.Committed` and persists it, so `GetUpdateStatus` now finds `Committed != Target` and reports the previously-terminal update as **FAILED** — a settled outcome silently changes. Relatedly, `rollback` (L231) awaits `rec.Committed` while the installer rolls back to the podman `:previous` tag; if those two diverge (e.g. via this heartbeat, or a prior partial op) it waits for the wrong version and times out despite a successful rollback. The string `Committed` and the `:previous` image are tracked independently with nothing cross-checking them.

internal/cloud/sim/config_deployments.go
```diff
@@ -240,10 +240,10 @@ func (s *Server) getDeployment(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var item queries.Deployment
+	var item queries.ConfigDeployment
 	err = s.store.Read(r.Context(), func(tx *store.Tx) error {
```
mattnathan, line 245: 🟠 **C6 — `default` status case is missing a `return`.**

The switch writes a 400 but falls through, so a POST with `{"status":"in_progress"}` still inserts the row (all five statuses pass the DB CHECK). That leaves the node with an `in_progress` config deployment out of nowhere → every later `createConfigDeployment` returns 409 and the config stream is **permanently wedged** — plus a corrupt response (400 header + error body, then a stray 201 body). Carried over from `main`'s `deployments.go`, but the rework's own sibling `createUpdateDeployment` (`update_deployments.go`) has the `return`, so the two now sit inconsistently side by side — worth fixing while it's being touched.

internal/cloud/sim/check_in.go
```diff
@@ -213,10 +427,16 @@ func (s *Server) checkIn(w http.ResponseWriter, r *http.Request) {
 		},
 	}
 	if hasDeploy {
-		resp.LatestConfig = &LatestConfig{
-			Deployment:    toDeployment(deployment),
-			ConfigVersion: toConfigVersion(r, configVersion),
```
mattnathan, line 218: 🟠 **C7 — a missing binary payload 500s the check-in and drops the config offer too.**

The active update artefact is fetched (which stats the payload file) *after* the write tx, and on error the handler `writeError(500)` and returns **before** `writeJSON`, discarding the `LatestConfig` already prepared for this poll. So if a binary payload file is absent (removed out-of-band, restore issue, orphan sweep), the node gets a 500 every check-in and can no longer receive even config updates — wedged, rather than degrading to "binary unavailable". (Same coupling in `store/update_artefacts.go:145`: `ListUpdateArtefacts` stats every payload and 500s the whole admin list if one blob is missing.)

internal/cloud/config_updater.go
```diff
@@ -51,113 +50,179 @@ func NewDeploymentUpdater(store *DeploymentStore, client Client, opts ...Updater
...
+	active, err := u.store.ActiveID()
+	if err != nil {
+		return false, fmt.Errorf("get active deployment id: %w", err)
+	}
 
 	latest := resp.LatestConfig
 	if latest == nil {
 		return false, nil
 	}
 	if latest.Deployment.ID == active {
```
mattnathan, line 117: 🟠 **C8 — bare→prefixed deployment-id transition isn't reconciled against persisted state, and the re-report omits `Running`.**

This compares the offered (now type-prefixed) id against the id persisted in the DeploymentStore. A node carried over from the pre-rework scheme (bare ids, per the "separate ID spaces" description) has its active mark stored as e.g. `"5"`, but the server now offers `"c-5"`, so `latest.Deployment.ID == active` is false → BOS treats a config it already runs as new: re-reports installing, re-downloads the identical payload, reboots once to "apply" it. Self-heals after one cycle, but it's a spurious reboot on the coordinated upgrade. Also this re-report check-in carries only `Progress` and no `Running` block, unlike the primary poll — if the server derives the running projection per check-in it can transiently blank its view.

internal/cloud/sim/check_in.go
```diff
@@ -27,34 +33,278 @@ type CheckInAck struct {
...
+	Running  RunningState     `json:"running"`
+	Progress []ProgressReport `json:"progress,omitempty"`
 }
 
-// CheckInInstallingDeployment references a deployment being installed, optionally with error and attempt info.
-type CheckInInstallingDeployment struct {
-	ID       int64  `json:"id,string"`
-	Error    string `json:"error,omitempty"`
-	Attempts int    `json:"attempts,omitempty"`
+// RunningState reports what the node is running and its platform. Only Platform is acted on here (it
```
mattnathan, line 130: 🟡 **Design / consistency (grouped):**

- **The sim isn't a faithful SCC stand-in.** It accepts `running.config`/`running.binary` versions "for wire compatibility" but acts only on Progress + Platform (here) — the running-version reconciliation real SCC presumably drives is unmodelled, so sim-backed tests can pass while production SCC diverges (the drift this alignment was meant to prevent). It also uses `c-`/`b-` id prefixes (L167) where SCC uses `config:`/`binary:` — three prefix conventions now coexist in one schema.
- **The "unified id space" is nominal.** The prefix is stripped on arrival (L182) and everything still branches config-vs-binary (two tables, two updaters, two response fields). It buys correct namespacing but no internal simplification.
- **Platform match is rigid** (`update_deployments.go:160`): exact `{os,arch}` equality, and a node's platform is empty until first check-in, so creating a deployment for a freshly-registered node fails; a node reporting a *changed* platform is still offered the old-arch artefact (`check_in.go:396`).
- **Config-reboot + binary-install race** (`conn.go:378`): both can be initiated in one poll, so a process-restart-for-config and the supervisor's binary swap/restart can interleave.

supervisor/docs/state-model.md
```diff
@@ -0,0 +1,119 @@
...
+| `INSTALLING`  | the host is being or has been changed; we don't know if the new version is healthy |
+| `COMPLETED`   | the target version is running and confirmed good                                   |
+| `FAILED`      | the target is not running the version we want                                      |
+
+Host state managed by the Supervisor
+------------------------------------
+
+On Rocky / Podman the host holds one image per applied version, plus two moving tags the Supervisor
+owns: `:current` (what the BOS unit boots) and `:previous` (the rollback pointer). The BOS service is
+recreated from `:current` on restart. FreeBSD has no container tags, but the shape is the same: an
```
mattnathan: 🟡 **Docs / naming drift the rework left behind (grouped):**

- This FreeBSD variant (L41-43) has no code — there's one `Installer` (Podman), and platform is now `{os,arch}`. A reader believes a FreeBSD path exists.
- `commit-protocol.md` is referenced from `server.go:266`, `main.go:5`, and `install/installer.go:4` but **doesn't exist** (folded into this doc). Three dead breadcrumbs.
- This doc and `server.go:42/150` still call the persisted verification value a "sha"/"hash"; it's now an algo-prefixed `checksum` (possibly md5) — inviting someone to reintroduce sha256-specific 32-byte handling.
- Migration columns are named `sha256` though they hold an algo-agnostic digest (see S3).
- `config.go:52` `Load` validates only `commitDeadline`; an explicitly-empty `imageRepo`/`unit`/`socket`/`stateDir` passes and yields malformed podman refs / bad paths at runtime.

internal/util/checksum/checksum.go
hexaglow: The client has to support MD5 because that's what SCC issues. We additionally support SHA256 because cloudsim was already using it. Cloudsim's database doesn't need to be able to hold MD5 hashes because the hashing is performed by cloudsim itself so it just chooses to use SHA-256. 

Is the objection that using cloudsim won't exercise the MD5 code paths in BOS?

internal/cloud/config_updater.go
hexaglow: 1. Checksum is treated as optional for config versions because we didn't originally have checksums, so existing cloudsim databases don't have them. Cloudsim databases are only for testing and development, so we can break compatibility if needed, but I'm not inclined to do that so I didn't force it. Can I take it that you want checksums to always be required?
2. There's two cases here - a) the artefact is corrupt or modified on the server; b) the artefact was corrupted during the download process. In case b, it makes perfect sense to retry. In case a, it doesn't. Unfortunately there's no good way to tell what situation you're in. Perhaps a limit on the number of times each deployment can be attempted (3?) before it's considered permanently failed would cover this (and probably other failure cases too). It seems reasonable you might want some manual intervention after 3 failures anyway. Thoughts?

internal/cloud/update_updater.go
hexaglow: The source of truth should be the artefact version injected at build time. Perhaps the Go module version fallback in `EffectiveVersion` is unhelpful and should be removed. 

Validation of versions in cloudsim using the same regex as the supervisor uses is a good idea, I'll implement that.

internal/cloud/binary_updater.go
hexaglow: This sounds like the intended behaviour to me. We don't commit until we know that BOS is in a healthy state. We don't know that BOS is in a healthy state until it's successfully performed a check-in. Can you be more specific? An example?

internal/cloud/sim/check_in.go
hexaglow: 1. Are you referring to detecting that the running version matches an in-progress/pending deployment, and if so, automatically marking that deployment as complete?
2. On prefixes - the client shouldn't care what the prefixes are at all. Cloudsim only uses prefixes because it does divide these internally, but that's an implementation detail. The client shouldn't be deconstructing IDs at all, instead they must be treated as opaque strings. So there's no particular motivation to have cloudsim match the IDs generated by SCC exactly. In fact, having different ID structures helps to confirm that the client indeed doesn't depend on ID structure.
3. I don't think this is really a problem. We can't deploy to a node until we know its platform, that wouldn't make sense. If we must be able to create a deployment before a node checks in, then we need to determine its platform at the time the node record is created. And we should never have a case where a node changes architecture - that should be considered a new node instead. It's probably worth adding a check that, if the os/arch is already known, any values supplied must match.
4. Interleaved installations - good point, we should force at most one installation in progress at once. I think this should be enforced on the client side.

internal/cloud/config_updater.go
hexaglow: No nodes are deployed in production with SCC cloud updates, so adding logic for this case adds complexity (and coupling to the server ID structure, which is supposed to be opaque) without actual benefit. Any existing development users will have to clear out their config stores. It's an acceptable limitation.

supervisor/docs/state-model.md
hexaglow: I've standardised on "checksum" except where the schema requires it to be hash.

`Load` now performs additional validation of config values.

The `sha256` column names are accurate - those only ever store a raw SHA256 hash.

supervisor/internal/server/server.go
hexaglow: I've fixed the first point - a commit will now only affect `rec.Committed` if there's an update in progress not yet completed. Commits outside of that won't affect it.

There is an unfortunate double state - it would be preferable to determine the version from the tags, but Podman tags point to an image ID (hash) rather than another tag, so we can't easily ask Podman "what version is `previous`".

internal/cloud/update_updater.go
mattnathan: Yes, please remove the `Main.Version` fallback. `buildVersion` (the ldflags-injected value) is the only source that can ever match the cloud's artefact version — `Main.Version` is `(devel)` for a container `go build`, and a pseudo-version otherwise, so it can never match. Keeping it means a build that forgot to stamp `buildVersion` silently commits a plausible-but-wrong version and every update rolls back with nothing pointing at the cause; dropping it makes an un-stamped build fail loudly (empty version) instead. The cloudsim-side tag-regex validation is a great addition too.

internal/cloud/config_updater.go
mattnathan: Agreed the retry belongs in the install step. Right now neither channel does it, and they fail at opposite extremes: the binary path downloads once and fails terminally (so a transient blip permanently fails a good update), while the config path never reaches a terminal state at all — on a checksum failure it reports installing-with-error and stays `in_progress`, so it's re-offered and re-downloaded every poll forever.

On your second question: a deployment that reaches terminal `failed` is **not** re-offered (both `GetActive*` queries filter to `pending`/`in_progress`) — but the config checksum failure never gets there, which is exactly why it loops. Your flow (DL → verify → fail → DL again → verify → same failure → terminal `failed`) is right, and I'd prefer it over a poll-level attempt cap: a single in-attempt retry-then-terminalise fixes both channels — it gives the binary path the transient retry it lacks and gives the config path the terminal state it lacks. So: yes, require checksums (fail closed on empty), and do the retry-then-terminalise in the install step.

internal/util/checksum/checksum.go
mattnathan: Agreed — md5 support is correct since SCC issues it; my "remove it" was wrong. Two things worth doing: (1) a test that drives an md5 artefact through BOS so that path is actually exercised, and (2) the more important one — cloudsim only ever emits sha256, so it isn't a faithful stand-in for the real API here. Either make the sim able to emit md5 too, or document clearly that it diverges, so nobody (myself included) reads cloudsim as the source of truth for the wire contract and gets md5 wrong.

internal/cloud/binary_updater.go
mattnathan: You're right — commit-only-when-healthy is intentional and correct. Withdrawing this one.

internal/cloud/sim/check_in.go
mattnathan: On point 1 — no, I'm not asking the sim to reconcile running versions or auto-complete deployments from them. Confirmed from the code that nothing BOS-side consumes anything the server derives from `running.*` (BOS only *produces* `running.config`/`running.binary` and reacts to `LatestConfig`/`LatestBinary` plus the supervisor status), so cloudsim ignoring those fields doesn't weaken any BOS test — it's fine as a fixture. The only residual is the same sim-fidelity point as the md5 thread (someone treating cloudsim as a faithful copy of the API), so treat this as folded into that rather than a separate concern. Agreed on 2 (opaque ids — divergent prefixes are a feature), 3 (require supplied os/arch to match when already known), and 4 (enforce a single install client-side).

supervisor/internal/server/server.go
mattnathan: Thanks — the heartbeat-flip fix looks right. Agreed the `Committed`-vs-`:previous` double-state is unavoidable given Podman tags resolve to image IDs rather than tag names, so there's no clean way to ask "what version is `:previous`". Fine to leave as-is; happy to close this.

internal/util/checksum/checksum.go
hexaglow: I'll add some documentation on ways that cloudsim diverges from our SCC implementation.

internal/cloud/config_updater.go
hexaglow: Ah, I'd already implemented it as a 3-attempt cap. I think that does the job.

internal/cloud/update_updater.go
hexaglow: I believe this has been addressed by other refactoring

mattnathan (approved):
Approving. Reviewed across three rounds; every point raised is now addressed or accepted as a deliberate tradeoff:

- **Version contract** pinned — `EffectiveVersion` resolves via the build-time-injected `buildVersion` only, and cloudsim validates the version as an image tag.
- **Update integrity** — config checksums now required (fail-closed), md5 supported per SCC and exercised by tests, and a 3-attempt cap terminalises a failing deployment on both channels (fixing both the config retry-loop and the binary no-retry).
- **Supervisor state machine** — commit-generation captured before Apply, commit wait cancelled on shutdown instead of rolling back, stray heartbeats can't flip a settled status, and the tag-lookup failures propagate.
- **Cloudsim** — binary fields surfaced in the check-in read projection, missing payloads degrade gracefully, invalid config deployments rejected cleanly, and deviations from SCC are documented.

Accepted tradeoffs: the `Committed`/`:previous` double-state (Podman tags resolve to image IDs), the opaque prefixed-id scheme, and cloudsim not being a byte-faithful SCC replica (now documented).

Assuming the `[WIP]` commits are squashed before merge per your note. Nice work.

hexaglow:
govulncheck failure unrelated to changes introduced in this PR. Merging.

PR #900, opened 2026-07-03
feat(api): add pagination to List RPCs that lacked it

deanredfern26:
[SCB-696](https://vanti.youtrack.cloud/issue/SCB-696)
Add page_size/page_token request fields and next_page_token/total_size response fields, plus server-side pagination, to the four List RPCs that previously returned unbounded lists:

- BookingApi.ListBookings
- HubApi.ListHubNodes
- TenantApi.ListTenants
- TenantApi.ListSecrets

Implementations follow the existing typespb.PageToken convention (default page size 50, max 1000), sorted by a stable key. A shared hubpage helper backs both the bolt and pgx hub stores. hub.NodeFetcher Poll and PullHubNodes page through / list all nodes so they still observe the full set.


mattnathan (approved):
Nice, mostly mechanical change. Good to be consistent

PR #901, opened 2026-07-03
feat(stores): support read, write, and admin pgx connection pools

deanredfern26:
[SCB-1038](https://vanti.youtrack.cloud/issue/SCB-1038) Support r, w, admin connection pools for pgx

## Summary

Splits each postgres store's single `*pgxpool.Pool` into three role-scoped pools — **read**, **write**, and **admin** — so a deployment can connect each privilege level as a distinct, least-privilege database user.

- New `pgxutil.RoleConfig` (embeds `ConnectConfig`, plus optional `read`/`write`/`admin` sub-blocks) and `pgxutil.Pools` / `ConnectRoles` helpers in `internal/util/pgxutil/roles.go`.
- `stores.PostgresPoolsFor(ctx, rc)` returns the shared node pools when `rc` is zero, or opens a dedicated set (auto-closed on `ctx` done) otherwise.
- Each postgres query is routed to the minimum-privilege pool it needs:
  - **read** — `SELECT` only
  - **write** — `INSERT`/`UPDATE`/`DELETE` and any transaction that both reads and writes
  - **admin** — schema DDL (`CREATE`/`ALTER`) and maintenance (`VACUUM`)
- Systems updated: `alerts`, `hub`, `tenants`, `publications`, `history`, plus the `history` automation.

Fully backward compatible: `RoleConfig` embeds `ConnectConfig` inline, so an existing `{uri, passwordFile}` config parses unchanged and opens a single shared pool (no behaviour change). Roles with identical configs share one underlying pool.

See the new `pkg/app/stores/README.md` for configuration and suggested grants.

## Notes
- `dataretention` `Compact` now runs `VACUUM` on the admin pool (it requires table ownership / the `MAINTAIN` privilege).
- Dedicated pools are closed automatically when their context is cancelled; the `tenants` factory was switched to `service.MonoApply` so this cleanup fires on config reload like the other factories.

## Test plan
- [ ] `go test ./internal/util/pgxutil/... ./pkg/app/stores/... ./pkg/system/... ./pkg/history/...`
- [ ] Existing single-`uri` deployments open one pool and behave as before.
- [ ] With `read`/`write`/`admin` sub-blocks + least-privilege grants, reads/writes/DDL each succeed on their scoped user.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/app/stores/README.md
+++ b/pkg/app/stores/README.md
@@ -1,0 +1,78 @@
+# Stores
+
+The `stores` package provides shared storage connections (databases) for the
+systems and automations running on a controller. It is configured via the
+top-level `stores` block in the controller config:
+
+```json5
+{
+  // other config
+  "stores": {
+    "postgres": {
+      "uri": "postgres://username@localhost:5432/smart_core",
+      "passwordFile": "/secrets/postgres-password"
+    }
+  }
+}
+```
+
+Systems that use postgres (`alerts`, `hub`, `tenants`, `publications`,
+`history`) and the `history` automation borrow this shared pool by default. A
+system can instead open its own dedicated connection by setting a `storage`
+block with its own connection config; see that system's README.
+
+## Read / write / admin roles (least privilege)
+
+Each postgres query runs over a connection scoped to the minimum privilege it
+needs:
+
+- **read** — `SELECT` only
+- **write** — `INSERT` / `UPDATE` / `DELETE` (and any transaction that both reads
+  and writes)
+- **admin** — schema setup (`CREATE` / `ALTER`) and maintenance (`VACUUM`)
+
+By default all three roles use the single `uri`/`passwordFile` above, so a
+single connection pool is opened (no behaviour change). To harden a deployment,
+connect each role as a distinct database user by adding `read`, `write`, and/or
+`admin` sub-blocks. Any role you omit falls back to the top-level config.
+
+```json5
+{
+  "stores": {
+    "postgres": {
+      // admin defaults to the top-level config below unless overridden
+      "uri": "postgres://sc_admin@localhost:5432/smart_core",
+      "passwordFile": "/secrets/sc_admin-password",
+      "read": {
+        "uri": "postgres://sc_read@localhost:5432/smart_core",
+        "passwordFile": "/secrets/sc_read-password"
+      },
+      "write": {
+        "uri": "postgres://sc_write@localhost:5432/smart_core",
+        "passwordFile": "/secrets/sc_write-password"
+      }
+    }
+  }
+}
+```
+
+The three roles are expected to point at the **same database** and differ only
+in the database user (and therefore its privileges).
+
+### Suggested grants
+
+The `admin` role owns the schema (it runs `CREATE TABLE`/`ALTER` on startup and
+`VACUUM`). Once the tables exist, grant the read/write roles the minimum they
+need. For example, per table used by the enabled systems:
+
+```sql
+-- read role: SELECT only
+GRANT SELECT ON <table> TO sc_read;
+
+-- write role: DML (writers often need to read within a transaction too)
+GRANT SELECT, INSERT, UPDATE, DELETE ON <table> TO sc_write;
+
+-- admin role: owns the table (DDL) and can run VACUUM (needs ownership or the
+-- MAINTAIN privilege on PostgreSQL 16+)
+ALTER TABLE <table> OWNER TO sc_admin;
+```
--- a/pkg/app/stores/README.md
+++ b/pkg/app/stores/README.md
@@ -76,3 +76,27 @@ GRANT SELECT, INSERT, UPDATE, DELETE ON <table> TO sc_write;
 -- MAINTAIN privilege on PostgreSQL 16+)
 ALTER TABLE <table> OWNER TO sc_admin;
 ```
+
+Tables with a serial/identity column (currently `history`, whose `id` is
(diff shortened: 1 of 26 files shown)


pkg/app/stores/README.md
```diff
@@ -0,0 +1,78 @@
...
+The `admin` role owns the schema (it runs `CREATE TABLE`/`ALTER` on startup and
+`VACUUM`). Once the tables exist, grant the read/write roles the minimum they
+need. For example, per table used by the enabled systems:
+
+```sql
+-- read role: SELECT only
+GRANT SELECT ON <table> TO sc_read;
+
+-- write role: DML (writers often need to read within a transaction too)
+GRANT SELECT, INSERT, UPDATE, DELETE ON <table> TO sc_write;
```
mattnathan, line 73: The documented writer grants are insufficient for any table with a serial/identity column — currently the `history` table (`id BIGSERIAL PRIMARY KEY`). `INSERT ... RETURNING id` calls `nextval()` on the implicit sequence, which needs `USAGE` on that sequence in addition to table DML.

Verified against a live postgres 14 with a role granted exactly these lines:

```
INSERT INTO history (...) RETURNING id  →  ERROR: permission denied for sequence history_id_seq
```

Adding the sequence grant fixes it:

```sql
GRANT USAGE, SELECT ON ALL SEQUENCES IN SCHEMA public TO sc_write;
-- cover sequences created later, too:
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT USAGE, SELECT ON SEQUENCES TO sc_write;
```

(The UUID-keyed tables need nothing extra — `uuid_generate_v4()` is `PUBLIC`-executable — so this can be scoped as "for tables with serial/identity columns".)

mattnathan (changes requested):
Solid feature — clean abstraction, backward compatible, pool de-dup and lifecycle handled well (nice catch switching `tenants` to `MonoApply` so dedicated pools close on reload like the others). Query routing looks correct on a read-through, including the subtle ones (`VerifySecret` is read-only, read-then-write txns go to `write`, `Compact`→`admin`). Build and the `pgxutil`/`stores`/`system`/`history` test suites are green.

Requesting changes on one blocking item; the rest are non-blocking.

### Blocking

**Writer grants in the stores README are wrong for the `history` table.** See the inline comment — an operator following the documented grants verbatim gets `permission denied for sequence history_id_seq` on the first history write, because `id BIGSERIAL` needs `USAGE` on its sequence. Confirmed empirically against a live DB. The whole value of the feature is operators getting these grants right, so the docs need the sequence grant (scoped to serial/identity tables).

### Non-blocking

- **Partial role override with no base URI.** If only `read` (say) is set with no top-level `uri` and no `write`/`admin`, those roles resolve to a zero `ConnectConfig` and `Connect("")` falls through to libpq env defaults (`PGHOST`, …) rather than failing clearly — `RoleConfig.IsZero()` won't catch it since `Read != nil`. Is a single-role override intended to be valid? If not, worth a guard; if so, a doc note.
- **Connection budget.** With three distinct role configs there are now three independent pools, each with pgxpool's default max-conns — so total connections to the same DB can ~2–3× vs. before. A sentence in the README about sizing `pool_max_conns` would help operators.
- **Nit:** `postgresStore` now carries both `pools` and the `r, w, admin` fields; the latter are redundant now that `pools` exists. Not worth churn unless you're in there anyway.

pkg/app/stores/README.md
deanredfern26: Good catch — confirmed. Added the sequence grant to the README, scoped to serial/identity tables (`history` currently), with both the one-off `GRANT USAGE, SELECT ON ALL SEQUENCES` and the `ALTER DEFAULT PRIVILEGES` for later migrations, and a note that UUID-keyed tables need nothing extra. Fixed in 00ac18932.

deanredfern26:
Thanks for the non-blocking notes — all three addressed in 00ac18932:

- **Partial role override with no base URI** — went with a guard. `pgxutil.ConnectRoles` now rejects any role whose *effective* config is empty (e.g. only `read` set with no base `uri`), returning an error naming the unconfigured role instead of silently falling through to libpq env defaults. `RoleConfig.IsZero()` still handles the fully-empty case for the shared-pool path. Added a unit test.
- **Connection budget** — added a "Connection sizing" section to the README: each distinct role config opens its own pool at `max(4, numCPU)` default max-conns, so splitting into read/write/admin multiplies peak connections up to 3×; size each pool with the `pool_max_conns` query param so the total stays within the DB's `max_connections`.
- **Redundant fields nit** — dropped `postgresStore`'s `r, w, admin` fields; the store now routes through the `pools` value (`pools.Read` is the connected sentinel).

mattnathan (approved):
All four items addressed in 00ac18932 — verified.

- **Sequence grants (blocking)** — README now documents `GRANT USAGE, SELECT ON ALL SEQUENCES` + `ALTER DEFAULT PRIVILEGES`, scoped to serial/identity tables, with the UUID note. This is exactly the grant I confirmed unblocks `INSERT ... RETURNING id` on the `history` table against a live DB. ✅
- **Partial role override** — nice, you went with a guard rather than just docs: `ConnectRoles` now rejects an empty effective role config before connecting, naming the role, and there's a test for it. ✅
- **Connection sizing** — clear new README section on the up-to-3× pool multiplication and `pool_max_conns`. ✅
- **Redundant fields** — `r/w/admin` dropped, routed through `pools`; sentinel switched to `pools.Read`. ✅

`go build ./...` clean; `pgxutil` (incl. the new guard test) and `stores` suites green. LGTM.

PR #902, opened 2026-07-08
feat(sccexporter): UDMI telemetry + discovery to Connect over MQTT v5 (+ DBO building-config)

deanredfern26:
[SCB-1371](https://vanti.youtrack.cloud/issue/SCB-1371)

## Summary

Implements the Connect telemetry plugin: the `sccexporter` automation now discovers
meter devices, polls the typed Meter trait, and publishes per-device **UDMI pointset
telemetry** plus periodic **device-metadata (discovery)** to the Connect telemetry
(Event Grid MQTT) broker over **MQTT v5** with **mTLS** using the node's Connect leaf
credential. Also adds:

- **`pkg/dbo`** — maps Smart Core meter data onto Google Digital Buildings Ontology
  standard field names/units/entity types (electricity + water), with a conformance
  test guarding field names against the DBO subfield registry.
- **`cmd/tools/gen-dbo`** — generates a DBO building-configuration from a node's meters
  (identity translation over the DBO field names the exporter publishes).

Related: SCC-592 (Connect ingest / consumer), SCC-611 (Event Grid MQTT front-door),
docs/bos-auth.md (auth contract).

## What's included

- MQTT v5 publisher (autopaho) with mutual-TLS; cloud-credential mode (node leaf cert,
  renewal-aware) or file-path certs for dev/test.
- UDMI pointset telemetry + discovery on distinct topics; read-only (Meter trait only).
- Configurable payload naming: `pointNaming: "dbo"` (default, canonical field names on
  the wire) or `"raw"` (raw `usage`/`produced` names, native units, mapping left to Connect).
- vanti-ugs example wired with the automation; `gen-dbo` output validates cleanly against
  Google's `instance_validator`.

## Verification

- `go build ./...` + unit tests green **on Linux/CI** (see caveat below).
- Mutual-TLS MQTT v5 broker smoke test (`mqtt_test.go`).
- Ran end-to-end against the vanti-ugs node + a local broker — telemetry + discovery
  observed for every meter.
- `gen-dbo` output verified with Google's DBO `instance_validator` ("All entities
  validated SUCCESSFULLY" bar the upstream `.units` regex issue below).

## Outstanding items (mostly need Connect sign-off — the ticket's open question)

- [ ] **Topic grammar + telemetry/discovery match rule** to agree with Connect (SCC-592).
      This branch emits `tlm/devices/<deviceRef>/events/{pointset,discovery}`; the ticket's
      illustrative form was `<prefix>/<deviceRef>/pointset/{events,points}`.
- [ ] **deviceRef** is currently the raw Smart Core device name (contains `/`); needs the
      `aka[bos].ref` convention before it is topic-segmentation-safe.
- [ ] **Discovery payload shape** is UDMI device-*metadata* on the `events/discovery` topic —
      confirm that matches what Connect's classifier expects.
- [ ] **Default point naming** is `dbo` (canonical names on the wire); the ticket specifies
      raw. Made configurable here — confirm the default with Connect.
- [ ] **`.units` (plural)** is the sole remaining `instance_validator` error — an upstream
      `google/digitalbuildings` `_UDMI_UNIT_FIELD_REGEX` bug (expects `.unit`); to report upstream.
- [ ] **`gen-dbo --point-naming`** flag not yet added, so a `raw`-mode plugin's building-config
      would need non-identity translations (follow-up).
- [ ] **Scope:** Meter trait only for now.

## Notes for reviewers

- **Does not build on Windows** due to a pre-existing defect in main's `internal/cloud`
  (`registration_store.go` calls root `renameio.WriteFile`, which is `//go:build !windows`;
  should be `renameio/maybe.WriteFile`) — unrelated to this branch. CI/Linux is fine.
- Rebased onto current `main`; the cloud X.509 mTLS work (already merged) supplies the leaf credential.


pkg/auto/connecttelemetry/auto.go
```diff
@@ -89,74 +87,96 @@ func (a *AutoImpl) applyConfig(ctx context.Context, cfg config.Root) error {
...
 			case <-time.After(time.Until(next)):
-				t = time.Now()
+				t = a.now()
 			}
 
-			// send the metadata on first run and then every now and again.
-			publishMetadata := (iterationCount % *cfg.Mqtt.MetadataInterval) == 0
+			// Publish discovery on the first run and then every N cycles; refresh
+			// the device list on the same cadence.
+			publishDiscovery := (iterationCount % *cfg.Mqtt.MetadataInterval) == 0
```
mattnathan, line 95: **`metadataInterval: 0` panics the automation.** An explicit `0` in config unmarshals to a non-nil `*int`, so the `== nil` default in `config/root.go` never fires and this becomes `iterationCount % 0` — an unrecovered divide-by-zero in the background goroutine, which takes the controller down. Config validation rejects an out-of-range QoS but not a non-positive interval. (Confirmed with a quick test.)

pkg/auto/sccexporter/auto.go
```diff
@@ -89,74 +87,96 @@ func (a *AutoImpl) applyConfig(ctx context.Context, cfg config.Root) error {
...
-			// send the metadata on first run and then every now and again.
-			publishMetadata := (iterationCount % *cfg.Mqtt.MetadataInterval) == 0
+			// Publish discovery on the first run and then every N cycles; refresh
+			// the device list on the same cadence.
+			publishDiscovery := (iterationCount % *cfg.Mqtt.MetadataInterval) == 0
 			iterationCount++
 
-			if publishMetadata {
+			if publishDiscovery {
 				clear(allDevices)
```
mattnathan: **A transient `ListDevices` failure blanks the export set until the next discovery cycle.** `allDevices` is cleared *before* the fallible `refreshDevices`, and on error the loop `continue`s with the map already empty. Since the refresh only runs on discovery cycles, at the default `metadataInterval: 100` × `sendInterval: 15m` that's a long silent gap from one blip. The example config's `metadataInterval: 1` masks it.

docs/connect-telemetry-ingest.md
```diff
@@ -0,0 +1,203 @@
...
+- `tlm/` is the fixed intent prefix — the publish-authz scope. The MQTT topic arrives
+  downstream as the CloudEvent `subject`.
+- **No nodeId in the topic.** Publisher identity rides the mTLS credential and is stamped via
+  routing enrichments, not the topic path.
+- `<deviceRef>` is the device (producer) segment. Connect is intended to resolve it as
+  `identity=(system=bos, ref)` against `aka[bos].ref` — see the deviceRef open issue below.
+
+### How kinds are distinguished
+
+Connect has **no `classifyKind` function or fixed subfolder rule**. `ingest.md` says a
```
mattnathan: This contract description has aged out. SCC-592 (merged 2026-07-06, two days before this PR opened) shipped a `classifyKind` plus a per-kind layout/match mechanism, so "Connect has **no `classifyKind` function or fixed subfolder rule**" is no longer accurate — as is the "PR #890 not landed" note at line 195 now the credential work has merged. The design still copes; flagging the doc because it's the artifact people will trust. (Leaving the topic-order line to the SCC side.)

example/config/vanti-ugs/app.conf.json
```diff
@@ -4898,6 +4898,22 @@
           }
         }
       }
+    },
+    {
+      "name": "van/uk/brum/ugs/automation/scc-telemetry",
+      "type": "sccexporter",
```
mattnathan: An observation on the name rather than the choice: `sccexporter` describes a narrower role than the plugin is growing into (return/control, cloud-driven config), and unlike the internals the name is baked into this `type` config contract and the `allautos` factory key — so it gets more expensive to revisit the longer it's deployed. Flagging the timing.

pkg/auto/connecttelemetry/udmiexport.go
```diff
@@ -0,0 +1,101 @@
...
+func meterTelemetry(r *meterpb.MeterReading, support *meterpb.MeterReadingSupport, now time.Time, naming dbo.Naming) udmi.PointsetEvent {
+	ts := now
+	if r.GetEndTime() != nil {
+		ts = r.GetEndTime().AsTime()
+	}
+	points := udmi.PointsEvent{}
+	for _, f := range dbo.MeterFields(support.GetUsageUnit(), support.GetProducedUnit()) {
+		value := r.GetUsage()
+		if f.Exported {
+			value = r.GetProduced()
```
mattnathan, line 59: I noticed meter values are forwarded as `float32` (the trait's type — upstream, not this PR). For a cumulative accumulator that's resolution loss once the running total gets large, and sooner for meters reporting in `Wh`. Raising it because fidelity is the whole point of the pipeline; whether it's in scope here is your call.

pkg/auto/connecttelemetry/mqtt.go
```diff
@@ -4,180 +4,161 @@ import (
...
-	return s
-}
 
-// reconnect attempts to reconnect to the MQTT broker with exponential backoff.
-// Returns true if reconnection succeeded, false otherwise.
-// do not call this more than once per sccConnector instance
-func (s *sccConnector) reconnect(ctx context.Context) bool {
-	if s.reconnecting {
-		return false
+	u, err := url.Parse(cfg.Host)
```
mattnathan, line 44: Minor: a `mqtt://` host silently discards the TLS config — autopaho keys transport off the URL scheme, so the certs / `useCloudCredential` are ignored and telemetry goes out in the clear with no guard. Production fails closed at Event Grid, but a permissive local broker wouldn't; the shipped example (`mqtt://localhost:1883` with cert paths set) is exactly this shape.

mattnathan (approved):
Approving. All six review threads addressed in the follow-up commits (`4ae1082c0`, `0ad159a90`, `cdcaeb12e`, `db10284fc`, `646c2e2b6`, `e283b15a9`): the `metadataInterval: 0` crash and the refresh-failure device-blanking are fixed, the non-TLS scheme now fails closed, the contract doc and float32 precision are documented, and the automation is renamed to `connecttelemetry`. Build and tests green at `e283b15a9`. The rename/test observations don't apply since the old names and bugs never reached main.

PR #903, opened 2026-07-08
cloud: fix registration store write on Windows

deanredfern26:
[SCB-1372](https://vanti.youtrack.cloud/issue/SCB-1372) fix renameio.WriteFile not working on windows

The top-level renameio package is unsupported on Windows and returns an error, so saving the cloud registration failed on Windows hosts. Switch to renameio/v2/maybe, which performs the atomic rename where supported and falls back to a best-effort direct write on Windows.

### DIFF
--- a/internal/cloud/registration_store.go
+++ b/internal/cloud/registration_store.go
@@ -10,7 +10,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/google/renameio/v2"
+	"github.com/google/renameio/v2/maybe"
 
 	"github.com/smart-core-os/sc-bos/internal/util/pki"
 )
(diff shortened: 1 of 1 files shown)


mattnathan (approved):
Approve. Correct, minimal fix and the diagnosis is accurate.

Verified locally:
- The top-level `renameio/v2` `WriteFile` does error on Windows, so the fix is on the mark; `maybe.WriteFile` calls straight through to `renameio.WriteFile` on non-Windows, so no behaviour change on Linux/macOS.
- `renameio/v2 v2.0.0` in go.mod ships the `maybe` subpackage; `go build ./internal/cloud/` passes.
- Brings this call site in line with the rest of the codebase — `internal/confmerge/confmerge.go` and `pkg/app/appconf/store.go` already use `renameio/v2/maybe`.
- Existing `TestFileRegistrationStore_RoundTrip` still exercises `Save`.

One optional thing to consider (not blocking): the `NewFileRegistrationStore` godoc still says "written atomically, 0600", which no longer strictly holds on Windows. Could soften to "written atomically where the OS supports it" so it agrees with the new `Save` comment — but happy to leave as-is if you'd rather.

PR #905, opened 2026-07-09
ops: add aggregated chart-data CSV export to PeopleCountHistoryCard

deanredfern26:
The existing export downloaded only raw device readings, not the aggregated max-per-bucket series shown on the graph. Add a second "Export chart data (CSV)..." action that serialises exactly what the chart displays.

PeopleCountHistoryChart exposes buildExportRows() so the card can reuse the already-computed series (no extra fetch, CSV matches the graph), including the prior-period baseline column when the overlay is shown. Bucket granularity follows the chart tickUnit automatically. Reuses the existing downloadCSVRows helper; datePeriodString is now exported for consistent filename dating.

I am just adding this only to the PeopleCountHistoryCard for now as it can from a client request to be able to do this.

### DIFF
--- a/ui/ops/src/dynamic/widgets/occupancy/PeopleCountHistoryCard.vue
+++ b/ui/ops/src/dynamic/widgets/occupancy/PeopleCountHistoryCard.vue
@@ -27,6 +27,7 @@
               <v-list-subheader title="Data"/>
               <period-chooser-rows v-model:start="_start" v-model:end="_end" v-model:offset="_offset"/>
               <v-list-item title="Export chart data (CSV)..."
+                           :disabled="!canExport"
                            @click="onDownloadAggregatedClick"
                            v-tooltip:bottom="'Download the aggregated data shown on the chart'"/>
               <v-list-item title="Export raw data (CSV)..."
@@ -132,6 +133,12 @@ const {summaryPct} = useOccupancyNormalized(
 
 const chartRef = ref(null);
 
+// Slug used for download filenames; collapse whitespace runs to single dashes.
+const titleSlug = computed(() => (props.title?.trim() || 'people-count').toLowerCase().replace(/\s+/g, '-'));
+
+// Whether the chart currently has any data worth exporting.
+const canExport = computed(() => Boolean(chartRef.value?.hasData));
+
 const currentColor = vColors.blue.base;
 
 const priorPeriodTooltip = computed(() => {
@@ -147,16 +154,16 @@ const onDownloadAggregatedClick = () => {
   if (!chart) return;
 
   const {rows, tickUnit} = chart.buildExportRows();
-  const slug = props.title?.toLowerCase()?.replace(' ', '-') ?? 'people-count';
+  if (rows.length <= 1) return; // header only, nothing to export
   const dateString = datePeriodString({startTime: startDate.value, endTime: endDate.value});
-  const filename = `${slug}-chart-${tickUnit}-${dateString}.csv`;
+  const filename = `${titleSlug.value}-chart-${tickUnit}-${dateString}.csv`;
   downloadCSVRows(filename, rows);
 };
 
 const onDownloadClick = async () => {
   if (!props.downloadEnterLeave) {
     await triggerDownload(
-        props.title?.toLowerCase()?.replace(' ', '-') ?? 'people-count',
+        titleSlug.value,
         {conditionsList: [{field: 'name', stringEqual: props.totalOccupancyName}]},
         {startTime: startDate.value, endTime: endDate.value},
         {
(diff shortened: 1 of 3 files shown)


ui/ops/src/dynamic/widgets/occupancy/PeopleCountHistoryChart.vue
```diff
@@ -232,6 +232,36 @@ watch(chartData, (data) => {
...
+  const header = ['Time', 'People Count'];
+  if (includeBaseline) header.push('Prior Period People Count');
+
+  const fmt = (v) => (v === null || v === undefined) ? '' : String(v);
+
+  const rows = [header];
+  for (let i = 0; i < totals.length; i++) {
+    const t = totals[i];
+    const time = t.x instanceof Date ? t.x : new Date(t.x);
+    const row = [time.toISOString(), fmt(t.y)];
```
mattnathan, line 257: Timestamps here are serialised as UTC (`toISOString()`), whereas the chart's x-axis and the bucket edges from `useDateScale` (`startOfDay` etc.) are local time. So a bucket the graph labels e.g. "00:00" can appear in the CSV as `…T23:00:00.000Z` for a viewer east of UTC.

We're deliberately keeping UTC for now to stay consistent with the full raw-data download, so no change needed — just flagging it as a known trade-off against the "CSV matches the graph exactly" intent, in case we want to revisit and emit local time later.

ui/ops/src/dynamic/widgets/occupancy/PeopleCountHistoryCard.vue
```diff
@@ -134,6 +140,19 @@ const priorPeriodTooltip = computed(() => {
...
 });
 
+// Downloads the aggregated series currently shown on the chart (max people count
+// per auto-sized bucket) as a CSV, rather than the raw device readings.
+const onDownloadAggregatedClick = () => {
+  const chart = chartRef.value;
+  if (!chart) return;
+
+  const {rows, tickUnit} = chart.buildExportRows();
+  const slug = props.title?.toLowerCase()?.replace(' ', '-') ?? 'people-count';
```
mattnathan: `String.replace(' ', '-')` only swaps the first space, so a multi-word title like "Ground Floor Lobby" produces `ground-floor lobby-chart-…csv` with a surviving space in the filename. Worth using `replaceAll(' ', '-')` (or `/\s+/g`) here.

(Also note the `?? 'people-count'` fallback is effectively dead — `title` has a default value so it's never nullish.)

This is copied from the existing `onDownloadClick`, so the raw-export path has the same quirk; might be worth fixing both while we're here.

ui/ops/src/dynamic/widgets/occupancy/PeopleCountHistoryCard.vue
```diff
@@ -134,6 +140,19 @@ const priorPeriodTooltip = computed(() => {
...
+  const dateString = datePeriodString({startTime: startDate.value, endTime: endDate.value});
+  const filename = `${slug}-chart-${tickUnit}-${dateString}.csv`;
+  downloadCSVRows(filename, rows);
```
mattnathan, line 160: `downloadCSVRows` doesn't escape `,`/`"` (it has a `// todo` for this and just `join(',')`s). For this call site the data is ISO timestamps + integers with a fixed header, so there's no injection/comma risk today and this is fine as-is.

Flagging as existing behaviour only — probably worth a follow-up ticket to harden the helper before any column ever carries free text (e.g. a device name).

ui/ops/src/dynamic/widgets/occupancy/PeopleCountHistoryCard.vue
```diff
@@ -26,16 +26,19 @@
             <v-list density="compact">
               <v-list-subheader title="Data"/>
               <period-chooser-rows v-model:start="_start" v-model:end="_end" v-model:offset="_offset"/>
-              <v-list-item title="Export CSV..."
+              <v-list-item title="Export chart data (CSV)..."
```
mattnathan, line 29: When the chart has no data, `buildExportRows()` still returns just the header row and the download fires with a header-only CSV. Could we disable this list item when there's no data to export (e.g. gate on the chart's `hasData`/loaded state, exposed alongside `buildExportRows`)? That'd match the raw path better, which no-ops when there are no records.

deanredfern26:
issues addressed and tickets made for the follow on points raised:

[SCB-1375](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1375) Harden `downloadCSVRows` to escape CSV special characters
[SCB-1376](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1376) People Count chart CSV exports timestamps in UTC while the chart displays local time

mattnathan (approved):
All four review points addressed in 34572b7e3:

- Filename slug fix extracted into a shared `titleSlug` computed (collapses all whitespace) and applied to both the aggregated and raw export paths — fixes the pre-existing quirk everywhere.
- Empty-export case now handled by disabling the item via the newly-exposed `hasData`, with a header-only guard as backup.
- UTC timestamps kept (consistent with the raw download) and CSV escaping left to the helper, both captured as follow-ups in SCB-1376 and SCB-1375.

Reactivity on `canExport` is sound. LGTM 👍

PR #906, opened 2026-07-10
ops: config-driven external links in the Operations sidebar

deanredfern26:
## Summary

Adds an optional, config-driven way to surface **external links** in the Operations sidebar — links to other applications served alongside the Ops UI (e.g. standalone SPAs on the same origin). This is a generic, reusable capability; no site-specific values live in the code.

Motivation: a deployment (40 Leadenhall) needs to link operators from the Ops UI to two standalone SPAs (out-of-hours AC request/approve). Today every nav item is an internal router link (`:to`); there was no mechanism for an external URL.

## Config shape

Deployments opt in via `config.ops.externalLinks` in `ui-config.json`:

```json
"ops": {
  "externalLinks": [
    { "title": "OOH AC Requests",  "icon": "mdi-air-conditioner",         "href": "/-/ooh-ac-request/", "target": "_blank" },
    { "title": "OOH AC Approvals", "icon": "mdi-clipboard-check-outline", "href": "/-/ooh-ac-approve/", "target": "_blank" }
  ]
}
```

`target` is optional (defaults to `_blank`). Items are gated purely by presence in config — not `pathEnabled()` — since external URLs are not router paths.

## Changes

- `src/routes/ops/nav.js` — `ExternalNavItem` typedef + `useExternalNavItems()` reading `ops.externalLinks`.
- `src/routes/ops/OpsNav.vue` — renders the external items with `:href`/`:target` after the existing nav list.

## Security hardening

- **href allowlist**: only `http:`/`https:` and root-relative (`/…`) hrefs are rendered; `javascript:`/`data:` and other schemes are dropped (defence-in-depth against a bad config, since `:href` bindings aren't sanitised by Vue).
- **`rel="noopener noreferrer"`** on the external links to prevent reverse tabnabbing and referrer leakage on `target="_blank"`.

## Backwards compatibility

Fully additive — deployments without `ops.externalLinks` are unchanged.

## Testing

- `eslint` clean on both files (UI CI is lint-only; the ops package has no unit-test suite).
- ⚠️ Not yet visually verified in a browser — I couldn't run a local `vite build`/dev server in my environment (missing `@rolldown` native binding). CI builds on Linux; a quick visual check of the sidebar rendering before merge would be worthwhile.

### DIFF
--- a/ui/ops/src/routes/ops/nav.js
+++ b/ui/ops/src/routes/ops/nav.js
@@ -12,14 +12,6 @@ import {computed} from 'vue';
  */
 
 /**
- * @typedef {Object} ExternalNavItem
- * @property {string} title - The title of the menu item
- * @property {string} icon - The icon to display for the menu item
- * @property {string} href - The external URL the item links to
- * @property {'_self'|'_blank'} [target] - Where to open the link, defaults to '_blank'
- */
-
-/**
  * Menu Items
  * This is the main list of items
  *
@@ -109,44 +101,4 @@ export function isRouteEnabled(item) {
  */
 export function useEnabledNavItems() {
   return computed(() => navItems.filter(item => isRouteEnabled(item)));
-}
-
-/**
- * Checks whether an href is safe to render as an external link.
- *
- * Only http(s) URLs and root-relative paths are permitted. This rejects dangerous schemes
- * such as javascript: and data:, which would otherwise execute in the Ops UI origin when
- * clicked (the href comes from config, so this is defence-in-depth against a bad config).
- *
- * @param {string} href
- * @return {boolean}
- */
-function isSafeHref(href) {
-  if (typeof href !== 'string' || !href) return false;
-  // Root-relative paths (e.g. /-/ooh-ac-request/) are same-origin and safe.
-  if (href.startsWith('/')) return true;
-  try {
-    const url = new URL(href, window.location.origin);
-    return url.protocol === 'http:' || url.protocol === 'https:';
-  } catch {
-    return false;
-  }
-}
-
-/**
- * Get a computed reference to the external navigation links defined in the UI config.
- *
- * These are deployment-specific links to other applications served alongside the Ops UI
- * (e.g. standalone SPAs on the same origin). They are not router paths, so they are gated
- * purely by their presence in config rather than by pathEnabled().
- *
- * @return {ComputedRef<ExternalNavItem[]>}
- */
-export function useExternalNavItems() {
-  const uiConfig = useUiConfigStore();
-  return computed(() => {
-    const items = uiConfig.getOrDefault('ops.externalLinks', []);
-    if (!Array.isArray(items)) return [];
-    return items.filter(item => item && isSafeHref(item.href));
-  });
 }
--- a/ui/ops/src/routes/ops/nav.js
+++ b/ui/ops/src/routes/ops/nav.js
@@ -112,6 +112,28 @@ export function useEnabledNavItems() {
 }
 
 /**
+ * Checks whether an href is safe to render as an external link.
+ *
+ * Only http(s) URLs and root-relative paths are permitted. This rejects dangerous schemes
+ * such as javascript: and data:, which would otherwise execute in the Ops UI origin when
+ * clicked (the href comes from config, so this is defence-in-depth against a bad config).
+ *
+ * @param {string} href
+ * @return {boolean}
+ */
+function isSafeHref(href) {
+  if (typeof href !== 'string' || !href) return false;
+  // Root-relative paths (e.g. /-/ooh-ac-request/) are same-origin and safe.
+  if (href.startsWith('/')) return true;
+  try {
+    const url = new URL(href, window.location.origin);
+    return url.protocol === 'http:' || url.protocol === 'https:';
+  } catch {
+    return false;
+  }
+}
+
+/**
  * Get a computed reference to the external navigation links defined in the UI config.
  *
  * These are deployment-specific links to other applications served alongside the Ops UI
@@ -125,6 +147,6 @@ export function useExternalNavItems() {
   return computed(() => {
     const items = uiConfig.getOrDefault('ops.externalLinks', []);
     if (!Array.isArray(items)) return [];
-    return items.filter(item => item && typeof item.href === 'string' && item.href);
+    return items.filter(item => item && isSafeHref(item.href));
   });
 }
--- a/ui/ops/src/routes/ops/nav.js
+++ b/ui/ops/src/routes/ops/nav.js
@@ -12,6 +12,14 @@ import {computed} from 'vue';
  */
 
 /**
+ * @typedef {Object} ExternalNavItem
+ * @property {string} title - The title of the menu item
+ * @property {string} icon - The icon to display for the menu item
+ * @property {string} href - The external URL the item links to
+ * @property {'_self'|'_blank'} [target] - Where to open the link, defaults to '_blank'
+ */
+
+/**
  * Menu Items
  * This is the main list of items
  *
@@ -101,4 +109,22 @@ export function isRouteEnabled(item) {
  */
 export function useEnabledNavItems() {
   return computed(() => navItems.filter(item => isRouteEnabled(item)));
+}
+
+/**
+ * Get a computed reference to the external navigation links defined in the UI config.
+ *
+ * These are deployment-specific links to other applications served alongside the Ops UI
+ * (e.g. standalone SPAs on the same origin). They are not router paths, so they are gated
+ * purely by their presence in config rather than by pathEnabled().
+ *
+ * @return {ComputedRef<ExternalNavItem[]>}
+ */
+export function useExternalNavItems() {
+  const uiConfig = useUiConfigStore();
+  return computed(() => {
+    const items = uiConfig.getOrDefault('ops.externalLinks', []);
+    if (!Array.isArray(items)) return [];
+    return items.filter(item => item && typeof item.href === 'string' && item.href);
+  });
 }
--- a/ui/ops/src/routes/ops/overview/OpsNavListItem.vue
+++ b/ui/ops/src/routes/ops/overview/OpsNavListItem.vue
@@ -21,6 +21,14 @@
           <v-list-item-title>{{ props.item.title }}</v-list-item-title>
           <template #append>
             <v-btn
+                v-if="showExternalButton"
+                v-bind="externalAttrs"
+                @click.stop
+                variant="text"
+                size="x-small"
+                style="font-size: 120%"
+                icon="mdi-open-in-new"/>
+            <v-btn
                 @click.prevent.stop="_props.onClick"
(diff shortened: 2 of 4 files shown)


ui/ops/src/routes/ops/nav.js
```diff
@@ -102,3 +110,43 @@ export function isRouteEnabled(item) {
...
+ * These are deployment-specific links to other applications served alongside the Ops UI
+ * (e.g. standalone SPAs on the same origin). They are not router paths, so they are gated
+ * purely by their presence in config rather than by pathEnabled().
+ *
+ * @return {ComputedRef<ExternalNavItem[]>}
+ */
+export function useExternalNavItems() {
+  const uiConfig = useUiConfigStore();
+  return computed(() => {
+    const items = uiConfig.getOrDefault('ops.externalLinks', []);
```
mattnathan: A dedicated `ops.externalLinks` section duplicates what `ops.pages` already is. A page item and an external-nav item carry near-identical shape — `title`/`icon` plus a destination — and the existing nav pipeline (`OpsNavListItem`) already runs on a rule this generalises cleanly into: `layout` owns the row's primary click, and other affordances demote to appended buttons when the row is taken. That's exactly how `children` renders today (with `layout`: row navs + chevron button; without: the row itself expands). An `href` slots in as a third citizen of that same rule rather than needing anything bespoke.

Folding `href` into `pages` would collapse a whole parallel concept — this `externalLinks` config key, `useExternalNavItems`/`isSafeHref` here, and the separate render block in `OpsNav.vue` — into the one pipeline. It's also the only shape that lets a link sit *between* dashboard pages rather than pinned below the fixed menu items, which is where these deployment links most naturally belong.

One case that would need a deliberate call rather than falling out for free: a leaf page without `layout` is a live `:to` link today (`OpsNavListItem.vue:59`), so a "no layout ⇒ no in-app nav" rule would change that behaviour — worth checking no current config leans on a layout-less leaf being navigable.

deanredfern26: Agreed — reworked to fold `href` into the `pages` pipeline as you described.

- `href` is now a third destination on a `dashboardPage`: `OpsNavListItem` renders an external anchor when the item has a (safe) `href`, an in-app `:to` when it has a `layout`, and expands when it only has `children`. The primary destination owns the row click and the chevron demotes to an appended button — same `hasOwnPage` rule for both the layout and href cases.
- Removed the parallel apparatus: `ops.externalLinks`, `useExternalNavItems`, `isSafeHref`, and the extra render block in `OpsNav.vue` are all gone. `nav.js` and `OpsNav.vue` are back to their `main` state; the whole change now lives in `OpsNavListItem.vue` plus a schema addition.
- Re your caveat: layout-less leaves still render as `:to` links (unchanged) — `hasOwnPage` only *adds* href/layout as reasons to nav; it doesn't gate the leaf branch, so no "no layout ⇒ dead row" behaviour change. I verified the 40LH config has no layout-less leaves anyway, and the overview `beforeEnter` still only auto-lands on layout-bearing pages.
- Kept the href scheme allowlist (http/https/root-relative) and `rel="noopener noreferrer"`, now colocated in `OpsNavListItem`.
- Added `href`/`target` to the `dashboardPage` def in `schema/ui-config.schema.json`.

Deployment config (40LH) now places the two links as `href` children under an "Out of Hours AC" group in `ops.pages`. Still not visually smoke-tested in a browser locally (missing `@rolldown` binding here) — CI builds on Linux.

ui/ops/src/routes/ops/overview/OpsNavListItem.vue
```diff
@@ -127,6 +128,61 @@ const active = computed(() => route.path === toAreaLink.value || route.path.star
...
+ */
+const hasOwnPage = computed(() => isExternal.value || Boolean(props.item.layout));
+
+/**
+ * The link attributes to bind to the row's list item: an external anchor when the item has a
+ * safe href, otherwise an in-app router link to the area.
+ *
+ * @type {import('vue').ComputedRef<Object>}
+ */
+const linkAttrs = computed(() => {
```
mattnathan, line 207: Have you considered rendering a `layout`+`href` row as *both* — the in-app route on the row plus an appended `[ext]` link button — rather than letting `href` win outright?

Right now `linkAttrs` treats the two as mutually exclusive: if a safe `href` is present the row becomes an external anchor and the `layout` is silently unreachable. But `$defs/dashboardPage` in the schema permits `layout` and `href` together with no constraint, so a `{layout, href}` page is valid config the UI can't fully render — the routed page just disappears with no warning at load, validation, or render. (Confirmed in a browser against vanti-ugs: such a page renders external-only.)

Surfacing the external link as an appended icon button next to the chevron would reuse the exact "row owns the primary action, appended button owns the secondary" pattern the chevron already uses — and would close the gap between what the config can express and what the nav can show.

ui/ops/src/routes/ops/overview/OpsNavListItem.vue
```diff
@@ -127,6 +128,61 @@ const active = computed(() => route.path === toAreaLink.value || route.path.star
  * @type {import('vue').ComputedRef<string>}
  */
 const toAreaLink = computed(() => `/ops/overview/${currentPath.value}`);
+
+/**
+ * Whether the item is an external link, i.e. it has a safe href.
+ *
+ * @type {import('vue').ComputedRef<boolean>}
+ */
+const isExternal = computed(() => isSafeHref(props.item.href));
```
mattnathan, line 156: When an `href` fails `isSafeHref` — a `javascript:`/`data:` scheme, or anything the URL parser rejects — `isExternal` is false and the row falls through to an in-app `:to` link at `/ops/overview/{path}`, a route that (for a config author who meant an external link) almost certainly 404s.

So a bad href doesn't get dropped or flagged; it silently degrades to a dead in-app link. It's safe from an XSS angle — the unsafe URL never reaches the DOM (confirmed in a browser: a `javascript:` href rendered as `<a href="/ops/overview/...">`) — but a config typo produces a broken nav entry with no signal at load, validation, or render. A `console.warn` when an item carries an href that `isSafeHref` rejects would at least make the misconfiguration visible.

ui/ops/src/routes/ops/overview/OpsNavListItem.vue
deanredfern26: Good catch — done, rendering both. A `layout` now owns the row (in-app `:to`) and the `href` is surfaced as an appended `mdi-open-in-new` button, using the same row-owns-primary / appended-secondary pattern as the chevron (`@click.stop` so the button's external nav doesn't trigger the row's route). An href-only row is still the external anchor itself, and a `{layout, href, children}` row shows both the ext button and the chevron.

Precedence is layout-owns-the-row (per your original "layout owns the primary click"), so the in-app page — the richer destination — is never the one that's lost. `showExternalButton = isExternal && hasLayout` drives the appended button; the leaf only provides the `#append` slot when it's needed, so ordinary rows don't gain padding. Config that sets both is now fully renderable, so I left the schema permissive rather than adding a mutual-exclusion constraint.

ui/ops/src/routes/ops/overview/OpsNavListItem.vue
deanredfern26: Agreed — added a `console.warn` in `OpsNavListItem` setup when an item carries an `href` that `isSafeHref` rejects, naming the item and the offending href, so a `javascript:`/`data:`/malformed URL is visible at load rather than silently degrading to a dead `/ops/overview/...` link. XSS safety is unchanged (the unsafe URL still never reaches the DOM).

ui/ops/src/routes/ops/overview/OpsNavListItem.vue
```diff
@@ -19,6 +20,14 @@
           </template>
           <v-list-item-title>{{ props.item.title }}</v-list-item-title>
           <template #append>
+            <v-btn
```
mattnathan, line 23: Three things worth tightening on this appended external control (both here and the leaf `#append` ~L81):

**1. The external control is an `<a>` nested inside the row's `<a>`** (DOM-confirmed) — invalid nesting, since an anchor can't contain another anchor. It works in the SPA today (no runtime warning), so flagging as a possible issue rather than something with an obvious fix. The tension: ideally both the in-app route *and* the external link behave as real links — middle-click / cmd-click to open in a new tab, and the status-bar URL preview on hover — which wants two genuine anchors; but pulling the external link out as a sibling isn't a clean way there either, since the row would no longer highlight on hover over the external icon. No solution in hand — just surfacing it. (The L+C row is the placement reference, but its secondary control is a `<button>` — the chevron — so nesting is fine there; a link doesn't have that luxury.)

**2. Give the link an accessible name.** It's icon-only (`mdi-open-in-new`) with `aria-label: null` and no text, so a screen reader announces an unnamed link. An `aria-label` like `Open <title> (external)` would name it.

**3. Icon size/colour looks off.** It renders noticeably bigger and full-white next to the nav's own icons — `size="x-small"` + `font-size:120%` on a text-variant button doesn't seem to pick up the nav-icon theme size/colour tokens the prepend `<v-icon>`s use. Worth matching the nav icon treatment so it sits with the rest.

deanredfern26: All three addressed (both sites — group activator and leaf `#append`):

**1. Nested anchor** — agreed two true anchors isn't attainable, and a sibling loses the row hover, so I took the shape you pointed at with the chevron: the external control is now a **button** that opens the href programmatically (`window.open(href, '_blank', 'noopener,noreferrer')`; `location.assign` for `_self`). That removes the `<a>`-in-`<a>` entirely and matches the established secondary-control precedent in this row. The deliberate trade-off: no middle-click / status-bar preview on the *secondary* control (the row keeps full anchor semantics) — partially recovered with a `title` attribute so hover still reveals the destination URL. Only applies to the layout+href combo; an href-only row is still a single real anchor.

**2. Accessible name** — added `aria-label="Open <title> (external)"`.

**3. Icon treatment** — dropped the `font-size:120%` hack and added `text-medium-emphasis`, so the icon uses the same on-surface/opacity tokens as the prepend nav icons rather than the v-btn's full-white default. Caveat: I still can't render this locally (broken `@rolldown` native binding on my machine), so the token match is by-inspection — since you've got vanti-ugs up in a browser, a quick visual confirm of the sizing would be appreciated.

mattnathan: Confirmed in a browser against vanti-ugs — sizing and colour look good, and the button approach is fine by me. The nested anchor is gone, the control is named, and the icon now reads as a muted secondary control. Approving.

mattnathan (approved):
Approving. Verified end-to-end in a browser against vanti-ugs: external links fold cleanly into the `ops.pages` pipeline (no parallel config), links sit between dashboard pages, `layout`+`href` renders both destinations, unsafe hrefs warn instead of degrading silently, and the appended external control is valid HTML, named, and matches the nav icon tokens. Nice work iterating through it.

PR #909, opened 2026-07-14
audit: exclude noisy RPCs from the audit log (SCB-1380)

deanredfern26:
## Summary

`ServerReflectionInfo`, `TestHubNode` and `CreateHistoryRecord` are high-frequency, low-value calls that drown out meaningful mutating operations in the audit log ([SCB-1380](https://vanti.youtrack.cloud/issue/SCB-1380)).

- Added `isAuditExcluded(service, method)` in `pkg/auth/policy/interceptor.go` — excludes `grpc.reflection.` services (covers both `v1` and `v1alpha` `ServerReflectionInfo`) plus `TestHubNode` and `CreateHistoryRecord` by exact method name.
- Gated the audit write with `!isAuditExcluded(...)`. Policy evaluation is untouched — only the audit-log side effect is suppressed.
- Added `TestIsAuditExcluded` covering the excluded RPCs plus negative cases (`TestEnrollment`, `StartFunctionTest`, `EnrollHubNode`, `CreateAccessGrant`) to guard against over-matching.

## Test plan

- `go test ./pkg/auth/policy/...` passes
- `go build ./...` succeeds

### DIFF
--- a/pkg/auth/policy/interceptor.go
+++ b/pkg/auth/policy/interceptor.go
@@ -225,7 +225,7 @@ func (i *Interceptor) checkPolicyGrpc(ctx context.Context, creds *verifiedCreds,
 		)
 	}
 	// Only audit once per RPC, not for every message on an open client/bidirectional stream.
-	if isWriteMethod(method) && !stream.Open {
+	if isWriteMethod(method) && !isAuditExcluded(service, method) && !stream.Open {
 		outcome := "allowed"
 		if err != nil {
 			outcome = "denied"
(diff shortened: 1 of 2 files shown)


hexaglow (changes requested):
The exclusion list should match on fully-qualified name, including service name, instead of method name alone, to prevent spurious matches if more services are added.

While I don't believe these method names exist in any other gRPC services, I don't want to place an expectation that gRPC service authors need to consider this kind of name matching when writing their services.

deanredfern26:
Good call — done in 33e4001. `auditExcludedMethods` is now keyed on the fully-qualified `service/method` name (`smartcore.bos.hub.v1.HubApi/TestHubNode`, `smartcore.bos.history.v1.HistoryAdminApi/CreateHistoryRecord`) rather than the bare method name, so an unrelated service reusing one of those method names won't be silently excluded. Added negative test cases covering exactly that scenario.

hexaglow (approved):

PR #911, opened 2026-07-17
ui/ops: escape CSV special characters in downloads (SCB-1375)

deanredfern26:
## Summary

Hardens the CSV download helpers in `ui/ops/src/util/downloadCSV.js` to escape special characters and neutralise spreadsheet formula injection, resolving [SCB-1375](https://vanti.youtrack.cloud/issue/SCB-1375).

`downloadCSVRows` and `csvDownload`'s `processRecordsForCSV` built CSV output by naively `join(',')`-ing each field (with a standing `// todo: escape ',' and '"' in data`). Any cell containing a comma, double-quote or newline corrupted the output — fields split across columns, broken rows, or CSV injection when opened in a spreadsheet.

This was already a latent bug, not purely theoretical: the helper has **five callers**, and the Notifications export (`routes/ops/notifications/export.js`) already emits free-text fields (`Source`, `Floor`, `Zone`, `Description`) that can contain these characters.

## Changes

- Add an `escapeCSVField` helper implementing RFC 4180 quoting: wrap a field in double quotes when it contains `,`, `"`, CR or LF, doubling any embedded `"`. Nullish values become an empty field (matching the previous `join` behaviour).
- Guard against spreadsheet formula injection: a non-numeric value beginning with `=`, `+`, `-`, `@`, tab or CR is prefixed with a single quote, so `=cmd|…`/`=HYPERLINK(…)` are neutralised in Excel/Sheets. The guard is scoped to non-numeric values (`isNaN(Number(str))`), so genuine numeric fields such as `-5` and ISO-8601 timestamps are left intact — no regression for numeric exports.
- Apply escaping in both `downloadCSVRows` and `processRecordsForCSV` (header + data rows).
- Remove the `// todo` comment.

## Testing

- eslint clean on the changed file.
- Round-trip smoke test against the **real** `downloadCSVRows` with realistic alert rows (comma in `AHU-1, North`, quotes in `Zone "A"`, embedded comma+newline in a description): the emitted CSV re-parses byte-for-byte back to the original rows via a strict RFC 4180 parser. Timestamp/integer fields stay unquoted (no regression).
- Formula-injection guard verified case-by-case: `=HYPERLINK(…)`/`=cmd|…`/`@handle`/tab-prefixed → prefixed with `'`; `-5`, `-5.5`, ISO-8601 timestamps and plain text → unchanged.

> **Note:** `ui/ops` has no unit-test harness today (no vitest/jest, no `test` script, CI only lints/builds). A dedicated unit test for `escapeCSVField` is tracked in [SCB-1400](https://vanti.youtrack.cloud/issue/SCB-1400) (stand up vitest + CI wiring), kept out of this security fix to avoid bootstrapping test infra on a patch.

## Acceptance criteria

- [x] Fields containing commas, quotes, and newlines round-trip correctly on re-import.
- [x] Existing People Count exports unchanged (no regression for timestamp/integer data).
- [x] Spreadsheet formula injection via free-text fields neutralised (optional criterion in the ticket).

### DIFF
--- a/ui/ops/src/util/downloadCSV.js
+++ b/ui/ops/src/util/downloadCSV.js
@@ -6,14 +6,20 @@ import {toValue} from 'vue';
  * quotes when it contains a comma, double-quote, CR or LF, doubling any
  * embedded double-quotes. Nullish values become an empty field.
  *
- * Note: no spreadsheet formula-injection guard (prefixing =/+/-/@) is applied,
- * as that would silently alter legitimate negative numeric values in exports.
+ * Non-numeric values beginning with =, +, -, @, tab or CR are prefixed with a
+ * single quote to neutralise spreadsheet formula injection (e.g. `=HYPERLINK(…)`,
+ * `=cmd|…`) when the export is opened in Excel/Sheets. Numeric values such as
+ * `-5` are recognised and left intact, so numeric/timestamp exports are
+ * unaffected.
  *
  * @param {*} field
  * @return {string}
  */
 function escapeCSVField(field) {
-  const str = field === null || field === undefined ? '' : String(field);
+  let str = field === null || field === undefined ? '' : String(field);
+  if (/^[=+\-@\t\r]/.test(str) && isNaN(Number(str))) {
+    str = `'${str}`;
+  }
   if (/[",\r\n]/.test(str)) {
     return `"${str.replace(/"/g, '""')}"`;
   }
--- a/ui/ops/src/util/downloadCSV.js
+++ b/ui/ops/src/util/downloadCSV.js
@@ -2,6 +2,25 @@ import {camelToSentence} from '@/util/string';
 import {toValue} from 'vue';
 
 /**
+ * Escapes a single field for RFC 4180 CSV output. Wraps the value in double
+ * quotes when it contains a comma, double-quote, CR or LF, doubling any
+ * embedded double-quotes. Nullish values become an empty field.
+ *
+ * Note: no spreadsheet formula-injection guard (prefixing =/+/-/@) is applied,
+ * as that would silently alter legitimate negative numeric values in exports.
+ *
+ * @param {*} field
+ * @return {string}
+ */
+function escapeCSVField(field) {
+  const str = field === null || field === undefined ? '' : String(field);
+  if (/[",\r\n]/.test(str)) {
+    return `"${str.replace(/"/g, '""')}"`;
+  }
+  return str;
+}
+
+/**
  * Generates a function to download records as a CSV file based on specified properties.
  *
  * @template T
@@ -70,11 +89,11 @@ export const csvDownload = async (params) => {
 
     // Generate the data rows
     const rows = flattenedRecords.map(record =>
-      Object.keys(headerMap).map(key => record[key]).join(',')
+      Object.keys(headerMap).map(key => escapeCSVField(record[key])).join(',')
     );
 
     // Return the CSV string
-    return [header.join(',')].concat(rows).join('\n');
+    return [header.map(escapeCSVField).join(',')].concat(rows).join('\n');
   };
 
 
@@ -137,8 +156,7 @@ export const csvDownload = async (params) => {
  * @param {string[][]} data
  */
 export function downloadCSVRows(filename, data) {
-  // todo: escape ',' and '"' in data
-  const csvContent = data.map(e => e.join(',')).join('\n') + '\n';
+  const csvContent = data.map(e => e.map(escapeCSVField).join(',')).join('\n') + '\n';
   download(filename, csvContent, 'text/csv;charset=utf-8;');
 }
 


ui/ops/src/util/downloadCSV.js
```diff
@@ -1,6 +1,25 @@
 import {camelToSentence} from '@/util/string';
 import {toValue} from 'vue';
 
+/**
+ * Escapes a single field for RFC 4180 CSV output. Wraps the value in double
+ * quotes when it contains a comma, double-quote, CR or LF, doubling any
+ * embedded double-quotes. Nullish values become an empty field.
+ *
+ * Note: no spreadsheet formula-injection guard (prefixing =/+/-/@) is applied,
+ * as that would silently alter legitimate negative numeric values in exports.
```
mattnathan: The docstring documents skipping the `=/+/-/@` guard because it would alter negative numbers — but that justification only holds for *numeric* fields. The fields actually exposed here are `Source`, `Zone`, `Description` (free text, device/user-influenced), which open directly in Excel/Sheets — the classic formula-injection vector (`=HYPERLINK(...)`, `=cmd|...`). A common compromise guards only non-numeric (or only quoted) fields, so `-5` stays `-5` but `=cmd` is neutralised. Given SCB-1375 reads as a hardening ticket, is formula injection explicitly in or out of scope here? Fine either way — just want it to be a decision rather than a side-effect of the negative-number argument.

ui/ops/src/util/downloadCSV.js
```diff
@@ -1,6 +1,25 @@
...
+ *
+ * @param {*} field
+ * @return {string}
+ */
+function escapeCSVField(field) {
```
mattnathan, line 18: `escapeCSVField` is pure with obvious cases (bare value, comma, embedded quote, CRLF, null) and this is a correctness/security fix — cheap and high-value to lock down with a unit test. There's no `downloadCSV` test file currently. Note it's not exported, so testing needs either exporting it or asserting through `downloadCSVRows`.

mattnathan (commented):
CSV escaping is implemented correctly and wired into both export paths. `escapeCSVField`'s trigger set `/[",\r\n]/` is exactly the RFC 4180 set, embedded quotes are doubled, and quoting is applied only when needed. Nullish→`''` matches the prior implicit `Array.join` behaviour, so numeric/timestamp fields are unaffected. Lint passes clean on the file. A few observations below — none blocking.

ui/ops/src/util/downloadCSV.js
```diff
@@ -70,11 +89,11 @@ export const csvDownload = async (params) => {
 
     // Generate the data rows
     const rows = flattenedRecords.map(record =>
-      Object.keys(headerMap).map(key => record[key]).join(',')
+      Object.keys(headerMap).map(key => escapeCSVField(record[key])).join(',')
     );
 
     // Return the CSV string
-    return [header.join(',')].concat(rows).join('\n');
+    return [header.map(escapeCSVField).join(',')].concat(rows).join('\n');
```
mattnathan, line 102: Minor / non-blocking: RFC 4180 specifies `\r\n` between records; both paths (here and `downloadCSVRows` at line 159) still join with `\n`. Pre-existing, not introduced here, and Excel/Sheets tolerate `\n` fine — just flagging that "RFC 4180" in the docstring refers to field escaping, not record separators.

mattnathan (changes requested):
Requesting changes on the formula-injection point (inline thread on line 10). SCB-1375 is a CSV-hardening ticket, but the primary injection vector — the free-text `Source`/`Zone`/`Description` fields opening in Excel/Sheets — is left unguarded. The negative-number rationale only justifies skipping the guard on numeric fields, so please either add a guard scoped to non-numeric/quoted fields, or explicitly confirm formula injection is out of scope for this ticket and record that decision.

The missing unit test (line 15) would also be good to address before merge given this is a security fix. The CRLF note (line 96) is minor and non-blocking.

The escaping implementation itself is correct and lints clean — this is about completeness of the hardening, not the code that is here.

ui/ops/src/util/downloadCSV.js
deanredfern26: Good call — brought it into scope. `escapeCSVField` now guards formula injection, but only on non-numeric values (0ae9f2122): a leading `=`, `+`, `-`, `@`, tab or CR is prefixed with a single quote, so `=cmd|…`/`=HYPERLINK(…)` are neutralised while genuine numbers like `-5` (`isNaN(Number(str))` is false) and ISO-8601 timestamps pass through untouched. The docstring now documents the guard rather than the omission. So `-5` stays `-5` but `=cmd` becomes `'=cmd`, exactly the compromise you described.

ui/ops/src/util/downloadCSV.js
deanredfern26: Agreed it's worth locking down. Turns out `ui/ops` has no test harness at all yet — no `vitest`/`jest` dep, no `test` script, no `.spec` files, and `ui-lint.yml` only lints/builds (no test job) — so a test can't be added in isolation without first standing up a runner and wiring it into CI. That's meaningful infra I'd rather not bootstrap on a security patch, so I've split it out into **SCB-1400** (introduce vitest + cover `escapeCSVField`) with the exact cases listed: bare/comma/quote/CRLF/nullish plus the numeric-vs-formula guard. https://vanti.youtrack.cloud/issue/SCB-1400

ui/ops/src/util/downloadCSV.js
deanredfern26: Thanks — agree it's non-blocking. Leaving both paths on `\n`: it's pre-existing, Excel/Sheets parse it fine, and "RFC 4180" in the docstring is specifically about field escaping/quoting, not the record separator. Happy to switch the separators to `\r\n` under a separate change if we ever want strict RFC 4180 record framing, but it's out of scope for this escaping fix.

mattnathan (approved):
Approving. The formula-injection guard in `0ae9f2122` resolves the blocking concern — I verified the behaviour directly: `=cmd|…`, `=HYPERLINK(…)`, `@SUM(…)` and payload-bearing `+`/`-` values are neutralised with a leading `'`, while genuine numerics (`-5`, `+5`, `-5.5`, `-1e3`) and timestamps pass through untouched. The prefix correctly sits inside the quote-wrapping, the char set matches the OWASP set, and the file lints clean.

Unit-test coverage is reasonably deferred to SCB-1400 — confirmed `ui/ops` has no test harness today (no test script/deps/specs, no CI test job), so standing one up is out of scope for a security patch. Trusting that ticket to follow. CRLF record separators remain out of scope as discussed.

Nice, thorough turnaround on the review points.

PR #912, opened 2026-07-17
task: return error on duplicate Group.Spawn tag instead of panicking (SCB-1386)

deanredfern26:
`Group.Spawn` panicked with `"tag already present"` whenever a tag collided. Since tags can be caller- or data-driven, a duplicate would crash the whole process rather than failing gracefully — a latent footgun for anyone who starts using `Controller.Tasks` (the field is wired up but not yet called anywhere today).

`Spawn` now returns `(*GroupTask, error)`, yielding an error that wraps the new sentinel `ErrTagExists` on a duplicate and leaving the existing task untouched. Returning an error (rather than auto-renaming the tag or silently returning the existing task) keeps a keyed group's contract explicit; with no current callers the signature change has no blast radius.

Verified by `go build ./...`, `go vet`, `staticcheck`, and `go test ./pkg/task/` (new `TestGroup_Spawn` and `TestGroup_Spawn_duplicateTag` cover the happy path and the duplicate-tag error). Race pass deferred to CI (no local C toolchain).

SCB-1386

### DIFF
--- a/pkg/task/group.go
+++ b/pkg/task/group.go
@@ -2,18 +2,26 @@ package task
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"maps"
 	"sync"
 	"time"
 )
 
+// ErrTagExists is returned by Group.Spawn when the group already contains a task
+// with the requested tag.
+var ErrTagExists = errors.New("tag already present")
+
 type Group struct {
 	m     sync.Mutex
 	tasks map[string]*GroupTask
 }
 
-func (g *Group) Spawn(ctx context.Context, tag string, task Task, options ...Option) *GroupTask {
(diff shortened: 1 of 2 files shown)


hexaglow (approved):
This is an improvement. 

I have the following thoughts:
- This is a public package and changes the API contract. Not a large one granted. You are better placed to know how many of our other repos use this API, so I'll assume you've checked and it's OK
- Any thoughts on the (removed) TODO comment idea, about autogenerating a nonconflicting name? Would that be better or worse for your use case?

deanredfern26:
I think just returning the error and leaving it up to the caller to decide what to do is fine for now, though either is better than a panic. I don't actually have a use case in mind, just clearing up a few `todos` in the code base.

PR #916, opened 2026-07-20
auth: document the permitted signature algorithm list (SCB-1383)

deanredfern26:
Clears the TODO on `keycloak.DefaultPermittedSignatureAlgorithms` (`internal/auth/keycloak/config.go`), which asked for the permitted signature algorithms to be narrowed for all Keycloak installations.

**The answer is that it shouldn't be narrowed**, so this PR records why rather than changing behaviour. An earlier revision of this PR did drop HS256; @hexaglow correctly pointed out that the key-confusion vector I cited doesn't apply to go-jose, and on checking the library I agree — see the discussion below.

## Why narrowing isn't warranted

go-jose picks its verifier from the **Go type of the key**, not from the token's `alg` header (`newVerifier`, `signing.go`), and each verifier rejects algorithms its key cannot carry: an HS256 token checked against an RSA public key fails in `asymmetric.go`, and the reverse fails in `symmetric.go`. The algorithm is bound to the key regardless of what the allow-list permits, so the RS256→HS256 key-confusion forgery is already blocked. Narrowing the list would be hardening with no attack to prevent, weighed against a behavioural change for any installation whose Keycloak signs with HS256.

## What changed

- **`internal/auth/keycloak/config.go`** — the TODO is replaced with documentation covering: why the list exists (`jose.ParseSigned` takes an allow-list as a required argument), why HS256 is in a list named for Keycloak (the self-issued access-token path shares it and signs with an HMAC key), why the breadth is safe, and the one condition that would overturn that argument — a key set serving a symmetric `oct` key, where this list would become the only defence. **The algorithm list itself is untouched.**
- **`pkg/system/authn/factory.go`** — a comment at the call site that borrows the Keycloak-named list, since that line is what invites the "surely this is wrong" reading.
- **`pkg/auth/jwks/local_test.go`** — `TestLocalKeySet_VerifySignature_keyConfusion` forges an HS256 token using the published RSA public key as the HMAC secret, with a matching `kid`, and asserts it's rejected. It **permits HS256 deliberately**, so it proves go-jose's key binding blocks the forgery rather than proving the allow-list does. It also asserts the rejection happened at signature verification and not at key lookup — a forgery with a non-matching `kid` fails with `ErrKeyNotFound` before the signature is ever checked, which would make the test vacuous.

## No behavioural change

`git diff main -- internal/auth/keycloak/config.go` is comment lines only. No token that verifies today stops verifying, so the earlier request to confirm no deployment signs Keycloak tokens with HS256 no longer gates this.

## Verification

`go build ./...`, `go vet -lostcancel=false`, and `go test ./internal/auth/... ./pkg/auth/... ./pkg/system/authn/...` all pass. The new forgery test was mutation-checked: breaking the `kid` makes it fail with "rejected at key lookup, so it never reached signature verification", confirming it isn't passing vacuously. (One pre-existing Windows-only file-permissions test fails locally; it passes on CI/Linux.)

SCB-1383


hexaglow (commented):
I assume that by "classic RS256→HS256 key-confusion forgery" you're referring to an attack which tricks a JWT library into validating a signature using the incorrect algorithm, like the attack in [this blog post](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/).

This isn't applicable for go-jose, which [uses different validator types depending on the key used for verification](https://github.com/go-jose/go-jose/blob/v4.1.4/signing.go#L178), and [blocks attempts to verify signatures with algorithms incompatible with that key](https://github.com/go-jose/go-jose/blob/v4.1.4/asymmetric.go#L320). Therefore it can't fall victim to that kind of bug. If an HS256 JWT is supplied, and the key in use is RS256, then the validation will correctly fail.

It seems the design of go-jose provides adequate protection. Is this change intended to be extra hardening? The benefit is unclear to me. We could benefit from a test proving that the algorithm confusion scenario is blocked, though. Or if I'm missing the point and there is a bug here, a test would show me.

deanredfern26:
You're right, and thanks for pushing on it — I've reworked the PR so it no longer changes behaviour.

I went and read the library rather than take either of our word for it, and your reading holds. `newVerifier` (`signing.go:178-205`) dispatches on the Go type of the verification key, so an `*rsa.PublicKey` always gets `rsaEncrypterVerifier`, and `rsaEncrypterVerifier.verifyPayload` (`asymmetric.go:309-321`) returns `ErrUnsupportedAlgorithm` for anything outside RS\*/PS\*. It's symmetrical on the other side: `symmetricMac.verifyPayload` (`symmetric.go:482-501`) can't compute an HMAC for a non-HS alg. The algorithm is bound to the key regardless of the allow-list, so an HS256 token can't be forged against a JWKS public key. `DetachedVerify` then collapses that to `ErrCryptoFailure`, which is what a caller actually sees.

So the key-confusion framing in my original description was overstated, and that was mine, not the ticket's — SCB-1383 is a TODO-cleanup task whose own wording is just "widens the trust surface". There's no live vulnerability here, and I shouldn't have implied one.

Since the point was only ever to clear the TODO, the PR now does that with documentation instead. The algorithm list is byte-identical to `main` (`git diff main -- internal/auth/keycloak/config.go` is comment lines only), so the behavioural change is gone along with the request to confirm nobody signs Keycloak tokens with HS256. Worth noting the two are linked: HS256 is only in a Keycloak-named list because `pkg/system/authn/factory.go` borrows it for the self-issued access-token path, which signs with an HMAC key — so dropping HS256 and leaving the list shared was never coherent. The comment now records that coupling, your go-jose argument, and the one condition that would overturn it: a key set serving a symmetric `oct` key, where the allow-list *would* become the only thing between a published secret and a forged token. Keycloak publishes public keys only and `jwks.RemoteKeySet` is the sole production consumer, so that can't currently arise.

On the test — agreed, and it's the most useful part of this. `TestLocalKeySet_VerifySignature_keyConfusion` forges an HS256 token using the published RSA public key as the HMAC secret. Two details worth flagging:

- It **permits HS256** in the allow-list, so what blocks the forgery is go-jose's key binding, not our parse-time gate. That's the property the comment claims, so that's the one the test should pin.
- It asserts the rejection came from signature verification and *not* `ErrKeyNotFound`. My previous version signed without a `kid`, so it died at key lookup and never checked a signature against the RSA key at all — it would have passed no matter what go-jose did. I mutation-checked the replacement: breaking the `kid` fails it with "rejected at key lookup, so it never reached signature verification".

Also filed SCB-1393 off the back of this for something adjacent I noticed: `accesstoken.Source.SignatureAlgorithms` has no default, so an `accesstoken.Server` built without `WithPermittedSignatureAlgorithms` would pass an empty allow-list to `ParseSigned` and reject every token. Not reachable today — `factory.go` is the only constructor and always passes it — but a latent footgun.

internal/auth/keycloak/config.go
```diff
@@ -17,8 +17,28 @@ func (c *Config) Issuer() string {
...
+// asymmetric.go and the reverse fails in symmetric.go. The algorithm is bound to the key whatever
+// this list says. jwks.TestLocalKeySet_VerifySignature_keyConfusion forges such a token and asserts
+// it is rejected; narrowing this list is therefore hardening with no attack to prevent.
+//
+// That argument holds only while every key reaching a verifier on the JWKS path is asymmetric.
+// Keycloak publishes public keys only, and jwks.RemoteKeySet is the sole production consumer. Should a
+// key set ever serve a symmetric ("oct") key, go-jose would verify it with an HMAC verifier and this
+// allow-list would become the only thing between a published secret and a forged token — at which
+// point the two paths need separate lists and HS256 must come out of this one.
 var DefaultPermittedSignatureAlgorithms = []string{
```
hexaglow: This is too long. All we need is a note that tells the reader that there's no problem here like they might assume. 1 or 2 sentences max. e.g.  "tokens will only pass verification if their alg is appropriate for the key type in use, because of the design of go-jose, so we're not vulnerable to algorithm confusion attacks as described in [link to post]"

pkg/system/authn/factory.go
```diff
@@ -75,6 +75,8 @@ func (s *System) applyConfig(ctx context.Context, cfg config.Root) error {
 	var serveTokenEndpoint bool
 	tokenServerOpts := []accesstoken.ServerOption{
 		accesstoken.WithLogger(s.logger.Named("server")),
+		// These tokens are self-issued rather than Keycloak's, but the two verification paths share
+		// one allow-list; HS256 is in it for this path. See keycloak.DefaultPermittedSignatureAlgorithms.
 		accesstoken.WithPermittedSignatureAlgorithms(keycloak.DefaultPermittedSignatureAlgorithms),
 	}
```
hexaglow: You can put HS256-only here instead. That was fine, what I objected to before was the framing/reasoning for the change. It does make sense to have HS256-only here, but it's just a code cleanup thing rather than security hardening which was how the previous diff described it. Sorry if that wasn't clear.

That would also remove the need for the comment.

pkg/auth/jwks/local_test.go
```diff
@@ -35,3 +36,74 @@ func TestLocalKeySet_VerifySignature(t *testing.T) {
...
 	}
 }
+
+// TestLocalKeySet_VerifySignature_keyConfusion attempts the RS256->HS256 key-confusion forgery: the
+// attacker takes the RSA public key we publish and uses its bytes as an HMAC secret, hoping the
+// verifier will pick its algorithm from the token's alg header rather than from the key. HS256 is
+// deliberately permitted here, so what blocks the forgery is go-jose binding the algorithm to the key
+// type (newVerifier picks rsaEncrypterVerifier for an RSA key, which rejects a non-RS/PS alg), not
+// our allow-list. keycloak.DefaultPermittedSignatureAlgorithms relies on that property to permit
+// HS256 for the self-issued token path without weakening the JWKS path.
```
hexaglow: Would appreciate a URL reference describing what the problem we're trying to detect is - a link to a good description of the vulnerability is more useful than trying to explain it inline.

hexaglow (changes requested):

internal/auth/keycloak/config.go
deanredfern26: Cut to five lines: the list stays broad because Keycloak picks its own default suite, and one sentence saying that breadth is not an algorithm confusion exposure because go-jose picks its verifier from the key type, with your auth0 link doing the explaining.

HS256 is also now removed from the list, which is a change since you last looked. See the summary comment for why.

pkg/system/authn/factory.go
deanredfern26: Resolved by #933, which landed on main last week. It removed `WithPermittedSignatureAlgorithms` and `Source.SignatureAlgorithms` outright and derives the permitted algorithm from `Source.Key.Algorithm`, so `factory.go` gets HS256-only structurally rather than by configuration. The line you commented on no longer exists and `factory.go` has dropped out of this diff entirely.

pkg/auth/jwks/local_test.go
deanredfern26: Done, the doc comment is now four lines pointing at the auth0 post you linked rather than explaining the attack inline.

deanredfern26:
Rebased onto main and reworked, because #933 changed the ground this PR was standing on.

**What #933 did.** It removed `accesstoken.WithPermittedSignatureAlgorithms` and `Source.SignatureAlgorithms` entirely, deriving the permitted algorithm from `Source.Key.Algorithm` instead. That deleted the `factory.go` line this PR was annotating (the one conflict in the rebase, resolved by taking main), and it broke the previous version's central claim — that HS256 sits in a Keycloak-named list because the self-issued token path borrows the list. That coupling is gone. On main, `keycloak.DefaultPermittedSignatureAlgorithms` has exactly one consumer: `jwks.NewRemoteKeySet` at `internal/auth/keycloak/token.go:84`.

**So HS256 comes out.** #933's own commit message left the decision here: *"Whether HS256 should also come out of the Keycloak list is left to that PR."* Its only justification was the path that just went away, and the path that remains is Keycloak's JWKS endpoint, which publishes asymmetric public keys and never the realm `hmac-generated` secret — so the entry is configuration nothing can reach. This is tidying up after #933, not hardening, which I think is the "code cleanup thing" distinction you drew last time. The asymmetric families stay broad, deliberately, for the reason the original TODO gave: Keycloak picks its own default suite. That closes the TODO rather than leaving it.

**The comment** is down from ~20 lines to five, and links your auth0 post rather than restating the attack. Same for the test's doc comment.

**Diff is now two files**, `internal/auth/keycloak/config.go` and `pkg/auth/jwks/local_test.go`. `factory.go` is byte-identical to main and has dropped out.

`Go Security Scan` was failing on the old head with `GO-2026-6061` in gRPC; main fixed that in 258c9ab05, so the rebase clears it.

Both tests still pass, and I re-ran the mutation check on `_keyConfusion` — breaking the `kid` fails it with "rejected at key lookup, so it never reached signature verification" rather than passing silently.

PR #920, opened 2026-07-20
electric: rename internal MemorySettings proto package to match location (SCB-1389)

deanredfern26:
The `MemorySettings` proto for in-memory electric devices still declared
`package smartcore.go.trait.electric` — a leftover from when the file lived
elsewhere. It now sits under `smartcore/bos/electric/internal/` with a
`go_package` of `electricpb`, so the package name no longer matched its
location or its siblings (`smartcore.bos.electric.v1`), which the file's
TODO flagged.

Renamed the package to `smartcore.bos.electric.internal` (mirroring the file
path and sitting alongside `electric.v1`) and regenerated the Go and grpc-web
output. Message wire format is unaffected by the proto package name; the only
behavioural change is the gRPC method path, and this is a testing-only API
with no external clients. All hand-written consumers use the unchanged
`electricpb` Go symbols.

Verified by `go build ./...`, `go vet`, and `go test ./pkg/proto/electricpb/...`
(incl. the MemorySettingsApi model-server test); proto regenerated via the WSL
toolchain and committed.

#SCB-1389 State Done


hexaglow (approved):
Agree that we can change this despite the incompatibility, as it's an internal API.

PR #921, opened 2026-07-21
auto: validate configured trait against the anytrait registry (SCB-1385)

deanredfern26:
The `healthbounds` and `setpointhealth` automations take a free-form `trait.Name` in their source config, but only traits registered in the shared `anytrait` registry actually work. A misspelled or unsupported trait didn't fail at config-load — it surfaced late, per device, as an opaque `anytrait.ErrNotFound` that got logged and skipped, so checks silently never ran.

This validates the configured trait against the registry at `config.Read` time and returns a clear error naming the unsupported trait. `anytrait` gains a `Validate` that defers to the registry, so it stays the single source of truth (membership is checked, not duplicated). `healthbounds` had no validation step and gets one; `setpointhealth` already had a `Validate` that only checked the trait was non-empty, so it's extended with the same registry-membership check — fixing the whole class rather than just the one automation named in the ticket.

Verified by `go test ./pkg/auto/internal/anytrait/... ./pkg/auto/healthbounds/... ./pkg/auto/setpointhealth/...` (new regression tests: an unsupported trait is now rejected by `Read` in both automations), plus `go build ./...`, `go vet`, and `staticcheck` on the touched packages.

#SCB-1385 State Done

### DIFF
--- a/pkg/auto/internal/anytrait/registry.go
+++ b/pkg/auto/internal/anytrait/registry.go
@@ -3,6 +3,9 @@ package anytrait
 import (
 	"context"
 	"errors"
+	"fmt"
+	"slices"
+	"strings"
 	"sync"
 
 	"google.golang.org/grpc"
@@ -32,6 +35,16 @@ func (r *Resolver) FindByName(name trait.Name) (Trait, error) {
 		return Trait{}, ErrNotFound
 	}
 	return t, nil
+}
+
+// Names returns the sorted list of trait names known to the resolver.
+func (r *Resolver) Names() []trait.Name {
+	names := make([]trait.Name, 0, len(r.byName))
+	for name := range r.byName {
+		names = append(names, name)
+	}
+	slices.Sort(names)
+	return names
 }
 
 func (r *Resolver) add(name trait.Name, resources ...Resource) {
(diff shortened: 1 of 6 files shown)


pkg/auto/internal/anytrait/registry.go
```diff
@@ -100,6 +113,26 @@ func FindByName(name trait.Name) (Trait, error) {
...
+// Validate returns a descriptive error if name is not a trait supported by the registry.
+// The error lists the supported trait names.
+func Validate(name trait.Name) error {
+	if _, err := FindByName(name); err != nil {
+		names := Names()
+		supported := make([]string, 0, len(names))
+		for _, n := range names {
+			supported = append(supported, string(n))
+		}
+		return fmt.Errorf("%q is not a supported trait; supported traits are: %s", name, strings.Join(supported, ", "))
```
hexaglow: There are a lot of traits. This is going to produce a very long error message, longer than any other error string I can think of in BOS. I recommend against this.

hexaglow (changes requested):

pkg/auto/internal/anytrait/registry.go
deanredfern26: good point 😂

deanredfern26: Dropped the supported-trait list from the error, so it's now just `"<name>" is not a supported trait`. That left `anytrait.Names()` / `Resolver.Names()` with no production caller — they only existed to build that list — so I removed them and the `TestNames` that tested them, and trimmed `TestValidate` accordingly.

hexaglow (approved):

PR #922, opened 2026-07-21
auto/bms: add exponential backoff to write retries (SCB-1384)

deanredfern26:
A BMS automation that fails to write to a device was retried at a fixed
rate (`WriteRetryDelay`, default 1m). Against a persistently failing or
unreachable device this adds needless load and log noise, with no relief
under sustained failure.

Retries now back off exponentially from `WriteRetryDelay`, capped at a new
`WriteRetryMaxDelay` (default 15m), and reset to the base after the next
successful process. Equal jitter is applied so devices sharing an
upstream gateway don't retry in lockstep when it recovers; because of
the jitter each retry lands in `[delay/2, delay)`, so `WriteRetryDelay`
acts as an upper bound on the wait rather than an exact delay. Existing
config keys still apply unchanged. The never-wired
`retryFailedProcessing` scaffolding in the processing loop was removed
while here.

Verified by `go test ./pkg/auto/bms/...`: a unit test for the backoff
maths (growth, cap, jitter bounds, edges) and an integration test that
drives the real processing loop, asserting the retry TTL grows across
consecutive failures and resets after a success.

#SCB-1384 State Done

### DIFF
--- a/pkg/auto/bms/config/root.go
+++ b/pkg/auto/bms/config/root.go
@@ -37,7 +37,7 @@ type Root struct {
 	WriteCacheExpiry *jsontypes.Duration `json:"writeCacheExpiry,omitempty"` // Defaults to no expiry.
 	// If processing fails, how long to wait before trying again.
 	// This is the base delay; consecutive failures back off exponentially (with jitter) up to WriteRetryMaxDelay.
-	WriteRetryDelay *jsontypes.Duration `json:"WriteRetryDelay,omitempty"` // Defaults to 1m.
+	WriteRetryDelay *jsontypes.Duration `json:"writeRetryDelay,omitempty"` // Defaults to 1m.
 	// The maximum delay between retries when processing keeps failing.
 	// Backoff grows from WriteRetryDelay and is capped at this value.
 	WriteRetryMaxDelay *jsontypes.Duration `json:"writeRetryMaxDelay,omitempty"` // Defaults to 15m.
--- a/pkg/auto/bms/config/root.go
+++ b/pkg/auto/bms/config/root.go
@@ -36,7 +36,11 @@ type Root struct {
 	// How long should a past write affect future writes.
 	WriteCacheExpiry *jsontypes.Duration `json:"writeCacheExpiry,omitempty"` // Defaults to no expiry.
 	// If processing fails, how long to wait before trying again.
+	// This is the base delay; consecutive failures back off exponentially (with jitter) up to WriteRetryMaxDelay.
 	WriteRetryDelay *jsontypes.Duration `json:"WriteRetryDelay,omitempty"` // Defaults to 1m.
+	// The maximum delay between retries when processing keeps failing.
+	// Backoff grows from WriteRetryDelay and is capped at this value.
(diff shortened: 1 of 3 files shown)


pkg/auto/bms/config/root.go
```diff
@@ -36,7 +36,11 @@ type Root struct {
 	// How long should a past write affect future writes.
 	WriteCacheExpiry *jsontypes.Duration `json:"writeCacheExpiry,omitempty"` // Defaults to no expiry.
 	// If processing fails, how long to wait before trying again.
+	// This is the base delay; consecutive failures back off exponentially (with jitter) up to WriteRetryMaxDelay.
 	WriteRetryDelay *jsontypes.Duration `json:"WriteRetryDelay,omitempty"` // Defaults to 1m.
```
mattnathan: Every other json tag in this struct is camelCase (`writeCacheExpiry`, `writeEvery`, `writeReadPropagation`), and your new `writeRetryMaxDelay` follows suit — but this sibling tag is `WriteRetryDelay` (capital W). It's pre-existing, but the result is two directly-related config keys with different casing. Worth correcting this tag to `writeRetryDelay` while you're here.

pkg/auto/bms/factory.go
```diff
@@ -138,17 +140,17 @@ func (a *Auto) processReadStates(ctx context.Context, readStates <-chan *ReadSta
...
-	var retryFailedProcessing <-chan time.Time
-	cancelRetryTimer := func() bool { return false }
-
 	// writeState is only accessed from this go routine.
 	writeState := NewWriteState()
 	writeState.Now = a.now
 
+	// consecutiveFailures counts how many times processing has failed in a row.
+	// It grows the retry backoff and resets to 0 on the next successful process.
+	var consecutiveFailures int
```
mattnathan, line 149: `consecutiveFailures` is a single counter shared across everything this loop handles. That's consistent with the loop only ever reprocessing `lastReadState`, so it's correct as written — just flagging that if `readStates` ever interleaves genuinely distinct devices, a success on one would reset another's backoff.

mattnathan (approved):
Approving. The backoff logic is sound and nicely covered by both a unit test for the maths and an integration test driving the real loop.

Two small inline comments below; neither is blocking. One doc note: the description says the base delay means "existing config still applies", but with equal jitter the effective first retry lands in `[base/2, base)` — so `WriteRetryDelay` becomes a ceiling rather than the exact delay. That's fine behaviour (changing a retry timeout is harmless); the PR wording is just slightly off.

pkg/auto/bms/config/root.go
deanredfern26: Done — corrected the tag to `writeRetryDelay`. Go's JSON decoding matches keys case-insensitively, so any existing config using the capital-W key still loads unchanged.

pkg/auto/bms/factory.go
deanredfern26: Agreed — the single counter is correct precisely because the loop only ever reprocesses `lastReadState`, so there's only ever one in-flight backoff to track. If `readStates` ever interleaved distinct devices we'd need to key the counter per device; leaving it as-is for now since that isn't the current shape.

PR #924, opened 2026-07-23
driver/paxton: migrate Paxton Net2 driver from hbd-island (SCB-1123)

deanredfern26:
We told Paxton that our Net2 access-control driver is open source, but it lived in the closed `hbd-island` repo — so the claim wasn't yet true. This moves the driver into sc-bos, where it's public and maintained alongside the other access-control drivers (gallagher, hikcentral).

The driver was already written entirely against sc-bos primitives, so this is a near-verbatim move: the only internal import rewritten is its own `config` sub-package, and it registers in `alldrivers` like its peers. One behaviour change — the deployment-specific `securityEventsName` default (`hbd-island/bc-01/...`) is dropped in favour of requiring the field when `enableSecurityEvents` is true, so no hbd-island-specific name is baked into open source. A README documents config, the OAuth2 auth model, and the polling-vs-SignalR event toggle.

Follow-up (separate, in hbd-island): bump the sc-bos dependency, delete the local `internal/driver/paxton`, drop the bespoke factory wiring in favour of `alldrivers`, and set `securityEventsName` explicitly in deployment config.

Verified by `go build ./...`, `go vet`, and `go test ./pkg/driver/paxton/...` — the ported tests (dedup cache, event processing/sourcing, ListSecurityEvents pagination, SignalR parse/negotiate/stream) all pass.

#SCB-1123 State Done


pkg/driver/paxton/README.md
```diff
@@ -0,0 +1,67 @@
...
+  default; enable with `enableSignalR`.
+
+At least one event source must be active when `enableSecurityEvents` is true (i.e. you
+cannot set `disablePolling: true` and `enableSignalR: false` together).
+
+## Set up
+
+- The Net2 server must have its API enabled and an OAuth2 client registered (`clientId`).
+- Create an operator account for the driver and place its password in a file readable by
+  the building controller — referenced by `auth.passwordFile` (the password is never read
```
hexaglow, line 24: > the building controller

Could be another kind of node, suggest just 'Smart Core BOS'

hexaglow (approved):
I haven't performed a full re-review of this on the basis it's being moved from another repo.

PR #925, opened 2026-07-23
auto/udmi: event/pointset points list export (SCB-1397)

deanredfern26:
Building integrators need a per-device points list for UDMI-enabled kit — which points a device actually exposes — to hand over, commission against, and reconcile with the design. Until now that meant reading it off the MQTT broker by hand. This adds a first-class export.

The udmi automation taps every pointset event it publishes and keeps the latest one per (source device, topic). A new `UdmiExportApi.ListExportedPoints` turns that snapshot into points list rows — `source_name`, `topic`, `asset_name`, `points` — and a cohort-wide **Download points list** button on **System > Components** fans out across every node and writes the CSV.

`UdmiExportApi` is announced against the **node's** name, following the systems that expose node-scoped APIs that way. A node normally runs one udmi automation, but if it runs more (the reason would be publishing to two brokers) they share one export server and the node reports the points of all of them together — so no automation's points go missing.

### Where the work happens

Everything that needs to understand UDMI — pointset-topic matching, point-name parsing (both the conformant `{points: {...}}` envelope and the bare map the mock driver emits), BDNS functional asset-name extraction — is in Go, in `pkg/auto/udmi/export.go`, and unit tested. The ops UI receives typed rows and does no UDMI parsing: it de-duplicates by source+topic across nodes, widens the table to the device with the most points, and hands the rows to the shared `downloadCSVRows` util that the notifications and subsystem-health exports already use.

Serving the finished CSV as a file over a signed download URL, the way devices/history downloads work, is tracked separately in [SCB-1398](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1398). That is a scale-up path for large exports reusing this same Go logic, not a second implementation of it — and it needs two things this PR doesn't: a `download.Router` handle on `auto.Services` (it's currently only on `system.Services`), and hub-side aggregation, because a signed URL points at the issuing node's own HTTP endpoint and the ops UI reaches child nodes only over grpc-web routed through the hub.

### What this is — and isn't

The CSV lists the **event/pointset points for each device implementing the UDMI trait**, derived purely from what the automation is *sending to the broker*. It is deliberately **not** a definitive catalogue of every point a device could have:

- A point only appears once the device has **published a pointset event carrying it** since the automation last (re)started.
- Only pointset-event topics contribute; state, metadata and other topics are published as normal but aren't collected.
- A device that hasn't reported yet, or a point that's declared but never sent, won't show.

That limitation is a conscious trade for **simplicity**: capturing at the publish tap needs no extra device queries and no separate discovery protocol, and it guarantees the export is exactly "what we're sending to the broker" — which is the thing integrators actually reconcile against. "Only what's been published" is the price of that simplicity, and we think it's the right balance for a hand-over/commissioning aid.

Verified by `go build ./... && go vet ./...`, `go test ./...` (no new failures against a clean-tree baseline), and ops-UI lint (0 errors, only the pre-existing jsdoc warnings). The export logic — topic classification, point parsing, BDNS name extraction, multi-automation merging — is unit tested in Go, including an end-to-end pass through the node router. End-to-end functional test is being run on site by the author.

#SCB-1396 State Done
#SCB-1397 State Done


hexaglow (commented):
This implementation has the CSV generated on the client side. By contrast, our devices (incl. history) download generates the CSV file on the server. Was this alternative considered in the design, and if so, why was client-side generation selected?

The new API is announced under the name of the automation. I don't think we have precedent for this. Would it make sense to announce on the node instead? I can see that might create challenges if there were multiple instances of the UDMI auto per node - is that a common scenario? Related APIs:
- The devices API uses a non-routed service to provide downloads
- The Parent trait is announced on the node name

deanredfern26:
- Was this alternative considered in the design, and if so, why was client-side generation selected
[SCB-1398](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1398) Has a follow up ticket for server side

- if there were multiple instances of the UDMI auto per node - is that a common scenario?
Not something we have ever done. Only reason that would happen is if we use it to publish to 2 different MQTT brokers. Not implausible this could happen though, I will look at your suggested alternative

deanredfern26:
Followed up on the announcement point in 17b3ff8 — `UdmiExportApi` is now announced against the **node's** name.

You were right that the automation-name announcement matched nothing: systems announce node-scoped APIs on `Node.Name()` (`system/history`, `system/log`, `system/alerts`, …), and automations announce against a *source* device name (`auto/history`). An automation announcing its own API under its own config name was novel.

On the multi-instance risk you flagged — it was worse than just awkward. `Node.Announce` → `router.AddRoute` returns `ErrRouteExists` on a duplicate, and `announceLocked` only *logs* it, so a second udmi automation's points would have silently vanished from the export. Rather than accept that, each node now gets one export server that merges the collectors of every udmi automation on it, so the node reports all of them together. Records are keyed by (source, topic) and merged, since `ExportedMessage` has no broker field to tell two brokers apart with anyway.

This also paid for itself in the UI: the cohort export no longer lists each node's automations and parses `configRaw` for a name — it's one request per node, and the "automation has no configured name" failure path is gone.

I did look at the non-routed-service route you pointed at (`node.RegistryService` + `Node.AnnounceService`, as `system/hub` and `system/tenants` use). Not taking it yet, for three reasons:

1. A service-only default route matches **any** key (`router.ResolveMethod` tries `{service,key}` → `{"",key}` → `{service,""}` → `{"",""}`). On a gateway, a request naming something with no proxy route would silently return the *gateway's own* points instead of `NotFound`. Announcing on the node name keeps addressing exact.
2. It wouldn't have fixed the multi-instance case either — `AddRoute(service, "")` also returns `ErrRouteExists` — so the merging registry was needed regardless.
3. Its coherent form is to drop `name` from the request so the API is genuinely node-scoped, and that currently *breaks* cohort reachability: the gateway only proxies a remote service when it's name-routable (`gateway/announce.go`), and a non-name-routable service on a non-hub node hits the branch that warns "unable to announce non-routable service on non-hub node" and announces nothing. Reaching child nodes would need server-side aggregation on the hub — i.e. SCB-1398.

So it's a natural pairing with the server-side generation work rather than something to do now, and migrating then is a one-call swap inside the registry: no proto change, and no UI change since `name` keeps carrying the node name.

On CSV generation: agreed the server-side approach matches devices/history. [SCB-1398](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1398) tracks it — I've added a note to the PR description pointing at it.

pkg/auto/udmi/exportregistry.go
```diff
@@ -0,0 +1,75 @@
...
+
+// nodeExports announces UdmiExportApi against each node's own name, following the systems
+// that expose node-scoped APIs that way (see pkg/system/history, pkg/system/log).
+//
+// A node normally runs a single udmi automation, but nothing stops it running more — the
+// reason to would be publishing to two brokers. Only one server can hold the route for a
+// name, so all of a node's automations share one, and its snapshot merges their collectors.
+// That way the export covers the whole node however many automations produced it, rather
+// than whichever automation registered first.
+var nodeExports = &exportRegistry{}
```
hexaglow, line 18: This shared global state feels awkward, but the design of how automations interact with nodes kind of requires it at present. Unfortunately I can't think of any better ideas. We don't have a pattern for when a service type needs to associate data with a node but shared between all instances of that service type. 

(no changes required)

pkg/auto/udmi/export.go
```diff
@@ -0,0 +1,209 @@
...
+				existing.lastSeen = rec.lastSeen
+			}
+			if rec.firstSeen.Before(existing.firstSeen) {
+				existing.firstSeen = rec.firstSeen
+			}
+			existing.count += rec.count
+		}
+	}
+	out := slices.Collect(maps.Values(byKey))
+	sort.Slice(out, func(i, j int) bool {
```
hexaglow: Prefer `slices.SortFunc`

ui/ops/src/routes/automations/components/config-cards/UdmiExportCard.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <v-card elevation="0" tile>
+    <v-list tile class="ma-0 pa-0">
+      <v-list-subheader class="text-title-caps-large text-neutral-lighten-3">Points list</v-list-subheader>
+      <v-list-item>
+        <v-list-item-subtitle class="text-wrap">
+          Download the distinct messages this node has published since its udmi automations last
```
hexaglow: Does it still make sense to have this card in the automation sidebar now that it's a node-scoped API? Given that all such cards will return the same data.

proto/smartcore/bos/udmi/v1/udmi.proto
```diff
@@ -59,3 +61,45 @@ message MqttMessage {
...
+  string message_type = 3;
+  // payload is the most recent JSON payload published for this topic.
+  string payload = 4;
+  // first_seen is when a message for this topic was first published since the automation (re)started.
+  google.protobuf.Timestamp first_seen = 5;
+  // last_seen is when the most recent message for this topic was published.
+  google.protobuf.Timestamp last_seen = 6;
+  // count is the number of messages published for this topic since the automation (re)started.
+  int64 count = 7;
+}
```
hexaglow: `message_type`, `first_seen`, `last_seen` and `count` are never read by the client, so the effort to generate them is unnecessary.

hexaglow (changes requested):
Good solution to the multiple instances problem.

I understand that server-side generation is captured in a separate ticket, but I don't see why that implementation would be more complex than a client-side implementation. Therefore we can just implement it server side now. It doesn't make sense to create 2 implementations.

pkg/auto/udmi/exportregistry.go
deanredfern26: Agreed on both counts — kept as is. Noting for anyone reading later: the awkwardness is contained to `Add`/`remove` keying off the `*node.Node` pointer, and the registry is the only thing that would need to change if a general "service type with per-node shared state" pattern ever lands.

pkg/auto/udmi/export.go
deanredfern26: Done in 898258f. Both `sort.Slice` sites are gone — `Snapshot` uses `slices.SortFunc`, and the collect-then-sort in `mergeRecords` collapses to a single `slices.SortedFunc(maps.Values(byKey), …)`. Shared comparator, `cmp.Or` for the topic-then-source ordering.

ui/ops/src/routes/automations/components/config-cards/UdmiExportCard.vue
deanredfern26: No, it doesn't — card removed in 898258f, along with its registration in `AutomationSideBar.vue` (that file is now identical to main). The cohort-wide button on **System > Components** is the single entry point, and `pointsExport.js` moved from `routes/automations/` to sit next to it in `routes/system/components/`, since nothing under automations uses it any more.

Removing the second call site also made the server-side change below cheaper.

proto/smartcore/bos/udmi/v1/udmi.proto
deanredfern26: All four gone in 898258f, and `payload` with them — the message is now `DevicePoints{source_name, topic, asset_name, points}`, i.e. a finished points list row rather than a raw MQTT message. The `google/protobuf/timestamp.proto` import this PR added is no longer needed either.

One survives unexported: `exportRecord.lastSeen`. It isn't generated for the client — it only orders payloads when merging the collectors of two automations publishing the same device, so the merge is deterministic rather than dependent on registration order.

deanredfern26:
Agreed — two implementations was the wrong plan. Moved server-side in 898258f.

`ListExportedPoints` now returns finished rows rather than raw MQTT messages:

```proto
message DevicePoints {
  string source_name = 1;
  string topic = 2;
  string asset_name = 3;      // BDNS functional asset name, parsed from topic
  repeated string points = 4; // sorted
}
```

Everything that has to understand UDMI moved into `pkg/auto/udmi/export.go`: pointset-topic matching, point-name parsing (both the conformant `{points: {...}}` envelope and the bare `present_value` map the mock driver emits), and BDNS asset-name extraction. The collector also drops non-pointset topics at the tap now, instead of retaining state/metadata payloads that were never exported. `pointsExport.js` went from 117 lines of UDMI parsing to ~60 with none: de-duplicate by source+topic across nodes, widen the table to the device with the most points, hand the rows to the shared `downloadCSVRows` util that notifications and subsystem-health already use.

A point in favour of this that I'd missed when I filed SCB-1398: **there is no JS test infrastructure in this repo** — no `*.spec.js`/`*.test.js` anywhere, no `test` script in either `package.json`. So the parsing logic wasn't just duplicated-in-waiting, it was structurally untestable. It now has table tests in Go, including the payload shapes and topic forms that used to be JS-only concerns. (The old PR description claimed those were unit tested; that was wrong and I've corrected it.)

**What I did not do, and why** — the signed-URL download you pointed at, where the server serves the CSV *file*. It needs two things this PR doesn't touch:

1. Automations have no download router. `download.Router` is on `system.Services` (`pkg/system/system.go:36`) only; `auto.Services` (`pkg/auto/factory.go:18`) has no handle, so it means changes in `pkg/auto` and `pkg/app/controller.go`.
2. **It doesn't work per-node.** URLs are `https://<the issuing node's ExternalHTTPEndpoint>/download/…` (`buildDownloadRouter`, `pkg/app/controller.go:547`). The ops UI reaches child nodes only over grpc-web routed through the hub gateway — no UI feature consumes a signed URL from a child today (`system/log` generates them and nothing in `ui/ops` fetches one; devices and audit downloads are always served by the node the UI is connected to). So a cohort-wide file needs hub-side aggregation: a component with cohort-node knowledge, living somewhere other than the automation, since the udmi automation may not run on the hub at all.

That's a genuine design piece rather than a smaller version of this one, so [SCB-1398](https://vanti.youtrack.cloud/projects/SCB/issues/SCB-1398) stays — but re-framed. It's now a scale-up path for large exports that reuses this Go logic, not a second implementation of it. I've updated the ticket and the PR description to say so.

Also in this commit: `slices.SortFunc`/`slices.SortedFunc`, the four unread proto fields removed, and the per-automation sidebar card dropped.

pkg/auto/udmi/export.go
```diff
@@ -1,46 +1,84 @@
...
-			return "event"
-		case seg == "state":
-			return "state"
-		case seg == "metadata" || strings.HasPrefix(seg, "metadata."):
-			return "metadata"
+// pointsetEventRe matches the pointset-event portion of a UDMI topic, covering both the
+// UDMI standard suffix ".../event/pointset/points" and the shorter ".../events/pointset"
+// some drivers use. Whole path segments are matched, so a site or device segment that
+// merely contains "event" isn't mistaken for one.
+var pointsetEventRe = regexp.MustCompile(`/events?/pointset`)
```
hexaglow, line 21: Hardening opportunity for edge cases / adversarial data: 
- doesn't match on an empty prefix without leading slash i.e. bare "events/pointset". This shouldn't occur in practice, your choice on how to handle - use empty string as the asset name or just treat as invalid.
- Matches on something like `foo/events/pointsetfoo`

Could use a more specific regex like `(^|/)events?/pointset(/|$)`

hexaglow (approved):
I can see that a 100% server side implementation is a bit more involved. This is most of the way there. When the follow-up server download hander is added, make sure to remove the client-side one so there's only 1 implementation.

PR #926, opened 2026-07-24
health: bucket range-check deviation and add dashboard filter (SCB-1382)

deanredfern26:
Small range-check infractions — a humidity reading of 60.2% against a 30–60% band, a temperature a fraction over 24°C — flooded the health dashboard and drowned out the excursions operators actually care about. There was no way to filter by *how far* out of range a check sits, and because the measured value is deliberately stripped from embedded device health checks (and the device query engine has no numeric comparison), the magnitude has to be precomputed and bucketed into something the string query can match.

This adds a top-level `HealthCheck.Deviation` enum (`MINOR`/`MODERATE`/`MAJOR`), computed in the shared `healthpb` bounds engine alongside `normality`: the overshoot past the crossed bound as a fraction of the range width (or the bound's own magnitude for open-ended ranges), bucketed at 10% / 25%. Deadband-held states and values that have returned inside the range report `UNSPECIFIED`; open-ended *timestamp* ranges have no meaningful scale so stay unbucketed. The ops dashboard gains a "Deviation" filter — treated as a minimum severity, and OR'd with hard `ABNORMAL` faults so non-range faults are never hidden — plus a severity chip in the check row.

Two deliberate departures from the SCB-1382 sketch: the field is named `Deviation` rather than `Severity`, and the 10%/25% thresholds are fixed in code rather than per-automation config (simpler; easy to revisit). Computing it in `healthpb` rather than the `healthbounds` factory means every bounds producer gets it for free, and the field survives both `removeMeasuredValues` and the history `splitCheck`. The deferred UI unit test is tracked in SCB-1403 (blocked on the vitest harness, SCB-1400).

Verified by `go test ./pkg/proto/healthpb/...` (new deadband + timestamp cases), `go build ./...` + `go vet`, proto regenerated via WSL, ops UI `lint:nofix` clean, and the chip + filter rendered end-to-end in a Vuetify harness (all four chip states plus the filter dropdown).

#SCB-1382 State Done

### DIFF
--- a/pkg/proto/healthpb/bounds.go
+++ b/pkg/proto/healthpb/bounds.go
@@ -4,8 +4,18 @@ import (
 	"cmp"
 	"context"
 	"fmt"
+	"math"
 
 	"google.golang.org/protobuf/proto"
+)
+
+// Deviation buckets are assigned by how far a measured value sits outside its
+// expected range, as a fraction of the range width (or of the crossed bound's
+// magnitude for open-ended ranges). A value at or above deviationMajorRatio is
+// MAJOR, at or above deviationModerateRatio is MODERATE, otherwise MINOR.
+const (
+	deviationModerateRatio = 0.10
+	deviationMajorRatio    = 0.25
 )
 
 // ValidateValueRange checks that a value range is well-formed.
@@ -157,6 +167,7 @@ func (c *BoundsCheck) writeValue(dst *HealthCheck, v *HealthCheck_Value) {
 	oldState := dst.GetNormality()
 	newState := c.checker.valueToState(v, dst)
 	dst.Normality = newState
+	dst.Deviation = c.checker.valueToDeviation(v, newState)
 	out.CurrentValue = v
 	updateStateTimes(dst, oldState, newState)
 }
@@ -184,6 +195,10 @@ func (c *BoundsCheck) prepareChecker(b *HealthCheck_Bounds) (boundsChecker, erro
 // boundsChecker compares a value against a normal state, returning the appropriate check state.
 type boundsChecker interface {
 	valueToState(v *HealthCheck_Value, current *HealthCheck) HealthCheck_Normality
+	// valueToDeviation buckets how far v sits outside the expected bounds, given
+	// the state just computed by valueToState. Checks without a meaningful
+	// magnitude (equality and value-set checks) return DEVIATION_UNSPECIFIED.
+	valueToDeviation(v *HealthCheck_Value, state HealthCheck_Normality) HealthCheck_Deviation
 }
 
 func newNormalValueCheck(v, nv *HealthCheck_Value) (*valueCheck, error) {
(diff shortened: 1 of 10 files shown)


hexaglow (commented):
In my view, it may be more useful to sort health checks by (descending) severity, rather than filtering on them. Dividing abnormal checks into 3 buckets by severity is inherently arbitrary. I don't think any user will be exactly happy with the thresholds, as they'll all have different ideas of what the right thresholds should be. Perhaps having 3 abnormal buckets is too many. I can see a distinction between "OK", "acceptable but not ideal" and "unacceptable" being workable, however. Alternatively, if we have multiple thresholds from e.g. a building spec, we'd want to match that ideally.

Calculating a severity score, rather than an enum, would allow the client to sort by it, as well as filtering. To me, the most useful default view would show all failing health checks in order of most severe to least severe. Filtering would also be possible. As noted, this would require LT/GT operators in the query language. Worth it, IMO. Thoughts?

Nit: the ticket calls the new property `severity`, the proto calls it `deviation`.

deanredfern26:
> In my view, it may be more useful to sort health checks by (descending) severity, rather than filtering on them. Dividing abnormal checks into 3 buckets by severity is inherently arbitrary. I don't think any user will be exactly happy with the thresholds, as they'll all have different ideas of what the right thresholds should be. Perhaps having 3 abnormal buckets is too many. I can see a distinction between "OK", "acceptable but not ideal" and "unacceptable" being workable, however. Alternatively, if we have multiple thresholds from e.g. a building spec, we'd want to match that ideally.
> 
> Calculating a severity score, rather than an enum, would allow the client to sort by it, as well as filtering. To me, the most useful default view would show all failing health checks in order of most severe to least severe. Filtering would also be possible. As noted, this would require LT/GT operators in the query language. Worth it, IMO. Thoughts?
> 
> Nit: the ticket calls the new property `severity`, the proto calls it `deviation`.

The client specifically wanted to be able to hide the checks that were not too important, so things like `60.20 %RH (expected 30.00–60.00 %RH)` don't get shown. We can do both sorting and filtering, but sorting alone is not sufficient, for them or us. We don't want to show 100+ failing range checks when they are really only minor deviations. 

Agree with surfacing the actual deviation instead of doing buckets, it is a bit more complex for the user but we can still do arbitrary thresholds then on the front end, eventually they will be able to configure the thresholds anyway. The reason I didn't go with this in the first place is because the comment here  `proto/smartcore/bos/devices/v1/devices.proto:29-35` says measured values were deliberately omitted

Severity vs deviation: I went with deviation in the end, as that is objectively what it is. Severity is a bit subjective and clashed with the 'impact' semantics. I think we were deliberately avoiding severity

hexaglow (changes requested):
Zero-deviation doesn't match with less-than queries e.g. 

```go
func Test_conditionMatchesMessage_zeroValue(t *testing.T) {
	device := &devicespb.Device{Name: "Device01", HealthChecks: []*healthpb.HealthCheck{
		{Id: "check00", Deviation: 0},
	}}
	cond := &devicespb.Device_Query_Condition{
		Field: "health_checks.deviation",
		Value: &devicespb.Device_Query_Condition_FloatLte{FloatLte: 0.1},
	}
	if !conditionMatchesMessage(cond, device) {
		t.Errorf("deviation 0 should match float_lte 0.1")
	}
}
```

PR contains a duplicate of 704a41b50 (review voice), already merged.

deanredfern26:
`scanMessage` dropped any field equal to its proto3 default, so a zero deviation never reached the comparison. Numeric scalars now keep their zero; string/bool/enum defaults still read as unset. Added `Example_deviationThresholds` covering both directions.

Duplicate review-voice commit dropped, rebased onto current main.

hexaglow (approved):

PR #929, opened 2026-07-28
proto/supervisorpb: regenerate stale bindings and protoc stamps (SCB-1405)

deanredfern26:
Regenerating protos while working SCB-1392 surfaced drift in `supervisorpb`. The committed grpc-web bindings for `InstallUpdateRequest` still exposed `getSha256()`/`setSha256()`, but that field was renamed to `checksum` (field 3) in `supervisor.proto`. The renamed proto and the stale JS landed together in `751917d1a`, so the bindings have been out of sync with the proto ever since — the Go was already correct, only the UI bindings were left behind. Any full `scripts/gen-proto.sh` run regenerates them and leaks an unrelated diff into whatever PR happened to trigger the regen.

Nothing in ops UI source consumes these bindings yet, so the drift was latent rather than breaking, and field 3 remains a `string` — the rename is name-only with no wire-format change. The diff also brings the protoc stamps from `v7.34.0` up to the repo baseline `v7.35.1`. It is entirely generated output; no `.proto` file changed, because the proto was already right.

Two things a reviewer may want to know. A full regen currently also touches `cloudpb` (`v6.32.1` → `v7.35.1`) until SCB-1392's `chore/regen-cloudpb` merges, so regenerating locally will show extra churn that isn't in this PR — I discarded it deliberately to keep this to supervisorpb. And `pkg/driver/bacnet/rpc` is likewise stamped `v7.34.0`, but it has its own `gen.sh` invoked via `//go:generate` and is untouched by `scripts/gen-proto.sh`, so it cannot leak the same way; tracked separately as SCB-1408.

Verified by `go build ./...`, `go vet -lostcancel=false ./...`, `staticcheck ./...`, and `go test` over the supervisorpb consumers (`internal/supervisor`, `supervisor/internal/{config,install,server}`), plus `yarn --cwd ops lint:nofix`. A fresh `scripts/gen-proto.sh` run against this base reproduces the committed output byte-for-byte, confirming it is exactly what the pinned toolchain emits.

#SCB-1405 State Done


deanredfern26:
Heads up on the red **Run govulncheck** check — it is unrelated to this PR and does not indicate a problem with this change.

`govulncheck` is failing on [GO-2026-6061](https://pkg.go.dev/vuln/GO-2026-6061) in `google.golang.org/grpc` (we're on `v1.79.3`, fixed in `v1.82.1`). It's a newly published advisory rather than a regression: the same check **passed on the current `main` HEAD (`fe7a9bdb5`) on 23 Jul** against an identical dependency set, and the vulnerability database is fetched fresh on each run. Since `go-security.yml` triggers on `pkg/**`, it will now be red on every Go-touching PR until the bump lands.

Tracked as **SCB-1409**, deliberately kept out of this PR to keep it to generated-proto output only.

`lint-all` and `UI Lint` are green.

hexaglow (approved):

PR #931, opened 2026-07-30
auth/policy: audit in-process gRPC writes (SCB-1411)

deanredfern26:
The audit log from SCB-422 exists to answer "who turned the temp up to 30", but it's a policy interceptor on the external gRPC/HTTP servers, so it only ever saw writes that arrived over the network. Writes originating in-process go out through `Node.ClientConn()` — a router loopback that dispatches straight to the resolved conn — and never reach that chain.

Writes arriving over MQTT take exactly that path. The UDMI automation subscribes to each device's `/config` topic and forwards the payload to `UdmiService.OnMessage`, where the BACnet driver turns it into a real `WriteProperty` per configured point. Those are the least authenticated writes in the system — the broker supplies no smart-core identity — and they produced no audit entry and no log line at the default level.

The ticket proposed threading an `AuditSink` into `driver.Services` and auditing inside `udmiMerge.OnMessage`. This does it at the loopback instead, because downstream drivers implement the UDMI trait and actuate in their own `OnMessage`, and a driver-level fix would have left exactly the drivers we can't inspect unaudited until each one opted in. The two ingress paths are disjoint — external traffic reaches routed services via the server's unknown-service handler — so a decorator on the loopback conn is complementary to the existing interceptor, not a duplicate of it. No driver, automation or `Services` struct changes.

Two things worth a reviewer's attention. Audit-log content changes system-wide: every in-process write now appears, including `bms` `UpdateAirTemperature`/`UpdateModeValues`. Volume is bounded — `isWriteMethod` filters reads, `CacheWriteAction` suppresses unchanged writes, `CreateHistoryRecord` is already excluded — and entries are tagged `ingress=loopback` so consumers can filter. And `writeAuditEntry` still blocks when its 1024-entry queue fills; I kept that consistent with the external path deliberately, since silently dropping entries from a security audit log is the worse failure mode.

Per-point detail — which configured points were actually written, the resolved BACnet priority, per-point outcome — is only knowable inside the driver, so it's left for a follow-up along with `pestsense`, which mutates its resource model directly and so is out of reach of any gRPC-layer fix.

Verified by `go test ./pkg/auth/policy/ ./pkg/node/`, including a regression test that drives real router dispatch through the loopback and fails with `got 0 audit entries, want 1` when the wrapper is disabled, plus a test asserting an external call still yields exactly one entry. Full `go build`, `go vet`, `staticcheck` and `go test ./...` clean against this host's pre-existing baseline.

#SCB-1411 State Done

### DIFF
--- a/internal/node/nodeopts/options.go
+++ b/internal/node/nodeopts/options.go
@@ -4,6 +4,8 @@ package nodeopts
 import (
 	"context"
 
+	"google.golang.org/grpc"
+
 	"github.com/smart-core-os/sc-bos/internal/router"
 	"github.com/smart-core-os/sc-bos/pkg/proto/devicespb"
 	"github.com/smart-core-os/sc-bos/pkg/resource"
@@ -33,6 +35,15 @@ func WithRouter(r *router.Router) Option {
 	})
 }
 
+// WithClientConnWrapper sets a decorator applied to the connection returned by Node.ClientConn.
+// Every in-process caller obtains its connection from that method, so the wrapper sees all
+// in-process traffic. Used to audit writes that never reach the gRPC server's interceptors.
+func WithClientConnWrapper(wrapper func(grpc.ClientConnInterface) grpc.ClientConnInterface) Option {
+	return optionFunc(func(o *Struct) {
+		o.ClientConnWrapper = wrapper
+	})
+}
+
 // Join combines multiple options into a single struct.
 func Join(opts ...Option) Struct {
 	var o Struct
@@ -44,13 +55,17 @@ func Join(opts ...Option) Struct {
 
 // Struct contains all options for a Node as a struct for easy access.
 type Struct struct {
-	Store  Store
-	Router *router.Router
+	Store             Store
+	Router            *router.Router
+	ClientConnWrapper func(grpc.ClientConnInterface) grpc.ClientConnInterface
 }
 
 func (s Struct) apply(o *Struct) {
 	if s.Store != nil {
 		o.Store = s.Store
+	}
+	if s.ClientConnWrapper != nil {
+		o.ClientConnWrapper = s.ClientConnWrapper
 	}
 }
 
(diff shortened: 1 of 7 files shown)


hexaglow (commented):
This change takes the audit system away from being a log of mutating requests, and towards a hybrid of that and a log of all mutations. The way I normally think of an audit log, it sits at a permission boundary, recording security decisions and what actions were performed.

Have you considered an implementation that performs the logging closer to the ingress point? If the `udmi` automation performed the logging, it would be able to attach relevant details such as the origin of the control message. In the current implementation, these requests would be marked as loopback which, while true from a technical implementation point of view, doesn't attribute these requests to their actual source and therefore is less useful than existing log entries. 

Additionally, this change will result in multiple audit log entries for a lot of requests. Anything that forwards, fans out, or makes changes automatically, such as zones, lighting automations etc., will result in multiple (perhaps many) log entries. There's also no correlation ID that would let a log reader know that these actually all come from the same initial request. 

In summary I am curious whether you see a request audit system, or an internal trace system as being the right direction to take, and why.

deanredfern26:
> This change takes the audit system away from being a log of mutating requests, and towards a hybrid of that and a log of all mutations. The way I normally think of an audit log, it sits at a permission boundary, recording security decisions and what actions were performed.
> 
> Have you considered an implementation that performs the logging closer to the ingress point? If the `udmi` automation performed the logging, it would be able to attach relevant details such as the origin of the control message. In the current implementation, these requests would be marked as loopback which, while true from a technical implementation point of view, doesn't attribute these requests to their actual source and therefore is less useful than existing log entries.
> 
> Additionally, this change will result in multiple audit log entries for a lot of requests. Anything that forwards, fans out, or makes changes automatically, such as zones, lighting automations etc., will result in multiple (perhaps many) log entries. There's also no correlation ID that would let a log reader know that these actually all come from the same initial request.
> 
> In summary I am curious whether you see a request audit system, or an internal trace system as being the right direction to take, and why.

I think you're right this should be a request audit, not an internal trace. I will rework the PR

PR #933, opened 2026-07-30
auth/accesstoken: default SignatureAlgorithms when unset (SCB-1393)

deanredfern26:
`accesstoken.Source` decided which signature algorithms to accept from a field configured separately from the signing key, and `jwt.ParseSigned` treats that list as required — it fails outright when empty. A `Server` built without `WithPermittedSignatureAlgorithms` would hand out tokens over `/oauth2/token` quite happily and then reject every one with `no signature algorithms specified`: login succeeds, everything after it 401s. Nothing hit it, since `pkg/system/authn/factory.go` was the only constructor and always passed the option, but it was a footgun the tests structurally could not catch — `TestTokenServer` already built a `Server` without the option and only ever exercised issuing, never validation.

The fix removes the configuration rather than defaulting it. `NewServer` already required the signing key to be a `[]byte`, and go-jose routes a `[]byte` key to `newSymmetricSigner`, which accepts only HS256/HS384/HS512 and enforces a minimum secret length for each; `LoadOrGenerateSigningKey` mandates exactly 32 bytes, so HS256 was the only algorithm a `Server` could ever reach. The allow-list was unreachable configuration, and permitting the RS/ES/PS families could never match a token this `Server` issued.

So `WithPermittedSignatureAlgorithms` and `Source.SignatureAlgorithms` are gone, and the permitted algorithm is read off `Source.Key.Algorithm`. `GenerateAccessToken` and `ValidateAccessToken` now consult the same field, which is what makes this impossible rather than merely defaulted: a `Source` can no longer be configured to reject the tokens it issues, and `ParseSigned` can no longer receive an empty list. `NewServer` gained an `Algorithm == HS256` check beside the existing key-type check, so a key it cannot issue with is refused at construction rather than failing later at sign time.

`factory.go` stops passing `keycloak.DefaultPermittedSignatureAlgorithms` — required now the option is gone, and the same HS256-only change requested on #916. Whether HS256 should also come *out* of the Keycloak list is left to that PR: it is a behavioural change for any installation whose Keycloak signs with HS256, so it shouldn't ride along here.

Verified by `go test ./internal/auth/accesstoken/... ./pkg/system/authn/...`; the new `NewServer` validation was negative-controlled (its three algorithm cases fail with the check removed, the two key-shape cases still pass on the pre-existing `[]byte` check). `go build ./...`, `go vet`, `staticcheck ./...` and `gofmt` clean, and the full `go test ./...` failing-package set is unchanged.

#SCB-1393 State Done

### DIFF
--- a/internal/auth/accesstoken/accesstoken.go
+++ b/internal/auth/accesstoken/accesstoken.go
@@ -61,11 +61,35 @@ func (pa permissionAssignment) ToTokenPermissionAssignment() token.PermissionAss
 	}
 }
 
+// DefaultSignatureAlgorithms lists the signature algorithms a Source permits when
+// SignatureAlgorithms is not set.
+//
+// It contains only HS256 because that is the only algorithm this package ever signs with:
+// generateKey and LoadOrGenerateSigningKey both produce HS256 keys, and WithSigningKey rejects
+// anything that isn't the []byte an HMAC key needs. A Source therefore only ever verifies its
+// own HS256 tokens, so this is the narrowest list that still works rather than a lax fallback.
+//
+// jwt.ParseSigned takes the permitted algorithms as a required argument and fails when handed an
+// empty list, so without this default a Source built without WithPermittedSignatureAlgorithms
+// would issue tokens happily and then reject every single one of them.
+var DefaultSignatureAlgorithms = []string{string(jose.HS256)}
+
 type Source struct {
-	Key                 jose.SigningKey
-	Issuer              string
-	Now                 func() time.Time
+	Key    jose.SigningKey
+	Issuer string
+	Now    func() time.Time
+	// SignatureAlgorithms are the signature algorithms permitted when validating tokens.
+	// When empty, DefaultSignatureAlgorithms is used.
 	SignatureAlgorithms []string
+}
+
+// permittedAlgorithms returns the signature algorithms to accept when validating a token,
+// falling back to DefaultSignatureAlgorithms when none were configured.
+func (ts *Source) permittedAlgorithms() []string {
+	if len(ts.SignatureAlgorithms) == 0 {
+		return DefaultSignatureAlgorithms
+	}
+	return ts.SignatureAlgorithms
 }
 
 func (ts *Source) GenerateAccessToken(data SecretData, validity time.Duration) (token string, err error) {
@@ -108,7 +132,7 @@ func (ts *Source) GenerateAccessToken(data SecretData, validity time.Duration) (
 }
 
 func (ts *Source) ValidateAccessToken(_ context.Context, tokenStr string) (*token.Claims, error) {
-	tok, err := jwt.ParseSigned(tokenStr, joseUtils.ConvertToNativeJose(ts.SignatureAlgorithms))
+	tok, err := jwt.ParseSigned(tokenStr, joseUtils.ConvertToNativeJose(ts.permittedAlgorithms()))
 	if err != nil {
 		return nil, err
 	}
--- a/internal/auth/accesstoken/accesstoken.go
+++ b/internal/auth/accesstoken/accesstoken.go
@@ -17,7 +17,6 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/smart-core-os/sc-bos/internal/auth/permission"
-	joseUtils "github.com/smart-core-os/sc-bos/internal/util/jose"
 	"github.com/smart-core-os/sc-bos/pkg/auth/token"
 	"github.com/smart-core-os/sc-bos/pkg/proto/accountpb"
 )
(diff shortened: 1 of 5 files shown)


hexaglow (commented):
Reading the surrounding code as part of this PR, I've realised something. `NewServer` validates that the provided key is a `[]byte` which is used only for `HSxxx` algorithms. So the `Server` can only support those algorithms, even when an external key is used. Technically another HS-family algorithm, aside from HS256, could be used by passing a custom key and setting the algorithm accordingly, but we don't need this feature in BOS.

Consequence: there's no real reason to use WithPermittedSignatureAlgorithms with anything except HS256. 

So I suggest, instead of adding the default, we:
1. Remove WithPermittedSignatureAlgorithms entirely
2. In NewServer, validate that the passed signing key has `Algorithm == HS256`, in addition to the type check

This simplifies the code.

Further, Source.SignatureAlgorithms could be removed, with the permitted (single) algorithm instead taken from Source.Key.Algorithm instead.

Thoughts?

deanredfern26:
> Reading the surrounding code as part of this PR, I've realised something. `NewServer` validates that the provided key is a `[]byte` which is used only for `HSxxx` algorithms. So the `Server` can only support those algorithms, even when an external key is used. Technically another HS-family algorithm, aside from HS256, could be used by passing a custom key and setting the algorithm accordingly, but we don't need this feature in BOS.
> 
> Consequence: there's no real reason to use WithPermittedSignatureAlgorithms with anything except HS256.
> 
> So I suggest, instead of adding the default, we:
> 
>     1. Remove WithPermittedSignatureAlgorithms entirely
> 
>     2. In NewServer, validate that the passed signing key has `Algorithm == HS256`, in addition to the type check
> 
> 
> This simplifies the code.
> 
> Further, Source.SignatureAlgorithms could be removed, with the permitted (single) algorithm instead taken from Source.Key.Algorithm instead.
> 
> Thoughts?

Sounds good to me, I will do that

internal/auth/accesstoken/accesstoken.go
```diff
@@ -61,11 +60,19 @@ func (pa permissionAssignment) ToTokenPermissionAssignment() token.PermissionAss
...
 
+// Source issues and validates access tokens, both signed with Key.
+//
+// Key must hold a []byte secret and an HMAC algorithm it is long enough for: a []byte is the only
+// key type go-jose signs symmetrically with. generateKey and LoadOrGenerateSigningKey both produce
+// HS256, and NewServer requires HS256 specifically.
+//
+// Key.Algorithm is the single source of truth for both operations: GenerateAccessToken signs with
+// it and ValidateAccessToken permits it and nothing else. A Source therefore cannot be configured
+// to accept an algorithm it could never issue, nor to reject its own tokens.
```
hexaglow: Nit: Server is the user of Source, and Source is written to be independent of Server. Therefore any comments about the interaction of Source and Server would fit better on Server, perhaps as an implementation comment rather than a doc comment. Also, Source seems to be compatible with any key type, it's only Server that specifies HS256.

Nit: Comments describing the behaviour of a field should be on that field.

internal/auth/accesstoken/accesstoken_test.go
```diff
@@ -24,6 +26,58 @@ func TestTokenSource_createAndVerify(t *testing.T) {
...
+	ts := &Source{Key: key, Issuer: "test", Now: time.Now}
+
+	token, err := ts.GenerateAccessToken(SecretData{TenantID: "Foo"}, 10*time.Minute)
+	if err != nil {
+		t.Fatalf("GenerateAccessToken %v", err)
+	}
+	if _, err := ts.ValidateAccessToken(t.Context(), token); err != nil {
+		t.Fatalf("ValidateAccessToken %v", err)
+	}
+}
```
hexaglow: What is this intended to test? It doesn't need any extra configuration, true, but the field where that configuration would go has been removed, so that seems vacuously true. Looks like a duplicate of `TestTokenSource_createAndVerify`

internal/auth/accesstoken/token_server_test.go
```diff
@@ -173,6 +175,84 @@ func TestTokenServer(t *testing.T) {
 	}
 }
 
+// A Server must accept the tokens it issues. jwt.ParseSigned rejects everything when handed an
+// empty algorithm list, so while the permitted algorithms were configured separately from the key
+// a caller could build a server that handed out a token over HTTP and then refused that very
+// token — see SCB-1393.
```
hexaglow: Comment refers to a condition that no longer exists in the code, confusing for the reader. Rephrase as the invariant you are trying to preserve.

hexaglow (changes requested):

internal/auth/accesstoken/accesstoken.go
deanredfern26: Trimmed the Source doc to Source, moved the algorithm invariant onto Key, and put the HS256 rationale in NewServer where it's enforced.

internal/auth/accesstoken/accesstoken_test.go
deanredfern26: Agreed, vacuous. Deleted.

internal/auth/accesstoken/token_server_test.go
deanredfern26: Reworded to the invariant.

deanredfern26:
CI note: `Go Test` is red here for an unrelated reason — the paxton SignalR data race, which fails identically on `main` (`9b154513a`). Upstream bug in `nhooyr.io/websocket@v1.8.10`; tracked as SCB-1419. `lint-all` and `govulncheck` pass.

internal/auth/accesstoken/accesstoken.go
```diff
@@ -61,11 +60,13 @@ func (pa permissionAssignment) ToTokenPermissionAssignment() token.PermissionAss
...
 }
 
+// Source issues and validates access tokens, both signed with Key.
 type Source struct {
-	Key                 jose.SigningKey
-	Issuer              string
-	Now                 func() time.Time
-	SignatureAlgorithms []string
+	// Key.Algorithm is the only algorithm Source will issue with or accept, so a Source can
+	// neither be configured to permit an algorithm it could never issue nor to reject its own tokens.
```
hexaglow, line 66: Nit: this comment makes sense if you've read this PR thread. It's not very useful if you hadn't - someone coming across this code fresh would have to reason to believe it could "be configured to permit an algorithm it could never issue nor to reject its own tokens". Instead, the comment would do better to emphasise that the same key is used for both signing and validation.

hexaglow (approved):

PR #935, opened 2026-08-03
docs: record review voice conventions

deanredfern26:
Reviews and review replies written by Claude have been costing @hexaglow time: change-lists since the last round, recaps of points already resolved, opening compliments, the same point made in both the review body and a line comment, and nits whose cost to action exceeds their value. He gave the list; this records it so it binds every session instead of being pasted in as a one-off prompt argument.

The pointer sits in a new root `CLAUDE.md` rather than a skill, because the biggest producers of this prose are the built-in `/code-review` and `/review`, which can't be edited from the repo. `.claude/CLAUDE.md` would have been the natural home but it's git-excluded and machine-local, so the team would never have seen it. PR descriptions are deliberately out of scope — they carry reasoning a reviewer needs up front, and the doc says so to stop a future session over-applying the rules.

Verified by confirming both new files are trackable (`git check-ignore` clean) and that the pointer resolves; the real test is behavioural, on the next `/code-review` run in a fresh session.

### DIFF
--- a/.claude/review-voice.md
+++ b/.claude/review-voice.md
@@ -1,0 +1,43 @@
+# Review voice
+
+How to write anything a reviewer will read: PR review bodies, line comments, and
+replies to review comments.
+
+Not PR descriptions — those carry reasoning a reviewer needs up front and are
+deliberately out of scope.
+
+## Length follows the diff
+
+- Scale the review to the PR's impact, not to how much there is to say. Internal
+  docs, cleanup, dep bumps: shorter than normal.
+- A review longer than the diff it reviews is almost always wasted effort.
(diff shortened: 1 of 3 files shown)


.claude/review-voice.md
```diff
@@ -0,0 +1,43 @@
...
+  flattery, no opening compliment.
+
+## Replies to review comments
+
+- One or two lines. What was done, or why the current approach stands.
+- Don't restate the reviewer's point back at them. No thanks-for-catching-this, no
+  summary of the wider change.
+- Disagreeing is fine — give the concrete reason in a sentence.
+
+Source: review feedback from @hexaglow, August 2026.
```
hexaglow: Not necessary.

